PR #94, opened 2016-09-21
fix:  List item styling on long text

XristMisyris:


XristMisyris:
bug on list

rstoenescu:
There seems to be a miaunderatanding. Lists are either one line, two lines or multiple lines. In your case what you want is to add `multiple-lines` to your list item. Or in case you want the content on one line but truncated, then wrap string in a `div` inside list item node.

Do you think the documentation needa a tweak for this?

XristMisyris:
Every item of my list has different length. First item has 1 line length second item has 3 lines length, so i cant use any class( two-lines, three-lines etc ), so if i want the full content of the item i think then only solution is to change the height of the item to "auto"

rstoenescu:
What about a new class "auto-lines" besides one-line, two-lines etc? This new class should have height auto.

Lists are not for everything. Usually if you have lots of content a collapsible should be used.
Or simply use grid layout flexbox classes to craft your own UI.

XristMisyris:
Ok thank you @rstoenescu i will use grid layout !

PR #110, opened 2016-09-28
Add missing moment dependency

vuchl:
after forking and cloning the repo `npm run dev` errors out and demand moment to be installed

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,5 +127,8 @@
     "velocity-animate": "^1.2.3",
     "vue": "^1.0.26",
     "vue-touch": "^1.1.0"
+  },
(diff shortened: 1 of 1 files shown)


rstoenescu:
First of all thank you your effort!

Indeed there's an issue. "moment" should be a dev dependency, otherwise it will install a specific moment version in all apps. Each app must be able to manage its own dependencies... for example a certain momentjs version etc. That's why "moment" is currently only under "peerDependencies" which tells NPM that an app should install momentjs along quasar-framework.

Long story short, "moment" needs to be under "devDependencies" :)

PR #119, opened 2016-10-04
Vue2 migrations

kristianmandrup:
Vue2 migrations per the migration guide. 
- Next step is `.sync`, `.once`.
- Also still missing `twoWay` fixes for non-trivial cases.


smolinari:
Great work on the Vue 2.0 migration and congrats on the release! Not to far away from a 1.0! 😄 
 
Scott

PR #137, opened 2016-10-23
chore: Remove obsolete hammerjs reference

plievone:


rstoenescu:
Thanks for this! Nice catch.

PR #146, opened 2016-10-28
feat: Modal - ability to force use of buttons to dismiss

ccoxtn:
Use noCancelOnOutsideClick to prevent modal from closing with click outside modal area. Use noCancelOnEscKey to prevent modal from closing with Esc key press.

### DIFF
--- a/src/vue-components/modal/Modal.vue
+++ b/src/vue-components/modal/Modal.vue
@@ -4,7 +4,7 @@
       v-show="active"
       class="modal fullscreen flex"
       :class="positionClasses"
-      @click="close()"
+      @click="click()"
     >
       <div ref="content" class="modal-content" @click.stop :style="contentCss">
         <slot></slot>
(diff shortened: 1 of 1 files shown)


rstoenescu:
Good job! Thanks!

rstoenescu:
Any ideas on a shorter name for those props? They're too verbose.

ccoxtn:
How about: noClickCancel / noEscCancel?

rstoenescu:
Thought about this more. Will there ever be a case where you would not want to be able to dismiss with escape key but be able to dismiss by clicking/tapping outside of the modal? Should we merge the two props into a single Boolean one? "dismissable" or "notDismissable".. What do you think?

ccoxtn:
The click is dangerous if the user is filling out a form in the modal and accidentally clicks outside the form. However, it makes sense to still have the Esc key sometimes. If the form had a [Cancel] button, the Esc could allow an intentional close of the modal. Having the two properties is more powerful, but one would work for me.

rstoenescu:
You're right. Ok, let's go with "noBackdropDismiss" and "noEscDismiss"?

ccoxtn:
Sounds good.

rstoenescu:
Pushed commit to "dev" branch to change name of those props to "noBackdropDismiss" and "noEscDismiss". Will be available in next Quasar version.

PR #173, opened 2016-11-08
Making container center on >big screens (1280px)

murbanowicz:
Added margin auto to ios and mat to get `layout-padding` container center, as it should be


rstoenescu:
Thanks! Merged in.

PR #225, opened 2016-11-23
Dev Builds

blvz:
`npm run deploy` on a branch other than `master`, will publish quasar [with that flag](https://docs.npmjs.com/cli/tag). For example:

Running `npm run deploy` from the branch `dev`, will publish quasar to `quasar-framework@dev`. To install that version, one would need to run `npm i quasar-framework@dev`.

Running `npm run deploy` from `master`, will publish quasar to `quasar-framework@latest`, which is the default and can be installed with `npm i quasar-framework` as usual.

### DIFF
--- a/build/shell.deploy.sh
+++ b/build/shell.deploy.sh
@@ -1,12 +1,20 @@
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+if [ "$BRANCH" == "master" ]
+then
+  TAG="latest"
+else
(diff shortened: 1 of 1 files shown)


rstoenescu:
Merged. Thanks for this!

There is still an issue as you can't publish same version twice on NPM, even though the tag is different. We should also make a script to temporarily add a build number or something to the version, like "beta-1" etc.

PR #226, opened 2016-11-26
fix: Dialog/ActionSheet are not destroyed in some situations

llops:
These components are not properly destroyed when they are closed using the Escape key or clicking the Modal backdrop.  Although hidden, the instances are alive and it could lead to memory leak.

quasar


rstoenescu:
Good patch! Thanks for taking time for a PR!!! This will get into v0.9.

PR #248, opened 2016-12-13
fix: Removed max-width on layout style

arddor:
Removes the max-width on layout-padding, which results in just padding on all screen sizes. If a max-width is set, it will be centered on all screens.


rstoenescu:
Thanks!

PR #249, opened 2016-12-13
Support for password in a Dialog

livioalves:
Support for password in a Dialog


rstoenescu:
Good idea. Thanks for the PR!

PR #255, opened 2016-12-15
Firefox - Hide the numeric spinner

panovskidejan:
Hide the numeric spinner so the number is visible.


rstoenescu:
Thanks for the PR!

PR #272, opened 2016-12-19
Fix DataTable error when no data and not messages.

tant42:
Do and extra eval to handle when messages aren't defined on the config. Also make sure we look up the 'noData' message off config.messages and not config.message


rstoenescu:
Good catch. Thanks for the PR!

PR #306, opened 2017-01-10
stepper: allow custom action before moving to next step

zh012:
Allow inject custom action before the stepper moving to next step.

Usage: 

In template:

```
<q-stepper>
  <q-step :beforeNext="verifiyAnswer">
     <input v-model="answer">
  </q-step>
  <q-step>Congrats!</q-step>
</q-stepper>
```

In script:
```
export default {
  data () { return { answer: "foo" } },
  methods: {
     verifyAnswer (next) {
        if (this.answer === "bar") {
           next()
        }
        else {
           alert('wrong answer')
        }
     }
  }
}
```

### DIFF
--- a/src/vue-components/stepper/Step.vue
+++ b/src/vue-components/stepper/Step.vue
@@ -48,6 +48,10 @@ export default {
     ready: {
       type: Boolean,
       default: true
+    },
+    beforeNext: {
+      type: Function,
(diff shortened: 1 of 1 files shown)


rstoenescu:
Thanks for the PR! Good job.

PR #310, opened 2017-01-10
Simplify Infinite Scroll markup

boydcl:
This `<br>` shouldn't be here. I assume its here for styling, but that should be left to css.


rstoenescu:
Thanks. Will add CSS for it.

rstoenescu:
Actually scratch that. No Quasar CSS is required.

boyd91:
I agree. It's a good chance people will place already styles elements/components inside the slot.

rstoenescu:
Yep. Many thanks for taking time for a PR!

boyd91:
np I greatly appreciate the work you're doing

PR #315, opened 2017-01-11
Stepper components - event 'step' proposal

fsgiudice:
Hi @rstoenescu,
added event 'step' emitted on step change.


rstoenescu:
Yes, good idea. Thanks for taking time for a PR!

PR #318, opened 2017-01-12
Update Numeric.vue

Vaerum:
Added the tag step="any" so Firefox do not make a validation error.


rstoenescu:
Thanks for this! Fix available in edge version and in future v0.12.1 (soon).

PR #319, opened 2017-01-12
feat: Added Accordion component

fsgiudice:
Hi @rstoenescu, 
check if it is ok


smolinari:
Um. I haven't tried out your new component yet, but does the collapsibles do possibly the same thing or similar? 

http://quasar-framework.org/components/collapsible.html

Or asked differently, what does the accordion component do differently/ better?

Scott

fsgiudice:
Hi @smolinari,
absolutely, as I cloned collapsible component. But the difference AFAIK is that accordion has master component with childrens (<q-accordion-item>). 
Probably could be the inverse, collapsible is equal to accordion, but not close opened elements.

smolinari:
I guess I'll have to play with it to see. Thanks for the effort for sure!

Scott

fsgiudice:
You are welcome.
Francesco

fsgiudice:
Hi @rstoenescu , 
what have I to do with this PR? As I cannot approve my PRs.
TIA
Francesco

druppy:
I quess Razvan need to do that ? :-)

rstoenescu:
Hi. Please wait until I review this and will push it. Thx for tour efforts!

fsgiudice:
Hi @druppy , 
yes I know :-), normally, but because Razvan has added me yesterday as collaborator (with write access rights, as I saw)  and due to my poor knowledge of git, I was asking if I can do something to write PR to repo.

fsgiudice:
Added some other features, but now are on this PR.
Question: may I close a single PR and create new one for new changes?

rstoenescu:
@fsgiudice Hi, thanks for contributing. What I'd like is for us to avoid code duplication and make the lowest footprint possible. So rather than duplicating Collapsible it would be better to enhance Collapsible. What do you think?
Also, please try to make a separate PR for the other enhancements.

fsgiudice:
Ok Il will check

rstoenescu:
https://github.com/quasarframework/quasar/issues/326
Hope you don't mind. Had a quick idea and implemented it.

PR #323, opened 2017-01-14
Selected event in Autocomplete component

wilcorrea:
Can you analyse if this a good way to catch a selected event?

It works, but I dont know if is the best way, best practice and if can generate unwanted side effects


rstoenescu:
Good job. Thanks for taking time for a PR!! This will go into v0.12.1.

wilcorrea:
@rstoenescu do you think what is convenient change the docs and send PR there with that information?

PR #327, opened 2017-01-16
fix: Don't close drawer on resize

d3v2a:
when there is a textbox in the drawer, the keyboard resize the window and close the drawer


rstoenescu:
This needs careful thought. Will do some tests in the following days and merge it. Thanks for this PR!

rstoenescu:
The fix is more complex. Will merge this and make the necessary additions.

rstoenescu:
Thanks for reporting and making a PR for this!

PR #330, opened 2017-01-18
Display configured theme on dev server start

azurelogic:


rstoenescu:
Hi, thanks for contributing!
Got a quick question: how is this any different from the current banner?

rstoenescu:
And answering myself: guess it's a little more clean and does not compute theme name twice. Avoiding duplication is good. Merging it. :)
Thanks!

azurelogic:
Thanks for the prompt handling! It's worth noting that it's not just avoiding duplication. I changed the default in my project to the iOS theme and noticed that it was still claiming to be configured for material. I found this hardcoded string to be the issue. Thanks for the merge!

rstoenescu:
The default app starter kit now contains a new prop in `/config/index.js` called `defaultTheme`. Changing that will set the default theme to build with.

PR #331, opened 2017-01-18
Add validation to DrawerLink

azurelogic:
Leaving off the "to" attr causes a nasty error that is difficult to track down. If it's going to error without 'to', make it required.


rstoenescu:
Changed a little and merged. Thanks for taking time for a PR!

azurelogic:
I like the change. I forgot that you could specify multiple validation types in an array. Good catch and thanks for the merge!

rstoenescu:
Thank YOU for the PR!

PR #362, opened 2017-01-31
Added jump to step on click if step is active

dragospaulpop:
Whenever a user clicks on a step's badge, the stepper jumps to that step but only if that step is ready.

### DIFF
--- a/src/vue-components/stepper/Step.vue
+++ b/src/vue-components/stepper/Step.vue
@@ -3,7 +3,7 @@
     class="timeline-item"
     :class="{incomplete: step > stepper.currentStep}"
   >
-    <div class="timeline-badge">
+    <div class="timeline-badge" @click="goToStep()">
       <i v-show="!done">
         done
       </i>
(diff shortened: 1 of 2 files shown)


rstoenescu:
Hi, Thanks for taking time for a PR! Got one question: how about handling `beforeNext`? I think user should be able to just go back step-wise but not forward to avoid issues. Also, this should be enabled with a flag (probably as a prop for Stepper). What do you think? (Oops, that's the second question :) )

dragospaulpop:
Maybe `beforeNext` should only be triggered whenever a forward jump of 1 step occurs? Or maybe have a `beforeJump` handler? 
Thinking about my personal use case, I think something like this would make sense:
- only allow back jumps if stepper isn't finished
- allow free jumps between steps if stepper is finished
- enable this with a prop that's `false` by default

Or maybe just handle all possible use cases we can think of with props.

rstoenescu:
Hi, thanking you again for taking time for a PR. v0.14 has a revamp of Stepper which has a different approach, more flexible and allows to do what this PR is intended to do among other things.

PR #386, opened 2017-02-09
fix: fullscreen iOS modal fix

oswaldofreitas:
on iOS Safari part of the fullscreen modal is hidden because vh css unit doesn’t ignore browser toolbar

Refer to this:
https://github.com/bokand/URLBarSizing#proposed-changes-to-chrome

Used this example as base:
https://css-tricks.com/forums/topic/setting-a-div-height-to-the-window-viewport-size/


rstoenescu:
Merged it but will make some adjustments. Thanks for this!

PR #403, opened 2017-02-19
added - additionalFields functionallity to Uploader

delueg:
- useful for image services like cloudinary.com where you have to specify extra fields in order to have your upload accepted e.g. upload_preset


rstoenescu:
Thanks for this!

PR #428, opened 2017-03-01
Updated height & width variable reference

aherrick17:
Radio component .q-radio class properties height & width were pointing at checkbox variable.


rstoenescu:
Good catch. Thx for taking time for a PR!

PR #431, opened 2017-03-02
Header & footer flex-shrink

### LINKED ISSUE #422: q-layout issue on iOS safari
I have a template similar to this:

```html
<q-layout>
  <div slot="header" class="toolbar">
    ...
  </div>

  <div class="layout-view">
    <q-infinite-scroll>
      ...
    </q-infinite-scroll>
  </div>
</q-layout>
```

This works fine in desktop browsers and Android, but on iOS 9 Safari the content of the infinite scroll is covered by the toolbar. I'm using a flexbox based grid in the infinite scroller. Is this an issue with the framework? Or could there be something wrong on my end?

stephentuso:
Fixes #422 

[More info](https://github.com/philipwalton/flexbugs#1-minimum-content-sizing-of-flex-items-not-honored)

### DIFF
--- a/src/vue-components/layout/layout.ios.styl
+++ b/src/vue-components/layout/layout.ios.styl
@@ -37,6 +37,7 @@ $toolbar-padding           ?= 4px 10px
   color $toolbar-color
   background $toolbar-background
   z-index $z-marginals
+  flex-shrink 0
 
 .toolbar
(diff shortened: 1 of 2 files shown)


rstoenescu:
Thanks for taking time for a PR!

PR #433, opened 2017-03-02
Collapsible component - fixed content during animation

stephentuso:
I thought it looked weird to have the content size change when the collapsible component is animating. This adds a wrapper div for the animation to run on, preventing the content from moving.


rstoenescu:
I guess you are right. Will enhance the animation itself too. Thanks for this!

PR #434, opened 2017-03-02
Changed global dialog component to enable disabled radio options

andreaswillems:


rstoenescu:
Good idea.

rstoenescu:
Pushed a commit to add disable for checkboxes and toggles too.

PR #442, opened 2017-03-06
feat: add typescript definitions

d3v2a:
i think it's not perfect (i start using typescrypt one week ago) but it's a good starting point.

  Utils, LocalStorage, SessionStorage is not done due lack of time ...

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "dist",
     "src"
   ],
+  "typings": "types/index.d.ts",
   "scripts": {
     "clean": "node build/script.clean.js",
     "dev": "node build/script.dev.js",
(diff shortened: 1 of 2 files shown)


icfr:
typescrypt definition finished fully tested in vsc

igorgo:
I think that it is not necessary here.
There is a special project for such definitions - https://github.com/DefinitelyTyped/DefinitelyTyped.

John0x:
@igorgo it would benefit everyone if it was included in the main repo. The @types (formerly DefinitelyTyped) are for third party typings

nothingismagick:
@outofmemoryagain and @rstoenescu - is this still helpful or can it be closed because the work being done on the typedefs in core and this app-extension here:

https://github.com/quasarframework/app-extension-typescript

PR #453, opened 2017-03-10
Slider infinite scrolling

stephentuso:
I needed infinite scrolling so I modified this quickly, it should probably be cleaned up some. Opening this for some feedback.

This works by moving the first slide to the right of the last one when it is scrolled to the end, and the last to the left of the first when scrolled to the start (using the `order` css property). Empty divs were added before and after the `slot` to fill the space when the order is changed.

### DIFF
--- a/dev/components/components/slider.vue
+++ b/dev/components/components/slider.vue
@@ -34,9 +34,9 @@
       </q-slider>
 
       <p class="caption">
-        Slider with Centered Content
+        Slider with Centered Content and Infinite Scrolling
       </p>
-      <q-slider arrows dots class="text-white">
+      <q-slider infinite arrows dots class="text-white">
         <div slot="slide" class="bg-primary centered">
           Slide 1
         </div>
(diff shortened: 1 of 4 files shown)


rstoenescu:
Thinking of also adding an "animation" prop so it will jump from one slide to another like a carousel.

stephentuso:
Thanks, sounds good. By animation do you mean automatically moving to the next slide? That would work nicely with the infinite scrolling, and I think "autoplay" is a common term for that

rstoenescu:
Yes, exactly. Meant autoplay. Works beautifully with your work. Thanks again! Will release this tomorrow.

danikane:
Hey guys, thanks for this great addition! Is it possible to add this to Gallery Slider as well?

rstoenescu:
Damn, totally forgot about it. Wanted to do it for Gallery too. Will push a commit today or tomorrow. Thanks for pointing this out!

PR #454, opened 2017-03-10
fix: Slider next/back button not working corretly after mount()

stephentuso:
There is a bug with the slider the first time the right arrow button is pressed - can be seen in the [demo](http://quasar-framework.org/components/slider.html#Basic-Slider), click the right arrow without dragging the slider at all and it will scroll behind the first slide, showing nothing.

This happens because `slidesNumber` isn't up to date and is still `0`, so it scrolls to `-1`. It seems like setting it in `beforeChange` should've worked, changing it to a computed property fixes the issue though.

Edit: This bug also caused the indicator dots to not show until the slider was dragged

### DIFF
--- a/src/vue-components/slider/Slider.vue
+++ b/src/vue-components/slider/Slider.vue
@@ -1,6 +1,7 @@
 <template>
   <div class="q-slider" :class="{fullscreen: inFullscreen}">
     <div class="q-slider-inner">
+      <p class="text-dark">{{slidesNumber}}</p>
       <div
         ref="track"
         class="q-slider-track"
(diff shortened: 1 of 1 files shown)


rstoenescu:
The fix is more complex to catch slots number changing so took the liberty to make some additions. My first try was to use a computed property on $slots too, but $slots is not reactive, so it does not update the slides number value when it should.
Thanks for this one and for highlighting the issue!

stephentuso:
Ah okay I didn't know $`$slots` wasn't reactive, thanks for fixing that

PR #472, opened 2017-03-20
Allow to custom name when using upload component

qwIvan:


rstoenescu:
Made a small tweak. Will merge but got one thing to clear out before this: What about when uploading multiple files? All of them will have same name.

rstoenescu:
Thanks for taking time for a PR! This will go into v0.14 (may take a month until then).
Do you need this fast?

qwIvan:
Thanks a lot!
I can modify dist file in node_modules before the new version release.

rstoenescu:
Use the edge version instead. http://quasar-framework.org/guide/quasar-nightly-build.html
It's going to be pretty stable until v0.14 is released. Only safe fixes will go into edge (v0.13.X stream).

qwIvan:
Good, very useful

PR #484, opened 2017-03-27
init <input type="file"> after added a file

qwIvan:
If you select a file, then commit it, when you select the same file in next time, it will not trigger the `onchange` event, so the solution is set `file.value = ''` after added a file.


rstoenescu:
Hi,

Thanks for taking time for PR(s)! This is available in edge version and future v0.14.

PR #491, opened 2017-03-30
support focus and blur on search component

qwIvan:


rstoenescu:
Thanks for this! Available in edge.

PR #497, opened 2017-04-01
Fix #427: Make css grid width options go up to 12

### LINKED ISSUE #427: [Request] More css grid width options
It would be nice to have the option of more than 5 columns - Bootstrap has up to 12. I can always manually add it in my styles, but I think it would be convenient to have a few more built in. Even just one more would be nice, but 8 seems like a good number if that wouldn't make the CSS bloated.

stephentuso:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

Updates an existing feature - makes `width-XofY`/`offset-XofY` maximum 12 

- [ ] Yes
- [x] No

Fixes issue #427

### DIFF
--- a/dev/components/css/grid.vue
+++ b/dev/components/css/grid.vue
@@ -38,6 +38,22 @@
           <div class="auto"><div>auto</div></div>
         </div>
 
+        <div class="row gutter wrap justify-stretch content-center text-center">
+          <div class="width-1of2"><div>1/2</div></div>
+          <div class="width-1of3"><div>1/3</div></div>
+          <div class="width-1of6"><div>1/6</div></div>
+        </div>
+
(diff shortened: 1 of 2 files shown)


rstoenescu:
The problem with this is that it increased the CSS output with around 55KB (even after applying some tricks over your PR), which is not what we want. Pushing another solution: a Stylus variable called `$grid-cols`. Whoever wants can enhance the current number of possible columns in the grid.

Thanks for doing a PR though!

rstoenescu:
The CSS Stylus variable is available in edge and future v0.14.

stephentuso:
Ahh okay that's what I was worried about.. thanks, that's a good solution

PR #530, opened 2017-04-24
feat: Add direction as a second parameter to the Slider's slide event.

dobromir-hristov:
Adds the direction of the current slide action to the Slider component.

### DIFF
--- a/src/components/slider/QSlider.vue
+++ b/src/components/slider/QSlider.vue
@@ -153,10 +153,11 @@ export default {
       }
     },
     goToSlide (slide, done) {
+      let direction = ''
       this.__cleanup()
 
       const finish = () => {
-        this.$emit('slide', this.slide)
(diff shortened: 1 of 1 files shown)


rstoenescu:
Merged. Thanks for this!

dobromir-hristov:
I have to update the Docs.

rstoenescu:
@dobromir-hristov just visit http://quasar-framework.org/components/slider.html and hit the pencil icon on the right of page title OR the "Suggest an edit on GitHub" at the bottom of page. Many thanks!

dobromir-hristov:
Yeah, I just thought that being for v0.14 it used a special branch or something. Will do i there then.

rstoenescu:
We're preparing for v0.14 release, so all "dev" branches are for v0.14 now. Thanks for taking care of documentation too!

PR #554, opened 2017-05-09
Updated buttons.vue

smolinari:


rstoenescu:
Hey, I'd prefer keeping the current examples for testing purposes. The real examples will be part of Quasar Play, so that's where we can do something styling on what we show as demo.

smolinari:
Ok. Though, are the tests automated and need those exact examples? 

Because, all the examples are basically still the same, I just added one more button to show a customized spinner and I also added a click event to one button in the row of spinners examples, so all of the buttons do spin when clicked on. I also added a caption for the row of spinner buttons to show they are spinners. Nothing more really. 😄 

Scott

smolinari:
WTF! Oh my. Naturally not what I wanted. Ok. I'll fix and re-PR.

Scott

PR #592, opened 2017-05-26
IE related fixes

pdanpdan:
Can you please take a look on this fixes for QBtn in IE/Edge.
The buttons as they are now are broken.
I need some more eyes to check if this doesn't break something else.


pdanpdan:
Not good yet, need some more work.

rstoenescu:
Ah, dammit. Wanted to take care of this, but seems like you did a pull and this PR now contains a lot of extra commits of my own.

pdanpdan:
Sorry, I just found a corner case where it was not working - multiline text button.
I'll try some more fixes and come back.

rstoenescu:
Thank you so much for your hard work!

PR #594, opened 2017-05-27
fix utils isNumber

ceau:

 Bugfix

No

fix the utils isNumber

### DIFF
--- a/src/utils/is.js
+++ b/src/utils/is.js
@@ -20,7 +20,7 @@ export function isRegexp (v) {
 }
 
 export function isNumber (v) {
-  return typeof v === 'boolean'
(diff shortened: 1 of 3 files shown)


src/utils/is.js
```diff
@@ -20,7 +20,7 @@ export function isRegexp (v) {
 }
 
 export function isNumber (v) {
-  return typeof v === 'boolean'
+  return !isNaN(parseFloat(v)) && isFinite(v)
```
rstoenescu: This actually returns `true` when checking `'2`, which is wrong.
A better way would be: `return typeof v === 'number' && isFinite(v)`

Issue #611, opened 2017-06-01
v-link always replace instead of push when use a string


rottmann:
Problem is this line, when ctx.route is a string .replace is always true
```
vnode.context.$router[ctx.route.replace ? 'replace' : 'push'](ctx.route);
```

Version 0.13.9

rstoenescu:
Hi, thanks for reporting this! Will fix it tomorrow and release v0.13.10.

rstoenescu:
Anyone care for a PR? Should be really easy (just check type of `ctx.route` before looking at "replace" prop). I'm overloaded with v0.14 release.

rstoenescu:
Published v0.13.10 with the fix.

PR #614, opened 2017-06-03
Fix: v-link always replace instead of push

rottmann:
Issue: https://github.com/quasarframework/quasar/issues/611


rstoenescu:
Hi, thank you for this. Published v0.13.10 which contains this fix.

PR #615, opened 2017-06-05
Recursive ignore of .DS_store and Thumbs.db

kasp1:
Recursive ignore of .DS_store and Thumbs.db

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-.DS_Store
+**/.DS_Store
+**/Thumbs.db
 node_modules/
 dist/
 deploy/


kasp1:
Check this one before merging, looks like opinions are different https://stackoverflow.com/questions/18393498/gitignore-all-the-ds-store-files-in-every-folder-and-subfolder

rstoenescu:
Hi, don't really have to look into it. The .gitignore additions seem fine to me. Do we need anything else?

kasp1:
Should work like this

dhax:
This was already working recursively before just fine and .DS_Store is the de facto standard entry in like every .gitignore file out there. Just saying...

rstoenescu:
@kasp1 do we really need this change? I see that @dhax has a point and most repos on github simply specify `.DS_store` instead of `**/.DS_store` (same for `.Thumbs.db`). I don't own a Mac to test it out though.

kasp1:
It looks like it doesn't matter whether it's `**/.DS_store` or `.DS_store`, so it can be just merged, because it adds `.Thumbs.db`.

rstoenescu:
Ok, it's been already merged. Thanks for the confirmation.

PR #618, opened 2017-06-08
feature: add breakpoint specific col grow and add missing min-height to col auto

pdanpdan:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

col-{breakpoint}-grow specific classes

### DIFF
--- a/src/themes/core/flex.styl
+++ b/src/themes/core/flex.styl
@@ -31,6 +31,7 @@
 
 for $name, $size in $sizes
   .row > .col-{$name}
+  .row > .col-{$name}-grow
     width 100%
   .col-{$name}
     min-height 1px
@@ -117,8 +118,14 @@ for $name, $size in $sizes
       flex-basis 0
       flex-grow 1
     .col-{$name}-auto
+      min-height 1px
       width auto
       flex 0 0 auto
+    .col-{$name}-grow
+      min-height 1px
+      flex 1 1 auto
+    .row > .col-{$name}-grow
+      width auto
 
     for $i in (1..$grid-cols)
       .col-{$name}-{$i}


rstoenescu:
I believe that the `grow` doesn't quite makes sense to work with xs-, sm- etc prefixes. Not saying this isn't right, but also trying to keep the CSS footprint as low as possible. After the "Design Helpers > CSS flexbox" docs section will be written it will be more clear on what I mean.

pdanpdan:
Maybe I missed some details, but I can see no way to do this:
- 1 .row container with 2 cells inside
- you want them to be full width on <=sm
- you want them to fill the row on > sm, and have independent, not predefined sizes

What I can think of is .row>.col-sm-grow

rstoenescu:
```html
<div class="row">
  <div class="col">...</div> <!-- could use col-6 too -->
  <div class="col">...</div> <!-- could use col-6 too -->
</div>
```
By default rows are now "wrapping content". So you can define col-*-12 to make it spread whole width and next cols will go to next line.

```html
<div class="row">
  <div class="col-xs-12 col-sm-6">...</div> <!-- could use col-6 too -->
  <div class="col-xs-12 col-sm-6">...</div> <!-- could use col-6 too -->
</div>
```

Check path `#/css/grid-rows` on Quasar repo dev server for all it can do. Play with the width of window.

pdanpdan:
This would only create equal or predefined size splits.
For the moment breakpoint specific .col-auto is implemented, leaving .col-grow as an orphan.

rstoenescu:
You could also use:
```
<div class="row">
  <div class="col-xs-12 col-auto">...</div>
  <div class="col-xs-12 col">...</div>
</div>
```
This makes one each child div spread full width on smaller than `xs` devices and on sm+ first div has auto width while the second div will spread full width.

rstoenescu:
Also check "Variable width content" on the Q dev server page I mentioned.

pdanpdan:
Ok, I'll keep it in userland :)

rstoenescu:
Yes, but like I said, you don't really need it...

PR #683, opened 2017-07-02
Add color property to q-card-title and q-card-main 

jsfeldman:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
This pull request adds a color property to the card title and the card body so that they can be distinct colors. I noticed that this feature isn't discussed in the material design specifications (so it might have been taken out purposefully), but it was in v0.13 so I made the padding for colored title/main cards match the padding v0.13 had. 

I read through the contributing guide, but this is my first pull request to this project so let me know if there are any issues with the pull request or I need to do anything else.

### DIFF
--- a/dev/components/components/card.vue
+++ b/dev/components/components/card.vue
@@ -45,8 +45,18 @@
       </q-card>
 
       <q-card inline>
-        <q-card-title>
+        <q-card-title color="primary">
           Title
+        </q-card-title>
+        <q-card-main color="secondary">
+          {{lorem}}
+        </q-card-main>
+      </q-card>
+
+      <q-card inline>
+        <q-card-title color="primary">
+          Title
+          <span slot="subtitle">Subtitle</span>
         </q-card-title>
         <q-card-main color="secondary">
           {{lorem}}
--- a/dev/components/components/card.vue
+++ b/dev/components/components/card.vue
@@ -44,6 +44,16 @@
         </q-card-main>
       </q-card>
 
+      <q-card inline>
+        <q-card-title>
+          Title
+          <!--<span slot="subtitle">Subtitle</span> -->
+        </q-card-title>
+        <q-card-main color="secondary">
+          {{lorem}}
+        </q-card-main>
+      </q-card>
+
       <q-card inline color="secondary">
         <q-card-title>
           Title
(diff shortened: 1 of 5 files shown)


rstoenescu:
Hi,

I see this is your first PR on Quasar and I want to encourage you to continue on, but won't merge this one. Trying to get as close as possible to Material recommendations so that's why paddings were changed. Setting colors also is a much too trivial thing and won't add props for it just everywhere (also trying to keep footprint as low as possible).

Looking forward to your next contributions though! The PR code looks good!

-Razvan

jsfeldman:
Hello Razvan,

That makes a lot of sense. As I started working through the PR I thought that might have been your reason for taking it out. It's unfortunate they don't talk about multicolored cards in the specifications but perhaps they don't like the idea altogether.
Most of the styling to achieve it is pretty easy to add with separate css anyway, but one idea from the PR that I think might still be useful to add is the padding top I added to the subtitle in place of the 2rem line-height on the card title text.
I think this might be nice because otherwise if you're not using a subtitle you end up getting both the padding and the extra line-height space, which I think might be slightly too much. Also, if your card title ends up overflowing to two lines it looks a little awkward. Especially if you have a subtitle, because now there is more space between the title lines than there is between the title and subtitle. I noticed the same is true for list headers but that might be a separate topic. Unfortunately, the space between a title and body of a card without a subtitle and the space between card title lines aren't discussed in the spec either as far as I could see.

I noticed when I was writing the styling that way one issue I ran into is that the subtitle is rendered by vue regardless of there being any content in it, leading to extra padding in the title anyway. I got around it by doing this:
```
const subtitle = slots.subtitle ? [h('div', {staticClass: 'q-card-subtitle'}, slots.subtitle)] : []
return h(
  'div',
  data,
  [
    h('div', {staticClass: 'col column'}, [
      h('div', {staticClass: 'q-card-title'}, slots.default),
        ...subtitle
    ]),
    h('div', {staticClass: 'col-auto self-center q-card-title-extra'}, slots.right)
  ]
)
```
which isn't ideal style wise but there are other ways to handle it.

What do you think? Would that be useful to change?

Thanks,
Jonathan

PR #711, opened 2017-07-13
Feature: Add safe property (default false) to components prone to rec…

pdanpdan:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] Yes
- [ ] No

**Other information:**
There are some components where data is expected to originate from users.
Makes it so that in that components you need to explicitly require rendering as html.

### DIFF
--- a/src/components/chat/QChatMessage.vue
+++ b/src/components/chat/QChatMessage.vue
@@ -6,14 +6,20 @@
       'q-message-received': !sent
     }"
   >
-    <p v-if="label" class="q-message-label text-center" v-html="label"></p>
+    <template v-if="label">
+      <p v-if="safe" class="q-message-label text-center" v-html="label"></p>
+      <p v-else class="q-message-label text-center">{{label}}</p>
+    </template>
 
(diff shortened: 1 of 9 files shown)


spectrolite:
At a first glance, it seems like an appropriate and easy to use stop-gap solution until we get rid of most of these v-html for safer alternatives or enforce safe encoding in the right places. Thx @pdanpdan

spectrolite:
I thought about this one some more, and it feels pretty safe and simple to use. But I also feel it is too tempting for the dev to just "say it's safe" even though it isn't, just to get it working. In order to encourage safe use, maybe quasar could provide a help function to correctly sanitize Html output, and apply it by default to content that is declared as safe?

pdanpdan:
As far as I know, there is no way to sanitize html (and cover all cases). The only bulletproof way to be safe with user provided data is to write to textContent (so v-text/mustache).
Adding a safe option (default false) will make the framework treat content as unsafe by default and allow us to switch the behavior on request.

PR #720, opened 2017-07-16
Update key to use value as key

rettigd:
Vue 2.4 raises a warning when using a primative as the key

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Getting the following warning in Vue 2.4 when using selects [Vue warn]: Avoid using non-primitive value as key, use string/number value instead. Using the value instead of the object in the as the key resolves this issue

### DIFF
--- a/src/components/action-sheet/ActionSheet.vue
+++ b/src/components/action-sheet/ActionSheet.vue
@@ -15,7 +15,7 @@
           <div v-if="gallery" class="q-action-sheet-gallery row wrap flex-center">
             <div
               v-for="button in actions"
-              :key="button"
+              :key="JSON.stringify(button)"
               class="cursor-pointer relative-position column inline flex-center"
               @click="close(button.handler)"
               @keydown.enter="close(button.handler)"
@@ -32,7 +32,7 @@
           <q-list link v-else class="no-border">
             <q-item
               v-for="button in actions"
-              :key="button"
+              :key="JSON.stringify(button)"
               @click="close(button.handler)"
               @keydown.enter="close(button.handler)"
               tabindex="0"
@@ -70,7 +70,7 @@
         <div v-if="gallery" class="q-action-sheet-gallery row wrap flex-center">
           <div
             v-for="button in actions"
-            :key="button"
+            :key="JSON.stringify(button)"
             class="cursor-pointer relative-position column inline flex-center"
             @click="close(button.handler)"
             @keydown.enter="close(button.handler)"
@@ -87,7 +87,7 @@
         <q-list link v-else class="no-border">
           <q-item
             v-for="button in actions"
-            :key="button"
+            :key="JSON.stringify(button)"
             @click="close(button.handler)"
             @keydown.enter="close(button.handler)"
             :class="button.classes"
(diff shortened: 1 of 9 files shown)


abacram (approved):

pdanpdan:
Hi.
1. Do you think it's an expected user case to have multiple entries in options, with same value but different display options?
2. This does not solve the problem when value is a non string/number primitive (bool) or an object
Maybe it would be better to use JSON.stringify(opt) as key

rettigd:
1. I wouldn't expect this. Typically selects are defined as a key/value pair. I guess you could present the user multiple labels for the same value, but it doesn't seem like a best-practice to me - wouldn't you want to avoid duplicate values? Testing this use case: If select one item, the duplicate is also selected visually, but only one value is added to the model. Not sure what I expected.
2. Correct. If the value is an object, you get the same error. JSON.stringify resolves this (and doesn't impact #1). PR Updated. Thank you.

pdanpdan:
Do you think you can extend the PR to fix all \:key with possible non string/number values?
If not I can try to do it.

rettigd:
I probably won't get to it until late today or early tomorrow, but I can make a pass at it

rettigd:
Need a little more time...I was trying to use quasar-play to test my changes, but it suffers from the same issue and there seems to be other (existing?) warnings for undefined elements. In theory, this is a minor change, but I don't have a quick way to test all the components. In the meantime, I've updated the PR for review.

I can't help to wonder if some of these should be using index (they all use v-for). Probably marginally faster, but I don't know if there would be side effects. Thoughts?

rstoenescu:
Can only comment when I get back from vacation. Thanks for taking time for a PR!

rettigd:
After testing these, I had to change Stepper to use the index. I was getting the following warning while testing with quasar-play [Vue warn]: Error in render function: "TypeError: Converting circular structure to JSON" The PR is updated.

pdanpdan:
Ah, I didn't think of circular references :(
I'll take a look and maybe I can come up with something.
The problem with index key is that the key is used for deciding when to redraw an element. So if you have select options ["a", "b", "c"] and update them to ["x", "y"] you'll end up with "a" and "b" displayed.

huafu:
JSON will make problems with circular references, but also takes significant resources on huge lists.
I think it needs a (possibly) breaking change here, where the `key` would be `item.key || item.value`. 
Anyway, just dropping some thoughts here as I'm facing that issue now with a huge list...

smolinari:
@rettigd - FYI (and maybe you know this already), all the components needed to do basic testing while developing Quasar are in the `/dev` folder. So, when you are in the Quasar main directory with a console, you can start `quasar dev` to get a barebones versions of component examples, where you can see any changes you make directly. 😄  

Scott

rstoenescu:
Hmm, seems like there were two PRs for this. Merged the other one because it was the first I saw. Thanks for the efforts!

huafu:
FYI the other PR is #751

PR #802, opened 2017-08-14
Prevent QSelect's filter to propagate to parent QField

pdanpdan:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

When we have QField > QSelect[filter], the input from QSearch(filter) is published into QField's input, overriding the select. So in this case QField thought that it's child was the QSearch, not the QSelect.

### DIFF
--- a/dev/components/form/field.vue
+++ b/dev/components/form/field.vue
@@ -272,6 +272,31 @@
           <q-input class="col" v-model="model" stack-label="Stacked $ Label" suffix="#" prefix="@" :inline-count="10"/>
         </div>
       </q-field>
+
+      <q-field icon="cloud">
+        <q-select filter v-model="select" :options="selectListOptions" stack-label="Stack label with filter"></q-select>
+      </q-field>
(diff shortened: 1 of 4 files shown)


pdanpdan:
It's not ok. Need more work.

pdanpdan:
It looks like the first version was the right one

pdanpdan:
Do you know of a way to provide from a functional component? I tried it and it doesn't seem to work.

rstoenescu:
@pdanpdan Do a search on "functional: true" in the codebase. One difference is that you'll be using ".js" extension instead of ".vue".
There's a "but": don't think that the Vue provide API works for functional components. Might be worth a re-check.

pdanpdan:
It looks like it's not implemented, so I would leave it as it is now.

PR #819, opened 2017-08-21
Add date suffix (1st, 2nd, 3rd) token to formatDate()

dsl101:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

New token introduced in the date formatter string, so existing applications which use that token will not behave as expected.

Frequently I want to display nicely formatted 'human readable' dates, such as 18ᵗʰ August 2017. This was quite hard to do with the existing date formatter, so I've added a token - `^` which will be replaced by the appropriate 2-character superscript string (using the Unicode superscript character block). So, for example, you can now do:

    const nice = formatDate(new Date(), 'D^ MMM YYYY')  // 18ᵗʰ August 2017 on day of submitting

### DIFF
--- a/dev/components/index.vue
+++ b/dev/components/index.vue
@@ -5,7 +5,7 @@
         <img src="statics/quasar-logo.png">
       </div>
       <h4 class="uppercase">Date Ordinals:</h4>
-      <p v-for="date in dateList">{{date}}: {{ fmt(date, 'D^ MMMM YYYY') }}</p>
+      <p v-for="date in dateList">{{date}}: {{ fmt(date, 'Do MMMM YYYY') }}</p>
       <div
         class="list no-border"
         v-for="(category, title) in list"
--- a/dev/components/index.vue
+++ b/dev/components/index.vue
@@ -4,6 +4,8 @@
       <div class="row justify-center" style="padding-bottom: 25px">
         <img src="statics/quasar-logo.png">
       </div>
+      <h4 class="uppercase">Date Ordinals:</h4>
+      <p v-for="date in dateList">{{date}}: {{ fmt(date, 'D^ MMMM YYYY') }}</p>
       <div
         class="list no-border"
         v-for="(category, title) in list"
@@ -28,6 +30,8 @@
 
 <script>
 import pages from '../pages'
+import { date } from 'quasar'
+const { formatDate } = date
 
 let list = {}
 pages.map(page => page.slice(0, page.length - 4)).forEach(page => {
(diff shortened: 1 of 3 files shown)


dsl101:
BTW - the additions to the demo page probably aren't right - I just wanted a quick way to eyeball what I'd added. But I guess if it's going to be there, it should show more of the `formateDate()` options. And then what about the other helpers... And pretty soon it unravelled so I left it as is :).

rstoenescu (changes requested):
Hi,

Nice PR! Two comments though:

1. Can we make it inline with the other date libraries and use "o" instead of "^". Actually, take a look at this to get what I mean: https://date-fns.org/v1.28.5/docs/format -- notice tokens like "Mo", "Qo" and so on... I think it will be better to get in line with current major date libraries.
2. This will probably get merged (haven't decided yet) after the i18n work is finalized, as it might depend on it. For example for Romanian translation "Mo" should not contain the day endings ("st", "nd", "rd" etc), but this is out of the scope of current PR, so let's just focus on making it work like on point 1.

Thanks for contributing!

dsl101:
Great - I will have a look at that standard - I was originally trying to minimise potential conflict with creating new tokens, hence going for an unusual character. But if there are some standards out there, that's probably better. And I agree about internationalisation. I was trying to think if there was a 'formula' or other options object which could override the default English suffixes but (a) the English 'formula' isn't exactly simple, and (b) I have no idea what other languages use for date suffixes :). Happy to take a look in the context of the i18n stuff when it's ready.

rstoenescu:
Let's leave i18n out of this one for now and simply focus on adding those tokens. We'll deal with i18n later. Thanks again for taking care of this!

dsl101:
OK, I've added the 'Do' token, and also allowed the user to pass in a custom function to return a date suffix, so it could be used like this:

    methods: {
      fmt (date, format) {
        return formatDate(new Date(date), format, {dateSuffix: this.customSuffix})
      },
      customSuffix (d) {
        return d + '°'
      }
    }

The function expects a date (1-31) and should return the number _and_ suffix. The default is to provide the regular 1st, 2nd, 3rd, etc.

Did you want me to add the 'Mo', 'Qo' tokens as well? To be honest, I'm not sure where they'd be used but I guess for completeness they should be there?

spectrolite:
In french you would only need a suffix for the 1st day of the month (1er -for premier), all other days are without suffix.
The same rule is true in italian, and the suffix is the degree symbol as in : 1° maggio.

Please also note that months are never capitalised in dates (neither in fr nor in it), although they usually get Capitalized e.g. if they’re on their own on top of a calendar’s month view, but that’s more of an i18n issue.

dsl101:
I _think_ @rstoenescu decided to do this in a different way but I honestly can't remember now. But I think you can do everything you described by passing in a custom function to the code I original posted.

rstoenescu:
Hi,

Thanks for your work! Added Do and Qo, but in a slightly different way. Available in edge and future v0.14.7.

PR #859, opened 2017-09-05
feat: Inline Color Picker

altitudems:
This adds a simple Inline Color Picker Component. I imagine we will want to treat this similar to how the DateTime component is treated. I am open to any suggestions for improvement.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:
Requires the tinycolor2 package

- [ ] Yes
- [x] No

**Other information:**
screen shot 2017-09-05 at 11 25 00 am


altitudems:
If you get some time, let me know if you need anything changed for this to get accepted. Once I know the direction you want to take this I can implement it better with forms, etc. and write some docs. If we want to wait until there is a clearer path for community components, that is understandable.

rstoenescu:
This is such a great PR!! Will be part of v0.15. Please allow me some time before I merge it.

rstoenescu:
Can we avoid the additional dependency?

altitudems:
I'm sure we could, we would have to reimplement some of the math from tinycolor2.

How important is it to spit out color in the different formats? Is Hex + RGBA enough? Or do we need HSV, etc.?

altitudems:
I went ahead and removed tinycolor2, I had to add two color functions. I dropped HSL and the value/model is now a flat object: { h: 0, s: 0, v: 0, r: 0, g: 0, b: 0, a: 1, hex: '#000000' }

spectrolite:
Great job there.
But it’s really too bad you dropped HSL, which is probably the most intuitive colourspace for endusers.
If you need help with the math to put it back in, feel free to pm me on the forum.

altitudems:
@spectrolite Thanks for the feedback. I'm not sure how I feel about adding both HSL and HSV/B. I would rather choose one or the other and keep the value object flat for simplicity.

Some things to consider:

- HSB/HSV values cannot be present along with HSL values (H and S are not in common)
- The UI only shows RGBA and Hex values to the user, so user intuitiveness is not a factor
- If you need HSL values somewhere in your app you could still do a quick conversion yourself
- The math is not hard at all, and I can borrow that from somewhere else

Personally I don't care if we drop HSV for HSL or drop both. I think providing at least RGBA and Hex is enough.

PR #949, opened 2017-10-03
support bt prefixed icon fonts added

ahadrt:
needed to my individual fonts

its better to support all prefix font icons not only material and fa and limited some.

thankyou


rstoenescu:
Is this Black Tie font icons?

ahadrt:
bt prifix is for barteh fonts using for management and productivity science. without it we couldnt use ur nice framework quasar. we will publish this font  under MIT 
thankyou for merge pull request

rstoenescu:
Available in edge and future v0.14.5. Thanks for the PR!

PR #982, opened 2017-10-11
Makes Quasar Framework SSR friendly

codingfriend1:
Vue server side rendering throws errors if the window, document, Element, XHRRequest, navigator, etc... are referenced directly within the code. I have protected these accessors by first checking if `isServer` is false. isServer is a feature that checks if typeof window === 'undefined'.

Could probably use some better logic for platform.is detection while on the server, however Vue already comes with this `Vue.prototype.$isServer`.

This needs to be more thoroughly tested, but the changes should be minimal.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [] Yes
- [x] No

- [x] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar-framework.org/tree/master/source) (for faster update click on "Suggest an edit on GitHub" at bottom of page) or explained in the PR's description.


rstoenescu:
So basically why we're trying to do is make Quasar isomorphic. Quick question: what happens with Platform? Decisions are made based on what Platform contains. If we skip it then how can we be certain that pages look same rendered from SSR and from client-side?

codingfriend1:
That's a great question.

Browser globals may be referenced within methods that don't run until on the browser, but they shouldn't run in the root render. Not sure if we want to add a detection for the server and make decisions based off of that.

BTW I looked at the console.errors for the elements when running quasar dev and only noticed errors with these and my change. Not sure if those were in there before, I broke off of the dev branch.

qoptiongroup
QInputFrame
QAutocomplete

Platform should still largely stay the same. Of course some of the browser based settings won't be set. I've added

```
function getUserAgent () {
  if (isServer) { return 'isServer' }
  return (navigator.userAgent || navigator.vendor || window.opera).toLowerCase()
}
...

if (!isServer && window && window.process && window.process.versions && window.process.versions.electron) {
    browser.electron = true
  }
  else if (!isServer && document.location.href.indexOf('chrome-extension://') === 0) {
    browser.chromeExt = true
  }
  else if (!isServer && (window._cordovaNative || document.location.href.indexOf('http') !== 0)) {
    browser.cordova = true
  }

...

const Platform = {
  is: getPlatform(),
  has: {
    touch: (() => !isServer && (!!('ontouchstart' in document.documentElement) || window.navigator.msMaxTouchPoints > 0))()
  },
  within: {
    iframe: !isServer && window.self !== window.top
  }
}
```

It doesn't seem to break anything, thoughts?

codingfriend1:
There are more hydration errors. q-btn for one. I may try to setup a test environment and a new branch to see if I can fix them.

Would you mind making a new ssr branch for me?

codingfriend1:
We could potentially add a global `global.userAgent`. It would have to be set upon each request before running the server side render.

global.userAgent = req.headers['user-agent']

```
function getUserAgent () {
  return (global.userAgent || navigator.userAgent || navigator.vendor || window.opera).toLowerCase()
}
```

codingfriend1:
I made several improvements to get ssr working. Most of them had to do with changing paragraphs to divs and the others had to do with wrapping the default <slot></slot> in some form of container.

These files will all successfully render with SSR but will all have some sort of display bug, a momentary glitch between the way the server renders it and the client.

dev/components/form/size-test
dev/components/components/tooltip
dev/components/components/spinner
dev/components/components/stepper
dev/components/css/flex-columns
dev/components/css/flex-gutter
dev/components/css/flex-rows
dev/components/css/shadows
dev/components/form/field
dev/components/form/select
dev/components/components/button-dropdown

These will all ssr render, some with display glitches when userAgent is provided.

dev/components/components/data-table
dev/components/other/platform-detection
dev/components/components/data-table2
dev/components/form/size-test 
dev/components/components/scroll-area
dev/components/form/datetime

Only vue-transitions won't work in either case.

codingfriend1:
Made more progress, fixed most all hydration errors. The last couple errors are ui related. This has to do with text not appearing upon first ssr render. This can be solved by using platform.is v-shows instead of desktop-only, mobile-only, etc... classes.
http://localhost:8081/components/context-menu
http://localhost:8081/components/tooltip
http://localhost:8081/form/field

http://localhost:8081/components/back-to-top
The back to top component displays upon first ssr render

http://localhost:8081/components/stepper
The header is missing during the render. Not sure how to solve this one easily.

http://localhost:8081/test-layout/drawer
scroll out of place

Issue #244, opened 2016-12-12
DataTable - exporting to CSV, excel, ... feature


rstoenescu:

colinmollenhour:
IMO this is better to leave up to server-side, perhaps just adding a UI button to trigger the request.

Also note in my experience for one data table UI I often need to export multiple different grids. E.g. one might be parent rows only and another might be one-row-per-child with some parent columns duplicated, etc..

mesqueeb:
+1 for export

purell:
@colinmollenhour I'm totally agree with you. 

If we look at the Datatable.js, a lot of transfer time is consumed because of the csv, pdf, ... export plugins. So it's better to leave up the server do the conversion and call the route on a top-bar button of the QTable.

rstoenescu:
This will be left to an external Quasar plugin. There's going to be such a plugin system for v1.0 (after v0.15) that will allow the community to make their own Quasar plugins. Stay tuned.

purell:
Most of existing plugin I tried like [tableexport](https://tableexport.v4.travismclarke.com/#tableexport) use the whole table to produce the xls(x) file. I tried them with the quasar QTable v0.15 but it does not work as expected. The QTable only draw the `computedRows` property that is holding the rows **only** for the current page instead of all rows.  

Looking at the `jquery.datatable`, even if we are looking the page 1/50, the exported files contains the rows of all pages.

I think we will have to create a quasar plugin from scratch without help of third party library. 

PS : Do we want the exported file to keep the same row order and row filter as we see ?

PR #985, opened 2017-10-11
Add basic support for convert data (array<object>) to xlsx, csv...

up9cloud:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

#244 

TODO

1. [x] basic convert data to xlsx's format
2. [x] basic download file testcase.
3. [ ] add download button to dataTable
4. [ ] solve other platforms download issue.

### DIFF
--- a/dev/components/other/data-download.vue
+++ b/dev/components/other/data-download.vue
@@ -17,16 +17,22 @@
           {{v}}: {{JSON.stringify(headers[i])}}
         </li>
       </ul>
-      <p
-        class="caption"
-        v-for="ext in ['csv', 'xlsx', 'txt', 'html']"
-        :key="ext"
-      >
+      <p class="caption">
         <span class="desktop-only">Click</span>
         <span class="mobile-only">Tap</span>
-        to download {{ext}}.
-        <q-btn v-for="(label, i) in labels" :key="i" :label="label" @click="download(ext, i)" />
+        to download
+        <select v-model="ext">
+          <option disabled value="">(select)</option>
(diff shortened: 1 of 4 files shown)


nothingismagick:
This should be closed because the justifying issue was also closed.

#244

rstoenescu:
This would be a good app extension. Should not be closed.

nothingismagick:
ok - maybe @hawkeye64 will be bored soon. :D

nothingismagick:
For inspiration: https://tableconvert.com/documentation.html

edgar-koster:
I was reading along for something else when I saw this. What is the idea here for I have a working situation to export a q-table to excel where there the dataset of the table is used in the export. The export file uses the same functions I use in my q-table file. Only thing, that is enhanced too. So if the table is to use a specific defined function to display data, export does too. If the qtable is to hide the column (my own show/hide) then the export too doesn't have it. Sorting is also the same. Header row is filtered in the excel. I use exceljs for it.

rstoenescu:
Hi,

Awesome work!

However, this would be great for an App Extension, but definitely not for Core... it imports external dependencies (xlsx), which is a 100% no-go for Core UI. Only xlsx is almost 500KB minified, not to mention it also imports other packages as well. If we merge this into Core, then the full version of shipped Quasar would jump in size to at least 5x times more.

In any case, please do not close the PR.

rstoenescu:

PR #1115, opened 2017-11-10
Update table-body.js

alexdee2007:
Add Vue event @rowclick (Triggered when user clicks on a row in table)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

It would be nice to add this useful feature as it was in 0.14

### DIFF
--- a/src/components/table/table-body.js
+++ b/src/components/table/table-body.js
@@ -59,7 +59,7 @@ export default {
             ]))
           }
 
-          return h('tr', { key, 'class': { selected } }, child)
+          return h('tr', { key, 'class': { selected }, on: { click: this.emitRowClick.bind(null, row) } }, child)
         })
       }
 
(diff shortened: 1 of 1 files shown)


rstoenescu:
Hi,

Thanks for taking time for a PR!
In this revamp you can use a scoped slot for the row and add ANY events you wish. And not only on the row, on any column cell too. Basically you have total freedom. So I am going to close this as it would be kind of redundant and also a little confusing for devs as they would expect this event to be triggered even when using row scoped slot. Can we agree?

alexdee2007:
Hello Razvan, thanks for taking time and for your answer.
I totally agree with you. Of course i know about new scoped slots, and this is an excellent opportunity for developers. I just wanted to have some default behavior, so to say "out of the box", since this one is very often used.

P.S. Thank you for your work. This is great.  Quasar is the best js framework I've ever seen.

Issue #951, opened 2017-10-03
Autocomplete should close when tab or shift+tab is pressed


talski:
When you leave a field by pressing tab key, it will keep the results open, when it should have closed.

Also, there could be a option (select-on-tab) that when true, it will also set selected the focused item.

rstoenescu:
Available in next v0.15 Quasar beta.

Issue #761, opened 2017-07-29
Feature: Autofocusing filter inside of q-select


Spoja:
I've been digging around abit and couldnt find option to focus filter (QSearch) inside of QSelect component.
In most cases when you have QSelect with enabled filter you want it to be autofocused so you don't have to click twice ( first on QSelect and then on filter ) before you can start typing.
Im suggesting to add an option to QSelect which enables you to set that autofocus to filter, something like:
```
<q-select
  class="full-width"
  v-model="selection"
  :options="options"
  @change="changeOption()"
  filter
  autofocusfilter
/>

```
I made a workaround ( probably the messiest one since im pretty new to quasar and vue ) which looks like this:

```
<q-select
  ref="selectOption"
  class="full-width"
  v-model="selection"
  :options="options"
  @focus="selectOptionFocus()"
  @change="changeOption()"
  filter
/>
```
and method selectOptionFocus() :

```
selectOptionFocus () {
      this.$refs.selectOption.$children[0].$children[0].$children[0].focus()
}

```
Sorry if it already exists but i just couldn't find it.
On the other note i've noticed that @focus gets called twice ( if you confirm its a bug i can open new issue for it ), similar to this issue: https://github.com/quasarframework/quasar/issues/563

aldanor:
Ditto, having to click twice is questionable UX. Autofocusing filter should probably just be the default.

derekphilipau:
Additionally, it would be nice if keyboard events worked.  Pressing "enter" should select the top-most filtered result, while up/down cursor keys can navigate the list.

aldanor:
> Additionally, it would be nice if keyboard events worked. Pressing "enter" should select the top-most filtered result, while up/down cursor keys can navigate the list.

Yep. E.g. like [this](https://vuetifyjs.com/components/selects) (not the best example, but the one that works) where you can fill out a form full of selects using keyboard only.

danielcomcom:
I'll second this!

rstoenescu:
Will do.

rstoenescu:
`autofocus-filter` Boolean property available in edge and future v0.14.5. Navigating with keyboard is part of another ticket and will get into Quasar soon too.

PR #1186, opened 2017-11-26
[Feat] Improve QSelect

### LINKED ISSUE #903: [Request] QSelect add key events
It could be useful to handle key events to use keyboard on desktop version. 
On forms I would like to use keys to move between fields and also on a select (i.e. press enter to submit form, currently handled on simple qinput fields but not on qselect field, but also move between qselect values using up and down keys).

alexdee2007:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Some explanation of this proposed feature

When focused on select field (by press key "Tab" on previous field eg) then press key "Enter" will be show popover with list options. If there is a filter then it will be focused (assumed that user uses keyboard and dont want mouse click). Keyboard navigation when filter focused also available.

When popover with list options showing - if there is selected item(s) (field value) then viewport will scroll first encountered option in the middle viewport. 
And there is available navigation:

**Press key "Up" or "Down"** 
- navigate by list options. Viewport will scroll automatically.

**Pres key "Enter" on option:**
 - when select has multiple values - will be toggle value
 - when select has one value - this value will be selected, popover hide and this select field will be focused so user can "Tab" to the next field or reopen this

**Press key "Esc":**
 - popover hide and select field will be focused so user can "Tab" to next field or reopen this popover

**Press key "Tab":**
- popover hide and focus go to the next field (ordering by tabindex or by position in dom)

**Press keys "Shift+Tab":**
- popover hide and focus go to the previuos field (ordering by tabindex or by position in dom)

When user inputs  a filter then first option suggest to be "enter" (only on desktop platform)

**Additional info:**

* Add props list-max-height to select component (maxHeight popover nevertheless has 100vh so list options in any case will be within the screen area)
* Fix toggle show/hide popover when click on field
* Remove props autofocus-filter. Filter will be focus on desktop platform automatically
* Fix twice emitting blur and focus
* Dropdown icon rotate when open/hide popover
* Smooth popover reposition (popover jumps around the screen when input filter and popover located at the top of the field )
* Fix lose focus when click on chips and clear
* Add props autoOpen, loading, loadingLabel, noResultsLabel, noDataLabel and implements
* Mouse and keyboard have same selection
* Add selects labels loadingLabel, noResultsLabel, noDataLabel to i18n
* Allow prop multiple set limit
* Add prop limit to QOptionGroup

close #903


alexdee2007:
#761

pdanpdan (changes requested):

src/components/select/QSelect.vue
```diff
@@ -170,7 +178,11 @@ export default {
     autofocusFilter: Boolean,
     radio: Boolean,
     placeholder: String,
-    separator: Boolean
+    separator: Boolean,
+    scrollHeight: {
```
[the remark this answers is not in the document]
alexdee2007: Good point, thanks! listMaxHeight will be more appropriate.
I`m not sure about empty default. E.g. IE do not recognized scroll on viewport without setting "heigh" or "max-heigt" property...

src/components/select/QSelect.vue
```diff
@@ -248,6 +306,42 @@ export default {
...
+    __handleKeydown (e) {
+      if (this.$refs.popover.showing) {
+        switch (e.keyCode || e.which) {
+          case 38: // up
+            this.__moveCursor(-1, e)
+            break
+          case 40: // down
+            this.__moveCursor(1, e)
+            break
+          case 13: // enter
```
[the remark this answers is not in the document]
alexdee2007: Will add, thanks )

src/components/select/QSelect.vue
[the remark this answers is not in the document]
pdanpdan: Let's wait for Razvan's opinion on this also. (@rstoenescu)

src/components/select/QSelect.vue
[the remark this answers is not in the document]
alexdee2007: Hmm... Adding SPACE handler will make it impossible to enter a space in the filter. I revert this change. Let's wait for Razvan.

src/components/select/QSelect.vue
```diff
@@ -195,68 +210,119 @@ export default {
...
+      this.selectedIndex = this.multiple
+        ? this.options.findIndex(opt => this.value.includes(opt.value))
+        : this.options.findIndex(opt => this.value === opt.value)
+      this.filter && this.$q.platform.is.desktop ? this.$refs.filter.focus() : this.$refs.list.focus()
+      this.$nextTick(this.scrollToSelectedItem.bind(null, true))
+      this.focused = true
+    },
+    onHidePopover () {
+      this.selectedIndex = -1
+      this.$refs.input.$el.focus()
```
alexdee2007: Still not sure about this... The whole idea is that the field remains in focus until the user leaves it explicitly.
So the field remains in focus even after popover hide by click on body or option. Simply put, hiding popover does not cause field lose focus. Of course, except if you click (tap) on another field or element which can take focus. I wonder if this is the correct behavior?
I would appreciate hearing other opinion on this ...

frankbrussel: Great work on this! I haven't looked at the code (I'm UX designer, FE development is new to me), looking forward to having this working though. 

How I would expect it to work is quite similar to the q-search with autocomplete. However, with one small difference and there might be something strange with the QAutocomplete? 

1. The QSelect gets focus: this triggers the popover to show (autocomplete only shows after searching, in this case the options are already known and can be shown right away after receiving focus)

2a. Selecting an option in the popover (through clicking or through arrowUp/arrowDown and Enter): popover hides and field loses focus. The @selected can be used to trigger any other actions (such as setting the focus on another item)

2b. The user clicks somewhere else (also if it is nothing) or uses tab: the QSelect loses focus. I would expect the popover to disappear. The autocomplete does that the same for clicking somewhere else. However, it doesnt for using tab to move to the next tabindexed item, in that case the popover stays open (until a search is started in another QSearch or the page scrolls). I don't know whether this is intended or a bug. I would expect (here and in the QAutocomplete): using tab moves to a next field and thus the QSelect loses focus and the popover is hidden.

alexdee2007: Thanks for your comment! We need to clarify something here:
1. When QSelect field gets focus by click - this triggers the popover to show.
But when QSelect field gets focus by press "Tab" on previous field then field only displayed as "focused" (Highlighted).
In this way user can "Tab" to the next field.
Press key "Enter" (or click of course) on such focused field  - triggers the popover to show.
Although the field remains highlighted but physically focus goes to the options list (or to the "filter" input on desktop platform, if it is here).
By the way @blur not emitted in this case like now.

2a & 2b
 - click on option (not multiple)
 - press "Enter" on option (not multiple)
 - press "Esc"
 - click somewhere else 

all of this triggers the popover to hide. Physically focus go to the field so user can "Tab" to the next field.  Because if the field lose focus the user will not be able navigate by "Tab".

And of course you can use any @events to trigger any other actions.

It is also possible to implement prop like "auto-open" for popover opened when field get focus.
But notice - in order to navigate by "Tab" popover need to be closed (by press "Esc" e.g.)
To make another behavior, need to move the search filter from popover to the field inline . 
But i'm not sure if this is correct ("a la material")

P.S. Of course QAutocomplete need to be fixed for hide popover when field lose focus.

frankbrussel: I have selected some parts from your comment:

**1. But when QSelect field gets focus by press "Tab" on previous field then field only displayed as "focused" (Highlighted). In this way user can "Tab" to the next field.**
Even if the popover would show, the user could still tab to the next field, am I correct? Then, I'd say the popover could also show automatically when focus is gained through Tab, this saves the user from an extra step (pressing the Enter). On the other hand: the popover could obscure the next item. So I could definitely also understand no auto-open. (implementing a prop would definitely be an option too, of course)

W.r.t. 2a & 2b: **all of this triggers the popover to hide. Physically focus go to the field so user can "Tab" to the next field. Because if the field lose focus the user will not be able navigate by "Tab".** Ah yes, it is indeed very important that the user can continue the tab from the place they were. However, I would say it depends on the interaction: if the user presses Esc: indeed: the field should stay active. If the users clicks randomly in the body: I would say then the field is not active anymore. The user might question if everything is working correctly if they click in nowhere and the focus is not lost. Also it might feel a bit 'trapped' to them: as if they have to input something there, even if they don't want to. 

**But notice - in order to navigate by "Tab" popover need to be closed (by press "Esc" e.g.)
To make another behavior, need to move the search filter from popover to the field inline. (like in QAutocomplete e.g.)
But i'm not sure if this is correct ("a la material")**
Hmm, this is important indeed. I would suggest that pressing tab should be enough to go to the next field: in that case it is similar to how the user works with the QAutocomplete. If a user presses Tab they would want to go to the next field. Maybe they wouldn't know about escape and thus continue with the mouse as soon as they meet something they would want to skip. I understand that you also have to take the extra time for moving the search filter to the field inline into account. I'm just giving my ideas on the subject :). 

I have tried searching for material guidelines for how to deal with tab-use, couldn't find much information (likely because on mobile, there is no tab). So I think it is left open for interpretation. 

What I'm certain of: auto-opening and needing escape to close would be bothersome. Especially with multiple optional Qselects it would cause irritation. 

Let's say a user has two optional fields, wants to set the first one and skip the next one:
A: Tab, (auto-open), arrows, Enter (to choose), Tab (now on the second field, auto-open), Tab (5 actions)
B: Tab, Enter (to open), arrows, Enter (to choose), Tab (now on the second field), Tab (6 actions)
Option A (auto-open, perhaps as prop, & Tab can be used to move to the next field, even with the popover open) saves the user an extra action. Also it prevents hindrance for users familiar with using Tab but who don't use Escape as often. 

PS: and using the @selected on the field to focus the next field would save another action, in both cases :D. I know, these are all details, but I like make things as easy as possible, even the small actions.

alexdee2007: >This was your original comment about, right, for which you were mostly looking for comments?

Aha, exactly ) But other comments are also very useful

>Even if the popover would show, the user could still tab to the next field, am I correct?

Neah... as i said earlier  - in order to navigate by "Tab" popover need to be closed. And it's natural. When popover is open, physically the focus is on it. But the field remains highlighted (not physically focused).

>I have tried searching for material guidelines for how to deal with tab-use

I meant a little bit different. The material concept is the layers. The popover layer is higher and you work with it, everything else under it should not be in focus. Popover casts a shadow on the bottom layer, indicating that he is now in charge. Therefore, the filter input is now placed inside it.

On the other hand, in order for everything to work as you wrote - need to move the filter inside the field inline like in QAutocomplete. By the way, this will also solve another problem. When you start typing something into the search, the height of the popover changes, and it repositions, jumps around the screen... This happens when the field is somewhere at the bottom of the screen, and the popover does not fit below, and is located at the top of the field.

I can suggest looking at the current version of this PR )

frankbrussel: Ah yes, your link makes it indeed very clear. Especially the select with filter is really nice, great addition! I'll be using it as soon as it becomes available. 

- something I noted is that on the colored versions: it is not very clear whether they are in focus or not. (With me: nothing changes at all). Or perhaps this is something still left to do. 

- The tab key currently gives some unexpected behaviour when used within anything with checkboxes or radio buttons: to reproduce: click on for example the multiple selection, press tab a few times, then Enter. It seems that the images of the checkbox/radio button are also part of the tabindex. However, this gives confusing behaviour to what is the currently selected item: the one with the arrows or the one with Tab.

- I found myself using Tab a few times with an open popover, to switch to the next field (also without thinking). However, the Tab then doesn't always go to a logical place (sometimes for instance to the secure connection icon in the addressbar). I think that using the tab to move to a next field would fit the users mental flow: I fill this in, go to the next one, skip this: I'll go to the next one. First having to close the popover would be okay looking at it from the code: you have opened something and first need to close it (as with brackets etc.). Most users are not that literal, I think they would think: I just want to go the next one and they'll expect the popover to close automatically when that field is not active anymore. I think to them, the field and the popover would be one element (in their mental model), not two separate entities. Especially on PC/laptop. On a smartphone, if the select menu would be a menu on top of everything else, and everything else faded out, then it is more clear that you first need to close it. In that case you wouldn't use a Tab key anyway. And for PC/laptop pages, the popover is not a fullscreen, everything else fades out, type of menu (nor do I think it should be). It is an element in a larger page. I hope I am clear with what I mean :). Just my 2 cents, trying ot help find what would be ideal. I think in any way there are very nice improvements!

alexdee2007: > select with filter is really nice, great addition

It is not my addition, actually this feature already implemented in current release e.g. (0.14.7)

> on the colored versions: it is not very clear whether they are in focus or not

Yes it is. This happens when "inverted" option is set on QSelect. This is native behavior and can be solved by custom css.

> It seems that the images of the checkbox/radio button are also part of the tabindex

Сheckbox/radio are focusable (have a tabindex). This can be fixed by adding to them props "tabindex" with value "-1". This should be done in the QSelect component itself. I'll try what can be done...

>However, the Tab then doesn't always go to a logical place

Of course) Because at this time the focus is on popover which physically located at the bottom of the document (at the end of DOM). I repeat myself: To make another behavior, need to move the search filter from popover to the field inline. (like in QAutocomplete e.g.) But i'm not sure that @rstoenescu will approve this.  As an option, may be to add a property such as "inline-filter" or somthing like this...

alexdee2007: >PPS: I searched, and for the QAutocomplete, there is already a bug report: #951

I just added PR #1238

frankbrussel:
> select with filter is really nice, great addition
> It is not my addition, actually this feature already implemented in current release e.g. (0.14.7)

Ah yes, I noticed now, oops. Well, thanks for making me notice it then :)

> on the colored versions: it is not very clear whether they are in focus or not
> Yes it is. This happens when "inverted" option is set on QSelect. This is native behavior and can be solved by custom css.

Ah oké. I would suggest having something preset, perhaps to give them the same appearance as the QButtons. That way the out of the box functionality will also be clear. 

> It seems that the images of the checkbox/radio button are also part of the tabindex
Сheckbox/radio are focusable (have a tabindex). This can be fixed by adding to them props "tabindex" with value "-1". This should be done in the QSelect component itself. I'll try what can be done...
Great :). I wish I could help more with creating it also, at the moment I'm still learning to use Vue/Quasar, I think I would make a bit of a mess out of it. 

> However, the Tab then doesn't always go to a logical place
> Of course) Because at this time the focus is on popover which physically located at the bottom of the document (at the end of DOM). I repeat myself: To make another behavior, need to move the search filter from popover to the field inline. (like in QAutocomplete e.g.) But i'm not sure that @rstoenescu will approve this. As an option, may be to add a property such as "inline-filter" or somthing like this...
Ah now I understand better why that happens. Yes, perhaps @rstoenescu could add his ideas and opinions :)

src/components/select/QSelect.vue
frankbrussel: > select with filter is really nice, great addition
> It is not my addition, actually this feature already implemented in current release e.g. (0.14.7)

Ah yes, I noticed now, oops. Well, thanks for making me notice it then :)

> on the colored versions: it is not very clear whether they are in focus or not
> Yes it is. This happens when "inverted" option is set on QSelect. This is native behavior and can be solved by custom css.

Ah oké. I would suggest having something preset, perhaps to give them the same appearance as the QButtons. That way the out of the box functionality will also be clear. 

> It seems that the images of the checkbox/radio button are also part of the tabindex
Сheckbox/radio are focusable (have a tabindex). This can be fixed by adding to them props "tabindex" with value "-1". This should be done in the QSelect component itself. I'll try what can be done...
Great :). I wish I could help more with creating it also, at the moment I'm still learning to use Vue/Quasar, I think I would make a bit of a mess out of it. 

> However, the Tab then doesn't always go to a logical place
> Of course) Because at this time the focus is on popover which physically located at the bottom of the document (at the end of DOM). I repeat myself: To make another behavior, need to move the search filter from popover to the field inline. (like in QAutocomplete e.g.) But i'm not sure that @rstoenescu will approve this. As an option, may be to add a property such as "inline-filter" or somthing like this...
Ah now I understand better why that happens. Yes, perhaps @rstoenescu could add his ideas and opinions :)

alexdee2007:
@frankbrussel 
I made some changes and fixes. Pressing Tab now processed when the popover is open. Also added prop "auto-open". You can toggle it at the top of the page. Pls check the following link:
https://www.technofavorit.com.ua/quasar/#/form/select
Do not forget to refresh the page (F5) if there is a cached prevous version. Is there correct behavior as you described?

pdanpdan:
It looks very nice and functional.
Two things
1. Focus: check it on the selects with chips and with clear button:
  - open the select, select a value (or use a select with a value already set)
  - now click on remove chip or on clear
  - the select stays selected (correct) but the popup looses focus
2. Mouse/keyboard double navigation
  - you should use the same color for both
  - when mouse hovers over an option you should also change the keyboard focus to that option

alexdee2007:
>Two things...

Thanks for yout comment! I'll take it all into account, work is still in progress...

alexdee2007:
@pdanpdan 
@frankbrussel 

I made some changes:
* dropdown icon rotate when open/hide popover
* fix popover jumps around the screen when input filter and popover located at the top of the field (Still thinking about this. For now just removed reposition() from onInputFilter())
* fix lose focus when click on chips and clear
* add props loading, loadingLabel, noResultsLabel, noDataLabel and implements
* mouse and keyboard have same selection

Pls check new version at the following link:
https://www.technofavorit.com.ua/quasar/#/form/select

frankbrussel:
@alexdee2007 Looks like Christmas has come early this year! Yes, this is indeed the behavior as I intended, this is all I could hope for. I have checked it and it really works very smooth, both with the keyboard, the mouse and while switching between the two. Kudos for the auto-open also, I'm really glad that that has been implemented too. Great work, thanks!

alexdee2007:
I made some more changes, and I think that's all. Let's wait for @rstoenescu review this...

alexdee2007:
I made some more changes, and I think that's all. Let's wait for @rstoenescu review this...

alexdee2007:
So i made some changes, and I think that's all. Let's wait for @rstoenescu review this...

pdanpdan:
It looks good (I only checked functionality). I'll have a look on the code also.

1. Please move focusable prop to mixins/option and change it to noFocus: Boolean (it's easier to use), and also add a computed properties
```javascript
focusable () {
  return !this.noFocus && !this.disable
},
tabIndex () {
  return this.focusable ? 0 : null
}
```
and then use these 2 computed properties in QCheckbox, QRadio and QToggle
2. Use the color set for select to highlight the selected option in the list

alexdee2007:
> Please move focusable prop to mixins/option...

You are reading my mind ) I also thought about something like this ...

> Use the color set for select to highlight the selected option in the list

Do you mean instead of  class "text-primary" in non-multiple select?

pdanpdan:
Yes, instead of primary

alexdee2007:
@pdanpdan 
All done

wc-matteo:
These improvements would be really nice! Let's hope they get merged soon!

_wink wink nudge nudge_

dietrichg:
Can we also have an option to filter if the user starts typing and the QSelect is focused?
Much how the native select works where you can type to select available options in the list without having to drill into it. This helps for data-entry focused people who are familiar with the available options already

PR #1262, opened 2017-12-08
[Fix] QDialogSelect catch <-> then

alexdee2007:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Fix Promise chain. The .catch() is catching the error and not producing a new error. It has essentially "handled" the error, and the following .then() handler is free to be called.  So when click on "Cancel" in  dialog this caused emit value 'undefined'.


rstoenescu:
Thanks!

PR #1268, opened 2017-12-10
Allow autofocus prop to also select the text Closes #1250

### LINKED ISSUE #1250: Add autoselect prop to q-search component
Just like `autofocus` prop but `autoselect` also selects the whole text.

If both are provided, the input shall first be _focused_ and then its content _selected_, so next `keydown` replaces the whole value by the new one.

I can send a PR if the idea gets approved.

Frondor:
Closes #1250 
`autoselect` functionality. Can be used in conjunction with `autofocus`.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

We may need to include documentation for this prop in v0.15 docs, right?

### DIFF
--- a/src/mixins/input.js
+++ b/src/mixins/input.js
@@ -68,12 +68,10 @@ export default {
   mounted () {
     this.$nextTick(() => {
       const input = this.$refs.input
-      if (input) {
+      if (this.autofocus && input) {
+        input.focus()
         if (this.autofocus === 'select') {
           input.select()
-        }
-        if (this.autofocus) {
-          input.focus()
         }
       }
     })
--- a/src/mixins/input.js
+++ b/src/mixins/input.js
@@ -1,6 +1,7 @@
 export default {
   props: {
     autofocus: Boolean,
+    autoselect: Boolean,
     name: String,
     maxLength: [Number, String],
     maxHeight: Number,
@@ -68,8 +69,13 @@ export default {
   mounted () {
     this.$nextTick(() => {
       const input = this.$refs.input
-      if (this.autofocus && input) {
-        input.focus()
+      if (input) {
+        if (this.autoselect) {
+          input.select()
+        }
+        if (this.autofocus) {
+          input.focus()
+        }
       }
     })
   },
--- a/src/mixins/input.js
+++ b/src/mixins/input.js
@@ -1,7 +1,6 @@
 export default {
   props: {
-    autofocus: Boolean,
-    autoselect: Boolean,
+    autofocus: [Boolean, String],
     name: String,
     maxLength: [Number, String],
     maxHeight: Number,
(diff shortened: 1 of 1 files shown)


src/mixins/input.js
```diff
@@ -68,8 +69,13 @@ export default {
   mounted () {
     this.$nextTick(() => {
       const input = this.$refs.input
-      if (this.autofocus && input) {
-        input.focus()
+      if (input) {
+        if (this.autoselect) {
+          input.select()
+        }
+        if (this.autofocus) {
```
pdanpdan: this should be an else if

pdanpdan (changes requested):

src/mixins/input.js
alexdee2007: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/select
> Calling element.select() will not necessarily focus the input, so it is often used with HTMLElement.focus().

I think there is right logic, if we want to use both "autofocus" and "autoselect"...

pdanpdan: In this case it should be:
```javascript
if (this.autoselect) {
  input.select()
  input.autofocus()
}
else if (this.autofocus) {
  input.autofocus()
}
```

or

```javascript
if (this.autoselect) {
  input.select()
}
if (this.autoselect || this.autofocus) {
  input.autofocus()
}
```

as it makes no sense to select but not to focus.

alexdee2007: >In this case it should be..

Yes, it makes sense.

>input.autofocus()

Probably you meant input.focus() ? 

Also can change type of props "autofocus" to [String, Boolean], and check if it is equal to "select" or true...

pdanpdan: sorry, .focus ()

alexdee2007: So i suggest **Line 3** replace to:
```
autofocus: [String, Boolean],
```

And after **Line 72** add:
```
if (this.autofocus === 'select') {
  input.select()
}
```

Frondor: I realized that not every time you need `autoselect`, you also need `autofocus`. So that `else if` block won't work well.
Consider the case when you are on a mobile and you don't want the keyboard to popup because you don't need the user to input anything, but press a "Copy this" button next to the input with its text selected. Or in desktop you simply let the user press `ctrl + c`.
They are 2 different functionalities, that's why we can't use ´else if´ here.
Regarding to `this.autofocus === 'select'`, I agree

src/mixins/input.js
```diff
@@ -68,8 +68,13 @@ export default {
   mounted () {
     this.$nextTick(() => {
       const input = this.$refs.input
-      if (this.autofocus && input) {
-        input.focus()
+      if (input) {
```
alexdee2007: I think the whole block should look like this:
```
if (this.autofocus && input) {
  input.focus()
  if (this.autofocus === 'select') {
    input.select()
  }
}
```

Frondor: Done

rstoenescu:
Thank you for contributing! This goes into v0.15.

PR #1282, opened 2017-12-12
[Fix] QAutocompete show popover when changes v-model value

alexdee2007:
When changes v-model value this causes show popover because of watcher. I propose to use event listener on "input" instead of watcher. Checking "enterKey" is no longer needed.
Add prop triggerOnFocus

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No


pdanpdan:
Hi. Can you explain when do you find this behavior, because I cannot reproduce it?

alexdee2007:
Hello. There are many options.
For example, onfocus trigger function that fills the field with some data, based on other fields. 
Or an translator - by press Tab onblur from previous field fills this field with translated value... 
I personally experienced this problem during development.
In any case, I believe that for autocomplete the more correct method is to use input, rather than watch value. It must respond to input by the user, and not to some internal methods.

Other option - you can have autofocus on this field. When open the form or flipping through is filled from the database. and popover will be showing...

pdanpdan:
I can reproduce the sample case (fill on focus).
The problem is your solution is always one letter behind the input text.
And to be honest I find the current behavior to be the expected one (like if I fill an autocomplete on focus I would expect it to open with suggestions)

alexdee2007:
>The problem is your solution is always one letter behind the input text.

Not sure what you mean... "input" payload is the whole value, not only pressed key.  Besides it is not matter because this event call __delayTrigger()

> And to be honest I find the current behavior to be the expected one (like if I fill an autocomplete on focus I would expect it to open with suggestions)

I totally disagree.  The problem is that it's not you who filled the field. The field is filled with value behind the scene. And it's confusing when popover show...
Also you can leave this field by pressing "TAB". Considering that everything is correctly filled in. And when you're already focused on the next field, a popover will open in a while. This is absolutely not expected behavior... If we do not understand each other, I can provide jsfiddle with that case...

rstoenescu:
Hi, Thank you for contributing, but this is now obsolete.

PR #1293, opened 2017-12-13
close #1274: Don't hide popover that has child popover on click in child

### LINKED ISSUE #1274: autocomplete input inside q-popover having issue
### Software version

Quasar:0.14.7
OS:
Node:
NPM:
Browsers:chrome
iOS:
Android:
Any other software related to your bug:

### What did you get as the error?
Placing autocomplete input inside q-popover makes parent popover disappear after autocomplete list pop up.

### What were you expecting?
screens with multiple popover

### What steps did you take, to get the error?
```
<q-popover v-model="toggleNewOrderPopover" anchor="bottom right" self="top right" fit>
        <div class="row">
                <div class="col-12 pd-20">
                        <q-toggle v-model="downline" label="Order for downline"/>
                          <div class="without-downline" v-if="!downline">
                                <q-search v-model="terms"
                                                      placeholder="Search IBO ID / Name / IC"
                                                      :after="[{icon: 'chevron_right', handler () {}}]"
                                                      @change="reset">
                                                <q-autocomplete @search="search" @selected="selected"/>
                              </q-search>
                            </div>
                       </div>
             </div>
</q-popover>
```

popoverr-1
popoverr-2
popoverr-3

pdanpdan:
- also refocus QAutocomplete input element after click in option (so the behavior is the same as on enter)
- consider an anchorEl for popover if it's q-if-inner
- don't use capture for __bodyHide, so we can stop propagation

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**
close #1274

### DIFF
--- a/src/components/popover/QPopover.js
+++ b/src/components/popover/QPopover.js
@@ -103,16 +103,18 @@ export default {
 
       clearTimeout(this.timer)
       this.timer = setTimeout(() => {
-        document.body.addEventListener('click', this.__bodyHide, true)
-        document.body.addEventListener('touchstart', this.__bodyHide, true)
+        document.body.addEventListener('click', this.__bodyHide)
         this.showPromise && this.showPromiseResolve()
       }, 0)
     },
     __bodyHide (evt) {
       if (
-        evt && evt.target &&
-        (this.$el.contains(evt.target) || this.anchorEl.contains(evt.target))
+        evt && (
+          evt.skipPopup ||
+          (evt.target && (this.$el.contains(evt.target) || this.anchorEl.contains(evt.target)))
+        )
       ) {
+        evt.skipPopup = true
         return
       }
 
--- a/dev/components/components/popover.vue
+++ b/dev/components/components/popover.vue
@@ -9,7 +9,7 @@
 
       <div>
         <q-toggle v-model="toggle" class="z-max fixed-top" />
-        <q-btn color="primary" icon="assignment">
+        <q-btn color="primary" icon="assignment" class="q-ma-xs">
           <q-popover v-model="toggle" ref="popover1">
             <q-list link separator class="scroll" style="min-width: 100px">
               <q-item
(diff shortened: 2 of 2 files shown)


src/components/autocomplete/QAutocomplete.js
```diff
@@ -148,6 +149,10 @@ export default {
       this.searchId = ''
     },
     setValue (result) {
+      this.enterKey = true
+      if (this.inputEl && this.__input && !this.__input.hasFocus()) {
+        this.inputEl.focus()
```
alexdee2007: Hmm... Not sure if it useful. When you working with mouse or tuch it does not matter...
It only matters when you work with the keyboard... And there everything is ok...
Besides, double assignment **this.enterKey = true** when Enter press

src/components/popover/QPopover.js
```diff
@@ -74,7 +74,7 @@ export default {
     this.$nextTick(() => {
       this.anchorEl = this.$el.parentNode
       this.anchorEl.removeChild(this.$el)
-      if (this.anchorEl.classList.contains('q-btn-inner')) {
+      if (this.anchorEl.classList.contains('q-btn-inner') || this.anchorEl.classList.contains('q-if-inner')) {
```
alexdee2007: This is really a very useful addition!

src/components/autocomplete/QAutocomplete.js
pdanpdan: Fixed duplicate this.enterKey = true.
It matters to have it set because it blocks reopening the popover with the list when the value is set in parent.

alexdee2007: >when the value is set in parent.

Can you explain more detail? I cannot reproduce it... Just wondering is #1282 not fixes this problem?

pdanpdan: The idea is like this:
- parent component of QAutocomplete has a shadow property that keeps the value as `val` - this is updated according to debouncing and some delays for processing
- if we listen for input from the parent we're going to get in `val` the value before the last key press 
- if we listen for changes in `val` we get the correct value every time, but that value we are sometimes updating from the autocomplete, so we want to skip showing the options in that case

alexdee2007: Well, its clear ) But we can use target.value from input event? Then there will be a lot of redundant code and checks? I'll play around with this)
I still think it is improperly to use a watcher for autocomplete...

PR #1346, opened 2017-12-21
[Fix] Uncaught (in promise) Error when change route

alexdee2007:
Fix Uncaught (in promise) Error when change route and have disabled some components

- [x Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/fab/QFab.js
+++ b/src/components/fab/QFab.js
@@ -21,7 +21,7 @@ export default {
   },
   watch: {
     $route () {
-      this.hide()
+      this.showing && this.hide()
     }
   },
   created () {
--- a/src/components/layout/QLayoutDrawer.js
+++ b/src/components/layout/QLayoutDrawer.js
@@ -110,11 +110,11 @@ export default {
     },
     $route () {
       if (this.mobileOpened) {
-        this.hide()
+        this.showing && this.hide()
         return
       }
       if (this.onScreenOverlay) {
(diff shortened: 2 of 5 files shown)


rstoenescu:
Hi,

Thank you for contributing!
I'm a little confused, becase the "hide()" method returns early if "showing" is false. How can I reproduce the error you got? I want to understand why it's happening in the first place.

-Razvan

alexdee2007:
Hello!
For example we have disabled select (which have disabled popover).
When changed route then appears Uncaught (in promise) Error.
Here is a code:
src/mixins/model-toggle.js


```
  ... 
82   hide (evt) {
83     if (this.disable) {
84        return Promise.reject(new Error())         //ERROR APPEARS WHEN CHANGE ROUTE
85      }
86
87     if (!this.showing) {
88        return this.hidePromise || Promise.resolve(evt)
89    }
  ...
```
Of course, here we can also swap out checks for this.disable and this.showing. Check for this.showing should be in the beginning of method. This will also solve the problem...

rstoenescu:
Thanks for the details. I know see what is wrong. Pushed a commit. Will update quasar-future tomorrow and it will contain this fix.
Also thanks for reporting and offering a PR!

Issue #1383, opened 2018-01-08
QTd scope-slot props not available for QBtn click event callback method


realcarbonneau:
### Software version

Quasar: 0.15
OS:
Node:
NPM:
Browsers:
iOS:
Android:
Any other software related to your bug:

### What did you get as the error?
When using a QBtn click event callback method within a QTd, the slot-scope props are not available as parameters, nor in the scope.  Other alternatives, such as inline javascript in the @click="" or forcing props cause the loss of the done() callback function for the QBtn. 

The errors displayed in the console log are as follows for others doing searches on this:
Error in event handler for "click": "TypeError: Cannot read property 'row' of undefined"
TypeError: Cannot read property 'row' of undefined

This is a typical use case to have buttons within a table cell wich require access to the row information.
```
<q-td slot="body-cell-actions" slot-scope="props" :props="props">
  <q-btn icon="edit" @click="edit" />
</q-td>
```
```
<script>
  export default {
    methods: {
      edit (event, done) {
        // Some various processing delay with loading spinner
        done()
        this.$router.push({ path: '/edit/' + this.props.row.id })
        // this.props and props are undefined
      }
    }
  }
</script>
```

Workarounds, however, in these workarounds, the QBtn done() callback function is not available

Workaround 1:
```
<q-btn icon="edit" @click="this.$router.push({ path: '/edit/' + props.row.id })" />
```

Workaround 2:
```
<q-btn icon="edit" @click="edit(props)" />
```
```
<script>
  export default {
    methods: {
      edit (props) {
        this.$router.push({ path: '/edit/' + props.row.id })
      }
    }
  }
</script>
```

### What were you expecting?
To have access to the slot-scope props in the QBtn click event callback method

rstoenescu:
Hi, you need to first understand how Vue works and the scope of variables in both Vue components & ES6.

Here is what you want to achieve:
```
<q-td slot="body-cell-actions" slot-scope="props" :props="props">
  <q-btn loader icon="edit" @click="($event, done) => { edit(done, props) }" />
</q-td>
```
```
<script>
  export default {
    methods: {
      edit (done, props) {
        // Some various processing delay with loading spinner
        done()
        // then
        this.$router.push({ path: '/edit/' + props.row.id })
      }
    }
  }
</script>
```

Explanation: you need to access `props`, so creating an anonymous arrow function (acting as proxy) to feed the click event in the scope where `props` is defined.

[...]

PR #1406, opened 2018-01-12
Demo - Table Cell Button with Loader

realcarbonneau:
I have added an example of a table cell button with loader in the table demo.  This results from @rstoenescu answer to my issue #1383.  I can add these details to the documentation also.

tablecellclickdemo


rstoenescu:
Yes, will need to add this to the docs page too. But the doc page is non-existent at the moment. Thanks for taking time for a PR!

realcarbonneau:
Ok, tell me when it is ready and I will add what I can.

PR #1408, opened 2018-01-12
QBtnDropdown Disable Bug Fixes

realcarbonneau:
This adds "disable" test cases to the QBtnDropdown demo page and corrects the following bugs:
1) Non-split QBtnDropdown still does popup when in disable state
2) Split QBtnDropdown right dropdown sections is not in a faded color

In the fix for 2, I also removed the redundant check for disable state since this click is no longer possible in disable state.

Video of bugs
qbtndropdowndisabledbugs

Video of the fixed version
qbtndropdowndisabledfixes


rstoenescu:
Thanks! Nice finds for the PRs!

PR #1411, opened 2018-01-12
QBtnToggleGroup disable bug fix

realcarbonneau:
This adds "disable" test cases to the QBtnToggleGroup demo page and fixes the bug that the buttons where not disabled.  See the videos below:

Disable bug
qbtntogglegroup-disablebug

Disable fixed
qbtntogglegroup-disabletests


rstoenescu:
Good catch! Thanks for the PR!

PR #1412, opened 2018-01-13
QBtn Disable Active Effect Fix

realcarbonneau:
When the QBtn is "disable", the active lift shadow effect was still occurring.  This fixes it and adds two pushbuttons to the demo tests.  See the videos below.

Bug
qbtn-disable-raiseonactive-bug

Fix
qbtn-disable-raiseonactive-fix


realcarbonneau:
There is no raise effect on button clicks/active for iOS, so this bug does not apply to iOS

PR #1419, opened 2018-01-14
QBtn Round Push Active - Fix

realcarbonneau:
A Round Push QBtn was incorrectly displaying a raised shadow on click.  The active css code for q-btn-push was redundant because q-btn-push always also include the q-btn class wich manages the push. This fixes the problem, see the 2 videos below.

Round Push QBtn Active Raise bug
qbtnround-push-active-bug

Round Push QBtn Active Fix
qbtnround-push-active-fix


rstoenescu:
Thanks!

PR #1421, opened 2018-01-15
QChat no avatar fix #1420

### LINKED ISSUE #1420: q-chat-message not showing if avatar is omitted
If I do not supply a url to to avatar in q-chat-message the whole bubble dissapears - is there any way around this ?

realcarbonneau:
As per issue #1420, a QChat message without an avatar dissapears.  Thus, the component cannot be used without avatar images.  

This fix permits the use of the QChat without avatar images.  However, all messages must have avatars to keep them aligned. I did not add support for a mix of messages with and without avatar images because the user should plan for placeholder avatar images for those users without avatars.

This fix also adds a demo test of the QChat without avatar images.

Video of no avatar bug
qchatnoavatarbug

Video of fix
qchatnoavatarfix

### DIFF
--- a/dev/components/components/chat.vue
+++ b/dev/components/components/chat.vue
@@ -1,8 +1,8 @@
 <template>
   <div>  
     <div class="layout-padding" style="max-width: 500px;">
-      <h4>Chat with avatar section</h4>
-      <p>To mix messages with avatar and without in the same thread, use a placeholder avatar image.</p>
+      <h4>Chat with avatar</h4>
+      <p>To mix messages with avatar and without avatar in the same thread, use a placeholder avatar image.</p>
       <q-chat-message
         v-for="(msg, index) in messages"
         :key="index"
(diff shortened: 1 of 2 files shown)


realcarbonneau:
Another fix coming on this...

realcarbonneau:
These are more fixes for the QChat:
Fixed v-if slot section: it needed to be adjusted for the new structure
Fixed avatar named slot: Was never working, it is now implemented.  I also added a test case in the demo
Fixed v-for keys that were not unique:  The v-for keys where the label and the message text, which can neer be guaranteed to be unique, there were already errors in the console for non-unique keys.

See the video below:
qchatavatarmorefixes

realcarbonneau:
BTW, I wasn't sure if the original broken avatar slot was a mistake or by design.  Tell me if you want it removed.

rstoenescu:
Good work!

PR #1429, opened 2018-01-16
QBtnToggle Support for a Text Toggle Color

realcarbonneau:
A text toggle color is required for the QBtnToggle to support certain scenarios, such as toggling from a "Yellow Button" & "Blue Text" to "Blue Button" & "Yellow Text".  In addition, the text toggle color can now be specified per button in the options array.  I also added the test cases.

qbtntoggle-texttogglecolor

### DIFF
--- a/dev/components/components/button-toggle.vue
+++ b/dev/components/components/button-toggle.vue
@@ -66,12 +66,34 @@
                   ]"
                 />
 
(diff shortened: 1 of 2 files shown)


realcarbonneau:
I also set the required for toggleColor and set it to 'primary' by default to simplify it's most basic use.  This may also help novice users to use it more easily for a first try.

rstoenescu:
Awesome!

PR #1442, opened 2018-01-17
QDateTime+QColor: Explicit Popover/Modal props and cleanup

realcarbonneau:
This adds the flexibility to explicitly require a popover or modal QDateTime, thus overriding the default behavior when desired.

Minor cleanups included:
-Consolidated the isPopover/usingPopover into one computed property and made the reference to the global `this.$q.platform` variable reactive
-Replaced all the `this.value === 0 || this.value` with isValid from date utils.  This is cleaner, more general and avoids code duplication.

See the video below, the last part of the video is to show that the other scenarios continue working as is.

qdatetime-explicitpopovermodal

### DIFF
--- a/src/components/datetime/QDatetime.js
+++ b/src/components/datetime/QDatetime.js
@@ -1,11 +1,12 @@
 import FrameMixin from '../../mixins/input-frame'
+import DisplayModeMixin from '../../mixins/display-mode'
 import extend from '../../utils/extend'
 import { input, inline } from './datetime-props'
 import { QInputFrame } from '../input-frame'
 import { QPopover } from '../popover'
 import QDatetimePicker from './QDatetimePicker'
 import { QBtn } from '../btn'
-import { formatDate, isSameDate } from '../../utils/date'
+import { formatDate, isSameDate, isValid } from '../../utils/date'
 import { QModal } from '../modal'
 import clone from '../../utils/clone'
 
@@ -23,28 +24,25 @@ const contentCss = __THEME__ === 'ios'
 
 export default {
   name: 'q-datetime',
-  mixins: [FrameMixin],
+  mixins: [FrameMixin, DisplayModeMixin],
   props: extend(
     input,
     inline
   ),
   data () {
-    let data = this.isPopover() ? {} : {
+    let data = this.isPopover ? {} : {
       transition: __THEME__ === 'ios' ? 'q-modal-bottom' : 'q-modal'
     }
     data.focused = false
-    data.model = clone(this.value === 0 || this.value ? this.value : this.defaultSelection)
+    data.model = clone(isValid(this.value) ? this.value : this.defaultSelection)
     return data
   },
   computed: {
-    usingPopover () {
-      return this.$q.platform.is.desktop && !this.$q.platform.within.iframe
-    },
     actualValue () {
       if (this.displayValue) {
         return this.displayValue
       }
-      if (this.value !== 0 && !this.value) {
+      if (!isValid(this.value)) {
         return this.placeholder || ''
       }
 
@@ -67,9 +65,6 @@ export default {
     }
   },
   methods: {
-    isPopover () {
-      return this.$q.platform.is.desktop && !this.$q.platform.within.iframe
-    },
     toggle () {
       this[this.$refs.popup.showing ? 'hide' : 'show']()
     },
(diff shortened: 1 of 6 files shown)


src/components/datetime/QDatetime.js
```diff
@@ -26,25 +26,35 @@ export default {
...
+    data.model = clone(isValid(this.value) ? this.value : this.defaultSelection)
+    data.platformReactive = this.$q.platform // Make reactive. These values will probably never change, but just in case
     return data
   },
   computed: {
-    usingPopover () {
-      return this.$q.platform.is.desktop && !this.$q.platform.within.iframe
+    isPopover () {
+      // Explicit popover / modal choice
+      if (this.popover) return this.popover
```
rstoenescu: let's return true here. less code

src/components/datetime/QDatetime.js
```diff
@@ -26,25 +26,35 @@ export default {
...
+      if (this.modal) return !this.modal
```
rstoenescu: let's return false here

src/components/datetime/QDatetime.js
```diff
@@ -26,25 +26,35 @@ export default {
...
   ),
   data () {
-    let data = this.isPopover() ? {} : {
+    let data = this.isPopover ? {} : {
       transition: __THEME__ === 'ios' ? 'q-modal-bottom' : 'q-modal'
     }
     data.focused = false
-    data.model = clone(this.value === 0 || this.value ? this.value : this.defaultSelection)
+    data.model = clone(isValid(this.value) ? this.value : this.defaultSelection)
+    data.platformReactive = this.$q.platform // Make reactive. These values will probably never change, but just in case
```
rstoenescu: let's remove platformReactive and directly use $q.platform.
This is something that definitely won't change.

src/components/datetime/QDatetime.js
```diff
@@ -26,25 +26,35 @@ export default {
   mixins: [FrameMixin],
   props: extend(
     input,
-    inline
+    inline,
+    {
+      popover: Boolean,
+      modal: Boolean
```
rstoenescu: do you think it would be better to have a "mode" prop instead of the two? it also offers room for expansion later & it will be less code

rstoenescu (changes requested):
Another good idea with this PR.

src/components/datetime/QDatetime.js
realcarbonneau: Possibly, I had originally combined the two, but for now, it seems seperating them would be simpler and more clear for user declaration
`<q-datetime  popover ...`
`<q-datetime  modal ...`
instead of
`<q-datetime  mode="popover" ...`

But I leave the decision to you.

src/components/datetime/QDatetime.js
realcarbonneau: Good point! :)

src/components/datetime/QDatetime.js
realcarbonneau: Good point! :)

src/components/datetime/QDatetime.js
rstoenescu: let's go with the two props. one thing I think it's gonna bite is that devs will want same behavior for QColor too. can we make this into a mixin and use it in both datetime and color? let me know if you'd rather do this only for datetime.

realcarbonneau: I have alot more work on QDateTime before I move onto another.  But yeah, when other components are changed to combine these concepts, we can make a common mixin.  I will simply follows your design decisions are that time. :)

realcarbonneau: But if you feel it's a priority now to harmonize QColor with this QDateTime feature , tell me and I can do it! :)

rstoenescu: it would be a priority in order to maintain consistency. but it's your choice. however it will fit your schedule. I can do it tomorrow morning so you can focus on more datetime goodies if you want.

realcarbonneau:
This now integrates a general mixin and integrates the explicit display mode setting into QColor

However, I think in general, the QColor and QDateTime need to have their naming convention harmonized. At the moment, it is structured as follows, which is confusing.
QColor = QDateTimePicker
QColorPicker = QColor

But even further, they could each be just one component with three modes (inline, popover, modal).
QColor
QDateTime

But I don't want to do that in this right now in this PR. I will continue going thru the flexiblity features and then see later.

Here is the video of the QColor tests.
qcolor-explicitpopovermodal

sebastienroul:
@realcarbonneau and @rstoenescu : Is there anyway to set the mode for all the app ?
We use iframe all the time (to include new apps (with QUASAR) in our legacy apps). But setting popover="true" on all q-datetime is boaring... So anyidea how to set it globally for the app ?

rstoenescu:
Sorry, currently no way...

sebastienroul:
@rstoenescu : thanks for answer... currently we'll copy/paste :o)

realcarbonneau:
You can wrap the q-datetime into your own component which forces popover="true".   Then you can your your new component everywhere and not have to copy/paste, and it also stays flexible and compatiable with upgdates and upgrades.

PR #1444, opened 2018-01-18
QColor Explicit Popover and Modal selection

realcarbonneau:
Here is a first pass and integrating the Explicit Popover and Modal selection to the QColor and having a general mixin.  I will do another pass later and do more tests.

However, I think in general, the QColor and QDateTime need to have their naming convention harmonized. At the moment, it is structured as follows, which is confusing.
QColor = QDateTimePicker
QColorPicker = QColor

But even further, they could each be just one component with three modes (inline, popover, modal).  

But I don't want to do that in this right now in this PR.  I will continue going thru the flexiblity features and then see later.


realcarbonneau:
Combined into #1442

PR #1448, opened 2018-01-18
QChipsInput Custom Add Icons

realcarbonneau:
Adds the flexibility to specify a custom icon for the add action at the far right.  See the video below.

customicons


realcarbonneau:
I also tried to integrate it with your new global icons.  I am not familiar with the alternate icon sets, so I only added them to two.

PR #1456, opened 2018-01-20
[post v0.15] QInputFrame: Consolidation and redesign of details pattern.

realcarbonneau:
The details features provides (feel free to provide a better name) the user with an additional display that contains a helper or picker with additional details to make value selection more visual.

I have consolidated the dropdown option of the QColor, QDatetime and QSelect into a consolidated feature withing the QInputFrame.  I have also added it to the QInput as an example of it's use.

The feature is controlled by the detailsIcon prop which can be have the following values:
1) false: force deactivation of the icon
2) true, defined or empty string: force activation of the default icon
3) string: force the display of the specified icon in the string
4) undefined: If the component has a defaultDetailsIcon computed property, display the default icon, if not, no icon.

New default icon choices (integrated in icon globals):
1) Date and Datetime: calendar
2) Time: clock
3) Color: color palette

Clicking the details icon emits a 'details' event.  In the QInput, I have also passed the event on to the parent so that user can use this feature.

I have not yet changed the click behavior of the existing components to use the details event to keep this PR simple, and I want to think more about the full input box click versus the details icon click, and direct input box editing for details enabled inputs.

Please leave most of my comments in place for my future sanity and for others to better understand the code.  

Here are the videos:
qinputframe-qdatetime

qinputframe-qcolor

qinputframe-qselect

qinputframe-qinput

### DIFF
--- a/dev/components/form/color-picker.vue
+++ b/dev/components/form/color-picker.vue
@@ -68,6 +68,15 @@
       <q-color popover v-model="inputModelRgb" float-label="RGB Popover" />
       <q-color modal   v-model="inputModelRgb" float-label="RGB Modal" />
       <q-color modal   v-model="inputModelRgb" float-label="RGB Modal Readonly " readonly  />
+      
+      <h4>Details Icon Variations</h4>
+      <q-color v-model="inputModelRgb"                                      float-label="Default (undefined)" />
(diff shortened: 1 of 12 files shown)


rstoenescu:
Postponing this after v0.15.

PR #1468, opened 2018-01-22
fix(QChipsInput): more tweaks on chips color

pdanpdan:
Adjust chip color and text-color so that a default <q-chips-input inverted /> has visible chips


realcarbonneau:
I had attempted to have a generalized approach to avoid any of invisible default color combinations. I was close, but got stuck for a few special cases internally in quasar that required changes, and I didn't want to propose a PR that brought alot of changes.

There are alot of places where this similar code is repeated.  Tell me if it would be important to continue trying to consolidate this default color pattern.

PR #1487, opened 2018-01-25
Increase height to stop letters from being cut off

mesqueeb:
Letters like "j" get their bottoms cut off if the height is only 18px. Change to 22px.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/components/input/input.ios.styl
+++ b/src/components/input/input.ios.styl
@@ -8,7 +8,7 @@
   padding 0
   color black
   background transparent
-  min-height 18px
+  min-height 22px
   font-size inherit
   overflow hidden
   resize none
@@ -25,7 +25,7 @@
     box-shadow inherit
 
 input.q-input-target
-  height 18px
+  height 22px
   outline 0
 .q-input-chips
   min-height 36px !important
--- a/src/components/input/input.mat.styl
+++ b/src/components/input/input.mat.styl
@@ -8,7 +8,7 @@
   padding 0
   color black
   background transparent
-  min-height 18px
+  min-height 22px
   font-size inherit
   overflow hidden
   resize none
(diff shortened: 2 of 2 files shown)


Frondor:
Can you reproduce it on both themes?
Use: 
 - https://codepen.io/rstoenescu/pen/KQRZJg
 - https://codepen.io/rstoenescu/pen/paVpBN

I think its a font related issue, nothing to do with input's height. Be careful with that

mesqueeb:
Dear @Frondor 
I cannot reproduce it inside a pen, but as you can clearly see, the input gets cut off.
One of the reasons is: I need to put input font-size to 16px on mobile devices, because otherwise apple has an automatic zoom when focusing an input field. Putting the size at 16px prevents this.

However, I checked the source again and I'm really confused why quasar needs to set `q-input-target` to `height: 19px`. Why not just `height: inherit` and no one will have any trouble with the input fields, even when changing fonts. The `q-input-target` already has `min-height: 18px` so why set it at a default of 19px right after?

img_5254

mesqueeb:
@nothingismagick 
I confirmed the bug with the most recent version.
I have created a pen: https://codepen.io/mesqueeb/pen/XqGMMd

It seems it's **only on safari** and also it's because of the line height not set but the height being set.
This causes all "g's" to be cut off at the bottom on my screen:

image


I have added a "fix" button on the code pen. please check it out.

mesqueeb:
@nothingismagick Thanks I've opened the issue here:
https://github.com/quasarframework/quasar/issues/2084

PR #1503, opened 2018-01-27
QModal no imposed value prop changes on route change fix #1502

### LINKED ISSUE #1502: QModal auto-hide with route change, modal state becomes out of sync with the v-model/value prop
The QModal control has an auto-hide when the route changes, regardless if a v-model/value prop has be provided.  This results in the modal display becoming out of sync with the v-model/value prop.

I was testing a login modal and this is a problem because on the initial load and route changes, the modal closed even if the v-model/value prop was still in a state that requires the modal to be open.  I will submit a PR.

```
  watch: {
    $route () {
      this.hide()
    }
  },
```

realcarbonneau:
As per issue #1502, route changes should not impose a v-model change. 

Another alternative would be to have a noRouteChangeDismiss option. 

Please advise.

### DIFF
--- a/src/components/modal/QModal.js
+++ b/src/components/modal/QModal.js
@@ -78,7 +78,8 @@ export default {
   },
   watch: {
     $route () {
-      this.hide()
+      // If value prop is provided, always respect it.  Therefore no hide on route change.
+      if (typeof this.value === 'undefined') this.hide()
     }
   },
   computed: {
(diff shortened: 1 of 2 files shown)


realcarbonneau:
We need to keep my code for the undefined.  Vuejs imposes false to an undefined prop, the only way around this is to set the default as undefined.   However, they seems to bechanging that soon, see my last post here https://github.com/vuejs/vue/issues/4792, they just made a commit on this.

Also, see my jsfiddle showing the current behavior.
https://jsfiddle.net/realcarbonneau/daxs9722/

So, your adjustment mean that it will never be undefined, so the hide() on route change will never occur until Vuejs releases a change.

PR #1506, opened 2018-01-27
QModalLayout Parent-Child Check

realcarbonneau:
This adds a check to ensure that QModalLayout is a child of QModal.

Here is a video of some tests performed after the change to validated that there were not side effects.

parentchildcheck-tests

### DIFF
--- a/src/components/modal/QModal.js
+++ b/src/components/modal/QModal.js
@@ -48,6 +48,11 @@ let openedModalNumber = 0
 export default {
   name: 'q-modal',
   mixins: [ModelToggleMixin],
+  provide () {
+    return {
+      modal: true
+    }
+  },
   props: {
     position: {
       type: String,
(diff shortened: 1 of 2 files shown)


rossity:
@realcarbonneau @rstoenescu can we revert this change? or at least disable the console message in production? I have components that conditionally need to be either rendered normally or as modals and as such I use q-modal-layout as my template wrapper.

realcarbonneau:
Razvan had confirmed that q-modal-layout was designed to only be used to with the q-modal, therefore this check is to ensure it is being used according to the design, thus assisting the devs and also reducing the number of potential issues logged because of incorrect use.  

I will let @rstoenescu descide if he wishes to change this.

rossity:
Totally understand! Would love an option to silence it somehow though :)

If not I can find a way to work around it. Thanks!

PR #1507, opened 2018-01-27
QDocument bind to document title from template

realcarbonneau:
This is an extremely simple component that permits the user to change document header properties directly from the template.  At the moment, there is no simple way to reactively bind to the document title. Other properties such as keywords, description, etc. could easily be added if users find this useful.  

Of course, in many cases it would be better to update these in the router or programmatically, however, here are the advantages of the component:
  -Easier for beginners
  -Visual reference in the template
  -Reactive on creation, update and route changes. 

To do this programatically would require constants and logic programmed in the router or Vuex or other global variables.  
To do this programatically in the component requires setting the title on creation and also subsequently watching for route changes to update the title

There also exists https://github.com/troxler/vue-headful, which is more complete, but more complexe to integrate for beginners, and it is alot bigger. I would suggest to just support a minimal set of properties, mostly for beginners and simplicity.  For advanced or exhaustive support, devs can use the external vue-headful.

Video of demo:  ```<q-document :title="`My choice: ${option}`"/>```

documenttitledynamic


rstoenescu:
Hi. this is too essy to be done. Don't think it needs to be added to Quasar.

PR #1636, opened 2018-02-24
Fix many icons that the name was outdated

insign:
Outdated or never existed. Not sure since I'm new with this great Material Icons pack.

Icons search: https://materialdesignicons.com

### DIFF
--- a/icons/mdi.js
+++ b/icons/mdi.js
@@ -7,10 +7,10 @@ export default {
     warning: 'mdi-exclamation'
   },
   arrow: {
-    up: 'mdi-arrow-upward',
-    right: 'mdi-arrow-forward',
-    down: 'mdi-arrow-downward',
-    left: 'mdi-arrow-back'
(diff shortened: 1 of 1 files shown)


leopiccionia:
As I previously said in Discord, [material.io/icons ](https://material.io/icons/) and [materialdesignicons.com](https://materialdesignicons.com/) are not related. Quasar icons names are perfectly aligned with the previous.

Actually, if one read [this](https://github.com/Templarian/MaterialDesign/issues/2545), it's learned that materialdesignicons.com actually predates Google initiative, and aliasing material.io names are not top priority. And, per [this](https://github.com/Templarian/MaterialDesign/issues/1797), that some material.io icons probably won't be supported, so it's intended as a strict superset.

I'd suggest to eventually support materialdesignicons.com, with its own namespace. I don't have a suggestion for the name.

insign:
Quasar has a file for every package supported to align the names in favor of quasar components.

PR #1699, opened 2018-03-05
Enhancement: Added auto margins (e.g. q-mx-a)

mesqueeb:
Added auto margins (e.g. `q-mx-a`)

I used to heavily use my own helpers like so:
```css
.ml-auto{
    margin-left: auto;
}
.mr-auto{
    margin-right: auto;
}
.mlr-auto{
    margin-left: auto;
    margin-right: auto;
}
```

Since Quasar has a good method of helper construction, I'd like to add the `auto` keyword.

If you accept I'll also update the docs and do a PR. : )

### DIFF
--- a/src/css/core.variables.styl
+++ b/src/css/core.variables.styl
@@ -25,6 +25,10 @@ $spaces = {
   xl: {
     x: ($space-x-base * 3),
     y: ($space-y-base * 3)
+  },
+  a: {
+    x: auto,
+    y: auto
   }
 }
 


rstoenescu:
It's already in.

mesqueeb:
Are the docs not updated yet?
`q-[p|m][t|r|b|l|a|x|y]-[none|xs|sm|md|lg|xl]`
I meant to add `auto` or `a` for auto in the most right part.

So we can do something like `q-mx-auto` for auto margins on the x axis

PR #1908, opened 2018-04-10
improve fullscreen backdrop of modal experience

mesqueeb:
I would argue that seeing the backdrop of the modal droop over the borders on all 4 sides gives a much better user experience with dragging a bit over the edge / fast scrolling.
In case you do not agree feel free to close the PR! : D


rstoenescu:
Hi,

It would work for that backdrop, but those classes are used somewhere else too, and it breaks everything.. Sorry, but gonna close this.

PR #1941, opened 2018-04-16
QDialog and QModal: refocus original element after hide

pdanpdan:
- introduce noRefocus prop
- when noRefocus not present:
  - save the activeElement when modal was opened
  - restore focus on it when hidden


rstoenescu:
Dan, would you be so kind to PR this to docs too, pls?

PR #1946, opened 2018-04-17
Datetime input, default-view bug

Morgosh:

- [X] Bugfix

- [X] No

**Description**
If you check datetime input on quasar documentation. There is an error when you set default-view parameter. The default view is only set on the second click on date. So first time you click it, the default view is not the same as default-view parameter. After the second click it is. 

An example bug can be seen in the example itself. Click on the default view row twice, and you get different popout.
http://quasar-framework.org/quasar-play/android/index.html#/showcase/forms/datetime/datetime-input


rstoenescu:
Thanks for contributing! This will get into 0.15.11

KnowZero:
@rstoenescu I don't think default-view is completely fixed. It is acting really odd.

http://quasar-framework.org/quasar-play/android/index.html#/showcase/forms/datetime/datetime-input

When I visit on desktop chrome or firefox the demo page which has default view set to "month" and click on it, it shows "day" no matter how many times I click on it. Now if I lose focus by click on a different field or click on the page 2 times and click it again, it only then shows "month" properly.

PR #1974, opened 2018-04-24
QUploader - Hardcoded content-type field unnecessary.  file has own content-type

Lou-Der:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ x] Other, please describe:  

Unnecessary form field that is not removable from call.

- [ ] Yes
- [x] No

When submitting  multipart/form-data with file uploads.  The xhr file reference includes the content-type of the file being sent with the following details:

> Content-Disposition: form-data; name="files"; filename="test.csv"
> Content-Type: application/vnd.ms-excel

A hard coded "content-type" field is not necessary and will actually cause issues with some webserver back-ends/services like oracle, kinto, etc.

If a user actually needs to include a "Content-Type" field they can use the headers parameter.

### DIFF
--- a/src/components/uploader/QUploader.vue
+++ b/src/components/uploader/QUploader.vue
@@ -405,7 +405,6 @@ export default {
         this.additionalFields.forEach(field => {
           form.append(field.name, field.value)
         })
-        form.append('Content-Type', file.type || 'application/octet-stream')
         form.append(this.name, file)
       }
       catch (e) {
--- a/src/components/uploader/QUploader.vue
+++ b/src/components/uploader/QUploader.vue
@@ -184,6 +184,7 @@ export default {
       type: Array,
       default: () => []
     },
+    noContentType: Boolean,
     method: {
       type: String,
       default: 'POST'
@@ -405,6 +406,9 @@ export default {
         this.additionalFields.forEach(field => {
           form.append(field.name, field.value)
         })
+        if (this.noContentType !== true) {
+          form.append('Content-Type', file.type || 'application/octet-stream')
+        }
         form.append(this.name, file)
       }
       catch (e) {


nothingismagick:
Nice and simple pull-request. Do you have more information or an RFC where this info can be seen? And it would be excellent if you tested it on a few backends like a classical express.js and also kinto...

The only thing that concerns me is your comment: "If a user actually needs..." I feel that the way it currently works is actually one expects it to work - and changing it now might actually break some systems. I think it would be more "quasar-like" if this was a config element with the current behavior as a default and your approach as an option.

Lou-Der:
RFC for upload spec can be found at:
https://www.ietf.org/rfc/rfc1867.txt

There is no reference to a global content-type field as quasar 15.10 is currently forcing.  

A description of a standard upload:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type
You can see a standard format upload that is NOT including a separated content-type field.

If you debug at any standard html page that generates an upload to a server, it does not generate this field.

I can give you a set of testing results showing the difference between an upload using standard html and quasar uploader.  The only difference is the hard-coded content-type field.

Reference another user having same issue:
http://forum.quasar-framework.org/topic/455/hardcoded-form-data-in-uploader

Note: In oracle web services, you cannot have a incoming field with a hyphen in it. 

I would expect q-uploader to replicate standard html upload results.   I am confused as to why you think a global content-type field is "normal".   There is no indication anywhere I can find that this is a standard field for uploads.   Normally all content-type field parameters are encased in the file field itself.   It sounds like your assuming each file upload is individual which actually _isn't_ normal....  

FYI:
Another main difference between q-uploader and standard html is that multiple files are normally sent in a single upload using an array of file items.  While I don't need this now, I was considering making another PR later to adjust multiple with a new prop for "batch" or "individual" upload mode.

Thanks :smile:

nothingismagick:
Thanks for the detailed information, I must have been doing it "wrong" for years now. This part about the server response in the RFC is awesome:
> "application/x-please-send-files"
It's like stepping into a time machine.

Lou-Der:
I'm really new to GIT Pull requests.  Is there anything else I need to do to have this merged into the master code tree?   Was hoping it would be in the last release :)  

Also, should i git rebase this branch for the new dev release?  There currently are no conflicts.

Lou-Der:
@rstoenescu Any chance this could be merged into to the next release?   

If not, could you let me know what changes you would like instead to allow this hard-coded field to be turned off.

Thanks!

rstoenescu:
Hi @Lou-Der , taking care of this today. It will be in 0.16.1.

PR #2093, opened 2018-05-29
Add uploadFactory to the QUploader component

GastroGeek:
This will allow users to override the default XHR mechanism for file uploads and file status updates. Users may then call the included updateProgress(percentage: float) method to feed back status updates to the component. Also added global 'abort' $emit to notify parent component of full-blown abort(). This allows users to take action, provide any feedback and/or reset the component if desired.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/src/components/uploader/QUploader.vue
+++ b/src/components/uploader/QUploader.vue
@@ -181,6 +181,10 @@ export default {
       type: Function,
       required: false
     },
+    uploadFactory: {
+      type: Function,
+      required: false
+    },
     additionalFields: {
       type: Array,
       default: () => []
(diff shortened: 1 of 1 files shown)


geoidesic:
A suggestion: the __add method also makes hard-coded use of readAsDataURL, which is undesirable for websocket file transfer – we'd want to be able to customise that call as well to something else – for example: readAsArrayBuffer.

So in the same way that you have a `urlFactory` prop, you could also add a `fileAddFactory` prop, which could override the normal `__add`. I'm not sure if that will nerf the UI? I would also suggest moving the following logic from `__add` method into the `add` method:
```
if (this.addDisabled) {
        return
}
```

Issue #1902, opened 2018-04-09
Carousel not able to go full height in page due to q-page not defining height


gecko-8:
I'm building an app that has a q-carousel embedded inside a q-page. The problem is that no matter what I try, I can't get the carousel to go 100% height. I'm 99% sure I've narrowed down the issue to the q-page only defining it's min-height and not height. I've reproduced the issue in this fiddle [https://jsfiddle.net/MaelstromCdn/coctcox1/5/](url). You'll see that the carousel is squished up against the top but if you add style="height: 100vh;" to the q-page tag it works fine. The min-height value is obviously calculated based on what other page elements are visible so I'm not comfortable creating a pull request as I'm just getting started with Quasar.

For reference, here's my full page code:
```
<template>
  <q-page>
    <q-carousel :autoplay="5000" quick-nav infinite class="full-height">
        <q-carousel-slide class="flex flex-center">Slide 1</q-carousel-slide>
        <q-carousel-slide class="flex flex-center">Slide 2</q-carousel-slide>
        <q-carousel-slide class="flex flex-center">Slide 3</q-carousel-slide>
    </q-carousel>
  </q-page>
</template>

<script>
export default {
  name: 'PageWelcome',
}
</script>

<style scoped lang="stylus">
</style>
```
And my versions:
```
Operating System                Darwin(17.5.0) - darwin/x64
NodeJs                          8.11.1

Global packages
  NPM                           5.8.0
  yarn                          Not available
  quasar-cli                    0.15.14
  vue-cli                       2.9.3
  cordova                       7.0.1

Important local packages
  quasar-cli                    0.15.14 (Quasar Framework CLI)
  quasar-framework              0.15.10 (Build responsive websites, PWAs, hybrid mobile apps and Electron apps, all simultaneously using same codebase)
  quasar-extras                 1.0.2   (Quasar Framework fonts, icons and i18n.)
  vue                           2.5.16  (Reactive, component-oriented view layer for modern web interfaces.)
  vue-router                    3.0.1   (Official router for Vue.js 2)
  vuex                          3.0.1   (state management for Vue.js)
  electron                      Not available
  babel-core                    6.26.0  (Babel compiler core.)
  webpack                       3.11.0  (Packs CommonJs/AMD modules for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand. Support loaders to preprocess files, i.e. json, jsx, es7, css, less, ... and your custom stuff.)
[...]

PR #2120, opened 2018-06-04
[Q-Page] New `styleFn` prop for QPage

mesqueeb:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

I need this prop to be able to set **the q-page's height, in addition to the minHeight**.
I need it to be able to fix this kind of problem:

the codePen:
https://codepen.io/mesqueeb/pen/aGeKjm

Will probably also fix the carousel problem of #1902


mesqueeb:
PS I will update docs as well if this would be accepted. ♡

mesqueeb:
Update QPage.js

@rstoenescu 
I want to shortly discuss this component.
Today I had to make my app layout float in the middle of the screen when the window is high and wide.
(see screenshot below)

This made me having to override the `min-height` that's encoded on the `q-page` render. I had some trouble doing this! Trying to set `min-height` through style on the element got overridden sometimes by the quasar's style which is here.

This made me think the best way to tackle situations where the dev needs to custom manage `height`, `min-height` and `max-height`.

Instead of adding the possibility for the dev to also set the `height` through a boolean prop, I think it's better to make the developer set those things himself outside of this native component, but then give the option **to disable the `min-height` being set by quasar.**

I think this will give much more freedom.
However, the only problem in that case is that the developer would probably want to have access to the offset:
```js
const offset =
        (this.layout.header.space ? this.layout.header.size : 0) +
        (this.layout.footer.space ? this.layout.footer.size : 0)
```
to be able to keep this in mind when adding custom (min/max)height to the q-page.

So to summarize:

### my suggestion:
"I removed the `height` prop again, set a `minHeight` prop which the dev can set to `false` if he wants to prevent q-page from setting the minHeight. If it's `true` or `undefined` it will just set the minHeight like always.

### My Question:
Is it possible to access this offset somehow in vue components? If we can give the dev access to the offset in a component it's better to write in the docs on how to do that and further not set the props on this component.

PS: Good job on the Vue event today!!


image

rstoenescu:
Ok, so best way to go is to use a `style-fn` prop, which is a function. It takes the offset as parameter, and it needs to return the style that will be applied to QPage. Available in future Quasar v0.16.1.

Would you be so kind to add this to docs? Thanks!

mesqueeb:
@rstoenescu
Currently the the custom `styleFn` prop is available on QPage only.
Any chance we can extend this to the QLayout, `#q-app` and `body` as well?

Since all these also use `vh` which I must replace because of Mobile Safari's improper handling of `vh`.

PR #2188, opened 2018-06-19
emit click event on qFab

m0dch3n:
https://jsfiddle.net/xrdapu5t/ 

Currently qFab isn't emitting the click event, when clicked

### DIFF
--- a/src/components/fab/QFab.js
+++ b/src/components/fab/QFab.js
@@ -22,6 +22,12 @@ export default {
       default: 'right'
     }
   },
+  methods: {
+    click (e) {
+      this.toggle(e)
+      this.emit('click', e)
+    }
+  },
   watch: {
     $route () {
       this.hide()
@@ -51,7 +57,7 @@ export default {
           glossy: this.glossy
         },
         on: {
-          click: this.toggle
+          click: this.click
         }
       }, [
         this.$slots.tooltip,


rstoenescu:
Hi,

Thank you for wanting to contribute!
This is unnecessary though. You are trying to catch the native DOM "click" event and you can do that by using the ".native" modifier:

```html
<q-fab @click.native="onClick" ....
```

https://vuejs.org/v2/guide/components-custom-events.html#Binding-Native-Events-to-Components

m0dch3n:
Hi,

ok, I agree, it's the same like .native, so this commit is useless...

However .native is also triggered, when clicking on an action button...

c.f. https://jsfiddle.net/fwjpbLuz/

PS: The documentation also says
```
#QFab Vue Events

@click | Triggered when clicking/tapping on main FAB after it was already opened.
```

This @click event isn't triggered at all...

m0dch3n:
What do you think about this

```
  methods: {
    click (e) {
      e.stopPropagation();
      this.__qFabClose().then(() => {
        this.$emit('click', e)
      })
    }
  },
```

in QFabAction ?

rstoenescu:
Fixed the documentation. QFab `@click` is obsolete (went away long ago).
Why stopPropagation? If you need to add it you can do it from the `@click="e => { e.stopPropagation(); doSomethingElse() }`

m0dch3n:
.native on QFab is also triggered, if you click on QFabAction.

So if you only want to know if the main fab button was clicked, you need to stopPropagation or to check the origin in your event method

PR #2246, opened 2018-07-10
make throttle return last return value of throttled function

panstromek:
I use this version of throttle so I decided to PR it. 

This is also Lodash's throttle behaviour.


rstoenescu:
Thanks for the PR! Will be available in next beta to be released today.

PR #2681, opened 2018-10-16
Made multiple prop obsolete

smolinari:
Hi Razvan,

This is in reference to my short PM chat with you and it was simple. The idea is, "let the data form the select type". So now, if the model is an array, the selection is automatically a multi-select. If the model isn't an array (a string), the select is a single select. It works and makes the need for the multiple prop obsolete. This will also stop confusion as to why the selection doesn't work, when someone adds the multiple prop, but doesn't add an array as the initial model.

If you are in agreement, I'll update the docs too. 

Looking forward to hearing why this isn't a good idea. 😜 

Scott 

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

It means the use of the "multiple" prop isn't necessary. 

**Other information:**

### DIFF
--- a/dev/components/form/select.vue
+++ b/dev/components/form/select.vue
@@ -27,32 +27,32 @@
       </q-field>
 
       <p class="caption">Multiple Selection</p>
-      <q-select @change="onChange" @input="onInput" multiple v-model="multipleSelect" :options="selectListOptions" clearable/>
-      <q-select @change="val => { multipleSelect = val; onChange(val); }" @input="onInput" multiple :value="multipleSelect" :options="selectListOptions" clearable/>
-      <q-select @change="onChange" @input="onInput" multiple v-model="multipleSelect" :options="selectDisabledOptions"/>
+      <q-select @change="onChange" @input="onInput" v-model="multipleSelect" :options="selectListOptions" clearable/>
+      <q-select @change="val => { multipleSelect = val; onChange(val); }" @input="onInput" :value="multipleSelect" :options="selectListOptions" clearable/>
+      <q-select @change="onChange" @input="onInput" v-model="multipleSelect" :options="selectDisabledOptions"/>
 
       <p class="caption">Multiple Selection with Chips</p>
       <q-field label="gogu" :count="10">
-        <q-select @change="onChange" @input="onInput" multiple chips v-model="multipleSelect" :options="selectListOptions" float-label="Some label" placeholder="Gigi" max-height="36px" clearable/>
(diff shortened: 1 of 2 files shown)


pdanpdan:
This would not work if the select is not multiple but the values are arrays.

smolinari:
I'm trying something and it isn't working. If I set v-model to an option value as an array, the value  doesn't show up as initially selected. 

https://jsfiddle.net/smolinari/0pt9eydo/

Where is my thinking wrong? Or did I just find a bug? :) I don't think display-value should be needed here, should it?

Scott

pdanpdan:
2 array literals are not equal. Make the list of ouptions use one item in the list as initial value.

smolinari:
Ok. You got me with this:

> Make the list of options use one item in the list as initial value.

:smile: 

I thought that was done by setting v-model to the option value, which you say won't work. How else can I set the initial value?

Scott

pdanpdan:
Either like this:
https://jsfiddle.net/0pt9eydo/6/

or like this:
https://jsfiddle.net/0pt9eydo/8/

pdanpdan:
It's not related to Quasar, but to how javascript works.
And the first solution is simpler, but only works if you have a default preset.
If you have dynamic data from API you have to find the current option (dynamic) in the current list of options (also dynamic), and this must be done for any non basic primitive type (array, object).

pdanpdan:
The fact that:

```javascript
const
  a = { a: 1 }, // or [1, 2]
  b = { a: 1 };  // or [1, 2]
console.log('a === b => ', a === b); // a === b => false
console.log('a == b => ', a == b); // a == b => false
```

Is not specific to Quasar or any framework. It's specific to the language (Javascript in this case, and anything that compiles to it).

smolinari:
I get that. And I thank you for the lesson. 😄 👍 

What I don't get or don't believe is that a developer would know that such a comparison is being made in the component. They'd have to look into the source code to know what is happening (or in the docs), wouldn't they? 

The fact I didn't know that two exact arrays don't equal each other in JavaScript has nothing to do with knowing how the comparison between the option values and the v-model value within the component is carried out.  

I've seen a number of questions/ issues raised because, for instance, the dev would enter the multiple prop, but not add an array (even an empty one) into v-model. If they enter just `''`, Quasar stays silent until a selection is made and then an `i.push is not a function` error occurs . I added the need to enter an array as v-model in the docs, so this requirement is clearer. (this situation also prompted me to try out this PR too. 😄) 

So, if devs can't figure out they need to enter an array for multiple selections to work, they surely won't know they can also have arrays as values for single selections. And, they most definitely won't know how to get the value to show up as an initial value with those array values, because they simply don't know what is happening within the component. Especially morons like me. LOL!

Scott

PR #2767, opened 2018-11-20
Remove leftover console.log() from the previous commit

fansanelli:


rstoenescu:
Hi,

I thought this was in a component?? Having that console.log() in the demo pages is perfectly fine as it serves debugging purposes as it should.

PR #2796, opened 2018-11-28
Allow 0 (Number) to be printed in QBtn label

fansanelli:

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

If you will to have an empty button when you have the Number 0, you need to do something like:
```
<q-btn :label="myVar ? myVar : null" />
-or-
<q-btn :label="myVar ? myVar : ''" />
```

Current code don't allow printing the Number 0.
This behaviour lead me to write code like this:
```
<q-btn :label="'' + myVar" />
-or-
<q-btn :label="myVar ? myVar : '0'" />
```
but seems less elegant then the change in the component.
If someone prefer old-style-empty they can do my same "hack" (see above).
What do you think?

**Other information:**

### DIFF
--- a/src/components/btn/QBtn.js
+++ b/src/components/btn/QBtn.js
@@ -18,7 +18,7 @@ export default {
       return this.percentage !== void 0
     },
     isRectangleWithLabel () {
-      return this.label && this.isRectangle
+      return this.isRectangle && typeof this.label !== 'undefined' && this.label !== ''
     },
     width () {
       return `${between(this.percentage, 0, 100)}%`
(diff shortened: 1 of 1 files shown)


fansanelli:
Any interest in this change? I could rebase it  in case..

fansanelli:
I did a button in my code with an icon and a number that represents current progress, use can press it to increment the counter or go in edit and use arrows. The button initially was small then grew after the first click and it's not what I wanted.
Anyway I stumbled in this bug while developing it. So I imagine others could face the same behaviour and the question is: is it wanted?

panstromek:
I don't understand this PR 🤔 I tried all possible versions and they just work... Maybe it was already fixed?
https://jsfiddle.net/4k5rayhc/8/

pdanpdan:
This is already taken care of in the v1.

PR #2803, opened 2018-11-28
Pr dateformat

oger000:

- [ ] Bugfix
- [ x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ x] No

- [x ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

Description:
Optionally set language specific date format as part of the i18n files. Prefere the format of the i18n fiile if present. Otherwise use the default.

Reason: 
Using components with formated date output has to be given the language specific date format every time. With this PR the language specific date format is used by default and explicitly formating is only needed if differs from that.

Note: 
I know the PR is incomplete (e.g. missing the datetime type), but I wanted to know if the way I have choosen is sensible, before investigating deeper.


fansanelli:
Hi!
I have implemented a similar change in my code using Vue-i18n: for Italian (DD/MM/YYYY) and American (MM/DD/YYYY) date formats, but I convinced myself that it isn't correct to implement this at language level... It's more a kind of country preference and I know that the user will be from America or Italy in my case.
BTW I did also add currency formats ($ 1,000.00 vs € 1.000,00) in my project and could be a nice to have add-on if we follow this way.
Cheers
Francesco

smolinari:
Yes, normally there needs to be two different settings. Language and Locale. Language only sets the phrases used in the program for the set language. It is usually only two characters, like "en" for English or "de" for German (Deutsch). The locale sets the formatting or dates, currency and numbers and is 5 characters. For instance, en_US for the United States and de_DE for Germany. You could also have en_GB, which is for the UK. Usually, the first two letters of the locale denote the language, but you still need the two different settings, because someone might want Spanish as the language, but also want the en_US locale/ formatting. 

Above and beyond that, there needs to be a currency choice too, but that is sliding into the realm of business logic and not necessary for Quasar to have. :)

Scott

PR #2824, opened 2018-12-03
fix[#2815 ] animScrollTo - set final scroll position if no time left for the animation

### LINKED ISSUE #2815: setScrollPosition with duration is not precise
### Software version

Quasar: ^0.16.0

When scrolling with duration, its not precise. When the duration is short (500ms) and the distance is big (3000 pixels), you "land" around 100 pixels away from the real target. It looks a bug in the animation loop. This doesn't happen when you use duration zero.

panstromek:
fixes #2815 by setting desired scroll position directly if there is no time left for the animation 

This is just a quick  fix, though. Ideally, this method should be reimplemented to be framerate-independent (there is framerate hardcoded, now).

### DIFF
--- a/src/utils/scroll.js
+++ b/src/utils/scroll.js
@@ -24,6 +24,9 @@ export function getHorizontalScrollPosition (scrollTarget) {
 
 export function animScrollTo (el, to, duration) {
   if (duration <= 0) {
+    if (el.scrollTop !== to) {
+      setScroll(el, to)
+    }
     return
   }
 
--- a/src/utils/scroll.js
+++ b/src/utils/scroll.js
@@ -23,18 +23,19 @@ export function getHorizontalScrollPosition (scrollTarget) {
 }
 
 export function animScrollTo (el, to, duration) {
+  const pos = getScrollPosition(el)
+
   if (duration <= 0) {
-    if (el.scrollTop !== to) {
+    if (pos !== to) {
       setScroll(el, to)
     }
     return
   }
 
-  const pos = getScrollPosition(el)
-
   requestAnimationFrame(() => {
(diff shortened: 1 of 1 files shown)


panstromek:
The position is but I was talking about hardcoded '16' . Are you sure it is always 16? AFAIK 60fps is just recommended but it should match the framerate and 1000/60 is just rougly 16. It's more 17 actually :D. Anyway, callback argument is timestamp, so it is probably more accurate to calculate the position based on timestamp diff. I'd guess it should work better when browser lags..

rstoenescu:
You're missing the point. The method needs to **animate** the scroll. If browser takes more time to call the handler than the equivalent of a frame it means it's already busy doing something else. So using a timestamp instead of relying on requestAnimationFrame() can possibly make the scroll not animated at all (sometimes jumping really long distances). requestAnimationFrame() ensures that the page is updated right before the next page repaint and does not skip "frames". Also, any possibly multiple updates between page repaints are useless -- which can happen with a wrongfully set timestamp handler.

panstromek:
> So using a timestamp instead of relying on requestAnimationFrame() can possibly make the scroll not animated at all (sometimes jumping really long distances).
This is true - And I thought that's actually the expected behaviour. So the animation is more important than being "in time" on the desired position. Ok, that makes sense.

But what about those 16ms? Does the animation run twice as fast on 120fps monitor? And why it's not 17?

rstoenescu:
Getting more and more close to the v1.0 launch date, so can't really focus on this as it's in no way critical. If you got a better way of handling this then please PR it. And "yes", it would run twice as fast unfortunately on 120 fps monitor.

panstromek:
Definitely understood, it's not critical ;) I want to look at it eventually, I just wanted to make sure I understand it correctly ;) thx

PR #2847, opened 2018-12-08
[v1] QMediaPlayer

hawkeye64:
QMediaPlayer component for Quasar


hawkeye64:
This is now an app extension

PR #2920, opened 2018-12-27
restore #2838 formatDate: Fix NaN on IE/Safari(iOS)

fansanelli:
This got lost


fansanelli:
Gentle ping, sirs

danicholls:
Oh, good, I thought I was going crazy! Wreaking a bit of havoc in my Cordova app. Is there a workaround besides doing it manually from scratch? Would like to keep things as uniform as possible.

danicholls:
By doing it manually I meant formatting the date myself (in an ad hoc fashion instead of with formatDate). Updating to the latest version fixed it.

fansanelli:
@danicholls latest version works for you because it contains this changes. I fear that your doing it manually also means not sharing fixes like this. In Quasar has been done a very good job with the dates and it's a pity if a new users find NaN/NaN/NaN after releasing their projects, isn't it?
Cheers,
Francesco

danicholls:
Yes, it would be! I don’t think I have anything special to contribute here, or I definitely would have worked up an issue. My note was just thanks for putting the change in and an issue because it was impossible to find it mentioned online otherwise. 

If anyone comes across this and needs help handling dates apart from Quasar’s function, what I can offer is best said by MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

PR #2930, opened 2018-12-28
Document QDate, QTime API

steveclarke:
Take a closer look at the datetime-mixin.json file. I have the "value" prop documented there. It's not quite generic enough to match the usage in the components. For QDate it needs to be a date, whereas with QTime it needs to be a time. Can this be better documented in the mixin docs or should I break them out separately into the respective component docs?


rstoenescu:
Put common stuff into the mixin, then add specific details (like "desc" prop) into QDate/QTime.json.

steveclarke:
Oh nice! Didn't realize I could do that.

PR #2934, opened 2018-12-30
Restart work upon reset

louisameline:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

When the `reset` method is called, I can only assume that it's because the content of the scrollable area was emptied and that we want to make a fresh new start on loading new content. But if `working` had been set to `false` previously (because all the old content was loaded for example), nothing will happen on the new content (the scripts blocks at line 40). So the `restart` call should make sure that the component is ready to work again.

- [ ] Yes
- [x] No

### DIFF
--- a/quasar/src/components/infinite-scroll/QInfiniteScroll.js
+++ b/quasar/src/components/infinite-scroll/QInfiniteScroll.js
@@ -70,7 +70,7 @@ export default Vue.extend({
 
     reset () {
       this.index = 0
-      this.working = true
+      this.resume()
     },
 
     resume () {
--- a/quasar/src/components/infinite-scroll/QInfiniteScroll.js
+++ b/quasar/src/components/infinite-scroll/QInfiniteScroll.js
@@ -70,6 +70,7 @@ export default Vue.extend({
 
     reset () {
       this.index = 0
+      this.working = true
     },
 
     resume () {


rstoenescu:
Hi,

Thanks for contributing!

I'd have to disagree a bit here. The point of "reset()" is to just reset the index (one thing only). Normally, if the Infinite Scroll is stopped, you need to also call "resume()". Or am I missing something?

louisameline:
My use case may help you understand what the issue is.
By default, I have a thousand icons that can be scrolled.
Outside of the scrollable, I also have a search field to help filter the icons. Whenever a search is made, the scrollable is basically emptied and filled again with only the matching icons. That's when I need to "reset" the infinite scroll.
When I need to "reset", I ideally don't want to have to care if the scroll was stopped or not previously, I just know that I want to start over.
I can't see what would be the point of only resetting the index if not to work again? But maybe you can see one.
Right now my workaround is to call `stop` and then `resume` right before `reset`, it doesn't make sense when you read my code. Thanks!

louisameline:
Thinking about it, I realize that the stoppage in this component can actually mean two things:
- If I call `done` with `true` as first parameter, it's "Stop because there is no more content"
- If I call the `stop` method, it rather means "Pause for a while until I ask you to resume"

Now, to the question "If the component is stopped, should `reset` resume the activity?"
I would answer yes to that in both cases, even if it's more obvious in the first case. I see no point in resetting and not work again.
Given the wording, I also expected it to work that way, as "reset" means "return to your initial state", which is to be working.
I hope it's clear :)

louisameline:
Side note: I see that `disable` is a recent addition to the component. But the wording is misleading as the component is not really disabled, it's paused.

I think `disable` should have the same effect as `stop()`, that is to say that the component unbinds its listeners and stays completely inert.

I would personnaly either 
- rename `disable` into `paused`, or (my preferred choice)
- augment the `disable` prop with the ability to unbind listeners. The prop could then be renamed `stopped` because it would be clear that it has the same effect as a call to `stop()`

louisameline:
@rstoenescu ? Ok for this 1-line PR? Thank you

louisameline:
I update this with the call to resume instead, you are free to accept or close it. Thank you

rstoenescu:
Hi,

Hope you don't mind, but I will close this. Left you a message on Discord for when you have time.

PR #2979, opened 2019-01-14
QSelect: Make expand-besides default to true

panstromek:
Why?
- it's native behaviour
- it's logical and intuitive
- you can click on the header to close the popover
   - right now, if you don't want to select or you select multiple values, you need to click outside to close the popover
- you can see list of all selected items easily (useful for long lists)
- My guess is that it will be more desired than the opposite

### DIFF
--- a/quasar/src/components/select/QSelect.js
+++ b/quasar/src/components/select/QSelect.js
@@ -60,7 +60,10 @@ export default Vue.extend({
       default: 500
     },
 
-    expandBesides: Boolean,
+    expandBesides: {
+      type: Boolean,
+      default: true
+    },
     autofocus: Boolean
   },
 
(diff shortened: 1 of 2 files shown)


rstoenescu:
Got 2 reasons why this is not a good idea:

1. It's against Material Design guide
2. If it's Boolean, it should never be set as default to `true`. Otherwise in your templates, you must write the bulky version `:something="false"` instead of a short: `not-something`.

panstromek:
1. Really? This surprised me honestly.. but I can't find it :D. I found only dropdown menu in design and select-menu in the reference implementation and both of them expand besides...

2. true

rstoenescu:
They change the material specs weekly, so who knows :))
I'd go with expand-besides on desktop and without on mobile (really important since small screens). But devs can pick and choose. `:expand-besides="$q.platform.is.desktop"`

PR #2981, opened 2019-01-14
Add addressbar-colors docs

codenamezjames:
Added addressbar-colors docs
Added a class to images inside of the markdown
Added rules the doc-img class


rstoenescu:
Nice adding the image plugin! :)

PR #3000, opened 2019-01-16
feat(docs): QCircularProgress

hawkeye64:


rstoenescu:
Hi,

Can you split up the examples by sections pls? This one is too big. Kind of like the Button page.

hawkeye64:
Can you please look at the following:
1) Min and Max example: It looks off to me.
2) No Motion Example: I could not get this to work as expected.

rstoenescu:
Merged so I can more easily see it in action.

rstoenescu:
1. min - max examples --> u're using different min/max and same model
Some are min=0 - max=50, some are min=50 - max=100... so model is incorrect as it cycles from 0 to 100.

2. No motion actually works; it jumps to the new value without animation. But there are so many values (100) and the increment is 1 and the cycling happens fast, so it's actually barely noticeable. To notice it, it needs less values (or higher increments), slower cycling.

3. `:show-value="true"` --> use short version `show-value`

PR #3001, opened 2019-01-16
add Platform Detection docs

steveclarke:


rstoenescu:
Example missing. I'll merge then add it.

PR #3033, opened 2019-01-19
Form validation page

panstromek:
Just a blind copypaste for start

### DIFF
--- a/docs/src/pages/vue-components/form-validation.md
+++ b/docs/src/pages/vue-components/form-validation.md
@@ -1,83 +1,74 @@
 ---
-title: Docs
+title: Form Validation
 ---
+Recommended package for Form Validations is `Vuelidate`.
+Get started with the [documentation](https://monterail.github.io/vuelidate/).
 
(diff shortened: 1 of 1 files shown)


rstoenescu:
Hi,

In this particular case, it's about http://0.0.0.0:8080/form/input-validate (`yarn dev` in /quasar folder).
Everything's obsolete here.

panstromek:
I have seen that page but I thought using `:rules` is just another option. From my understanding the code here is still usable in new Quasar.. do you imply that Vuelidate support has been dropped?

rstoenescu:
No, it can still be used. But it probably deserves another page in docs.

rstoenescu:
For now just focus on the internal `rules` prop as indicated.

panstromek:
I see.. well. This will take me a bit more time then.

rstoenescu:
No problem. Thanks for contributing! If you wanna choose something else it's ok too, just throwing that out.

Issue #3079, opened 2019-01-24
q-table, safari, ios


Lin84:
Problem with "q-table" component in safari or ios.
if you scroll to the end for right of table, the border will disappear. 
in case for checkable table, user will not even see what row is selected

https://quasar-framework.org/components/datatable.html

Video demo (open link to q-table component above in safari):
https://monosnap.com/file/x7YwMqTq1dwChK5TKMjY3ixoswBHHb

![image](https://user-images.githubusercontent.com/7355572/51684198-d4d4c400-1feb-11e9-9342-96a12bd1447c.png)

Lin84:
After some investigation.
FYI creator of Quasar:
the problem I described above is caused by "td" has style 'position: relative'

as hot fix for problem:
set 'position:static' for "td" inside "q-table"

Video to demo:
https://monosnap.com/file/GHZVDTRkplbnMvnyywswfW9ZGLfzyV

Big thx to Creators of Quasar :) for creating this nice framework, Keep it up :)

Lin84:
I am not sure about your process further with this discovery, I will keep this issue opened for now. @rstoenescu please close this after proper solution,

rstoenescu:
Hi,

Thanks for reporting and also investigating. Fixed it for v1 with the commit above. If you need this fast for 0.17, would you be so kind to port this fix to "v0.17" branch?

Many thanks,
Razvan

PR #3102, opened 2019-01-27
fix(quasar): q-table, safari, ios quasarframework (#3079)

Lin84:
video to demo issue:
https://monosnap.com/file/x7YwMqTq1dwChK5TKMjY3ixoswBHHb

video to demo solution:
https://monosnap.com/file/GHZVDTRkplbnMvnyywswfW9ZGLfzyV

### DIFF
--- a/src/components/table/table.ios.styl
+++ b/src/components/table/table.ios.styl
@@ -98,7 +98,6 @@
   th, td
     white-space nowrap
     padding 7px 24px
-    position relative
 
   thead, td, th
     border-style solid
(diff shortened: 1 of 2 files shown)


rstoenescu:
Thank you!

Lin84:
first Thx to You for this framework :)
second maybe too early to thx me :) because at the moment I am not sure why this bug exist, I hope it does not cause some difference bug.

rstoenescu:
The fix is safe. Relative position was there as a "bonus" anyway. If anyone needs it it can add it individually on the cell.

PR #3161, opened 2019-02-06
Validation improvements

TobyMosque:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [X] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**
`lazy-rules` is acting as a default value to the `rules`
the rules is still accepting `string` and `function` as validator.
Case the rule is a `string` as validator, the `component` will try to pick a `buildin` function, if none is available, the component will try to parse the string as a regular expression to use as validator.

As a third option, the rule can be a object, that object must have a property called handler, the handler must be a `string` or `function` and will be the validator described above.
The other options of that object are:
 - `lazy`: if true the rule will be trigged only when the component blur, if null, that will take the `lazy-rules` as value
 - `disabled`: if true the rule will be trigged only manually, if null, that will take the `disabled-rules` as value
 - `message`: accept a `string` or a `function`, the `function` must return and `string` and receive the label as argument
 - `label`: the `string` sent to the message function, if null, that will take the component' label as value.
 - `args`: a additional object who will be passed to the handler. that can be a `function`, in that case, must return a object, that property is useful to create more complex validations.
 - `handler`: the function who will avail if the field value is valid or not.

And the below properties as introduced
 - `disabled-rules`: if true, the rule will execute only the the method `.validate()` is called, that will act as a default value, so the rules can override it.
 - `validation-label`: can receive a `string` as value, if null, will be equal to the component label, your value will be passed to the `message` function of each rule, that will act as a default value, so the rules can override it.


quasar/src/components/chat/QChatMessage.js
```diff
@@ -1,6 +1,6 @@
 import Vue from 'vue'
 
-import slot from '../../utils/slot.js'
+// import slot from '../../utils/slot.js'
```
nothingismagick: why?

.gitignore
```diff
@@ -5,3 +5,4 @@
 debug.log
 package-lock.json
 .vscode/settings.json
+/.vs
```
nothingismagick: this is out of scope for the PR

quasar/src/mixins/validate.js
```diff
@@ -52,39 +53,97 @@ export default {
...
-        else if (typeof rule === 'string' && testPattern[rule] !== void 0) {
-          if (testPattern[rule](val) !== true) {
-            error = true
-            break
+        else {
+          try {
+            var regex = new RegExp(original)
+            rule.handler = (value) => regex.test(value)
           }
+          finally { }
```
nothingismagick: ??? a finally without a fall-through? using try without a catch?

nothingismagick (changes requested):
I don't see any changes to the docfile about this new behaviour, find this interesting as a sketch - but I don't think its ready for production.

quasar/src/mixins/validate.js
TobyMosque: how i would validate if a string is a valid regular expression?

quasar/src/components/chat/QChatMessage.js
TobyMosque: because slot is declared but never used, i know it also is out of scope of that PR.

rstoenescu:
Hi,

Thanks for contributing!

Quick question: what can you do having these improvements and you can't with current validations? Pinpoint it out. I honestly don't see the benefits. Otherwise it's just complicating things way beyond what they can be. Higher footprint, more code (so harder to maintain and possibly more bugs).

The async function may be a good idea. But it needs to stay sync unfortunately. Otherwise consecutive calls can be made and there's no guarantee that the methods will be fulfilled in the same order that they were called, which determines un unpredictable behavior.

quasar/src/components/chat/QChatMessage.js
rstoenescu: is this outdated? cause there's no such import in QChatMessage.

.gitignore
rstoenescu: this is ok

TobyMosque:
I undestood your point, but the current validation is way limited, i don't know how it can be reusable and cover common scenarios, like validate a range of values or compare a field to another. 

A other point, since the message must be the return of validation rule, that can be a problem if we wanna to support i18n.

So I want to do a much simpler suggestion, why not transform the validation in a addon and let the users decide if them will use it or add a third party lib as dependency.

Thank you very much for the feedback and I'm sorry if I seemed rude, unfortunately sometimes I can't express myself very well.

rstoenescu:
The rules are functions. What more freedom could you want? :)

Compare to another field and use vue-i18n for the message:
```js
val => val === myOtherField || $t('message.hello')
```

Issue #3214, opened 2019-02-11
[v1]  Discussion: Default date format ( YYYY/MM/DD ).


nanov:
**Background**
In v1, as opposed to previous versioning, everything that handles dates does not receive a format option, but rather forces you to comply with a default one. 

**Problem**
The choice of the default date format is ```YYYY/MM/DD```, which, at first glance, seems a rather strange decision as it does not comply with any standard.

The de-facto standard, ISO compliant date format is ```YYYY-MM-DD``` which seems a far better choice than the current one, which is also widely implemented in various REST apis, and part of the JSON-Schema standard formats. 

**Proposition** 
I would purpose to switch to this format as the default one. The sooner this is done ( before v1 is widely adopted/ out of beta ) it will eliminate the need to write multi-format/format-adjusting logic into components dealing with those and keep things simple.

pdanpdan:
The problem with YYYY-MM-DD is that the browser treat it (in some implementations) as date in local timezone and you'll get different dates in different timezones.
YYYY/MM/DD is always treated as date in UTC timezone.

https://stackoverflow.com/questions/30199130/different-result-for-yyyy-mm-dd-and-yyyy-mm-dd-in-javascript-when-passed-to-new

nanov:
From the linked issue it seems to be exactly the opposite, YYYY-MM-DD is treated as UTC ( and adjusted to local ) and YYYY/MM/DD is treated as local ( and therefore not adjusted ).

Timezone handling is something that the user needs to take care **before** passing the date to the components, as with both formats there are "specifics" he needs to take care of.

pdanpdan:
Sorry, I might have put it in reverse. Check this: https://github.com/quasarframework/quasar/issues/2658

nanov:
I did a (POC) pull request implementing my suggestion, as all the dates are handled as in the same timezone ( ie not string parsing of dates is done directly as ```new Date('2019/02/02')```) I really don't get point of using a not-standard format.

Additionally the an ```emit-object``` ( or similar ) flag could be added to the QDate component which would make it emit something like this: ```{ year: 2012, month: 5, date: 15 }``` and will allow one to work further with common date utils.

stefanvanherwijnen:
I agree that the ISO standard should be followed (i.e. YYYY-MM-DD).However, if JavaScript date messes up the interpretation of the dates in such a format, I see why a different format was chosen.
[...]

PR #3217, opened 2019-02-11
[v1] switch date format to YYYY-MM-DD 

nanov:
#3214


pdanpdan:
Already addressed in all v1 versions.

PR #3228, opened 2019-02-12
feat(docs): Select - Fix second example for "Create new values" 

hawkeye64:
This example did not work as advertised. Found by Boomer.


hawkeye64:
The last commit, I just optimized the code a bit.

PR #3333, opened 2019-02-20
feat(v1 docs) - Added a cookie notification.

webnoob:

- [ ] Bugfix
- [ x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x ] No

**Other information:**

This PR follows a discussion between myself (Allan Gaunt) and Denjell. As we're using cookies on the website (_ga) we need to have a cookie notification. This is also a pre-cursor to another PR that will contain a cookie.

### DIFF
--- a/docs/quasar.conf.js
+++ b/docs/quasar.conf.js
@@ -85,7 +85,11 @@ module.exports = function (ctx) {
         loadingBar: {
           color: 'amber'
         }
-      }
+      },
+
+      plugins: [
+        'Cookies'
+      ]
     },
 
     animations: ['fadeIn', 'fadeOut'],
--- a/docs/src/App.vue
+++ b/docs/src/App.vue
@@ -1,11 +1,55 @@
 <template>
   <div id="q-app">
     <router-view />
+
+    <transition
+      enter-active-class="animated fadeIn"
+      leave-active-class="animated fadeOut"
+      mode="out-in"
+      :duration="200"
+    >
+      <q-banner
+        v-if="showCookieMessage"
+        class="bg-primary fixed-bottom z-top"
+      >
+        <div class="row justify-between items-center">
+          <div class="text-white">
+            <span>This website uses cookies to ensure you get the best experience on our website. </span>
+            <a href="https://cookiesandyou.com" rel="noopener noreferrer nofollow" target="_blank" class="text-white">Learn more</a>
+          </div>
+          <q-btn
+            label="Got it!"
+            color="white"
+            text-color="black"
+            @click="hideCookieNotification"
+          />
+        </div>
+      </q-banner>
+    </transition>
   </div>
 </template>
 
 <script>
 export default {
-  name: 'App'
+  name: 'App',
+
+  data () {
+    return {
+      showCookieMessage: false
+    }
+  },
+
+  methods: {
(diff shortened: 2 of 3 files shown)


nothingismagick:
Background info: we need to adhere to the EU Cookie Law. What @webnoob made here is a direct result of that discussion and if need be we can discuss it on discord further.

smolinari:
To me, this is another great example of what a third party app extension should be. I'm not asking it be done as one right now, but there will be many other devs coming up with the need for this same solution (or similar), should they too be using cookies (and they probably will be). 😄 

Scott

webnoob:
You know, I didn't even think of that. Good call.

When app extensions get UI'ified (to the point we're happy) I can port it over.

panstromek:
This is just a notice that the site uses cookies but some sites are asking for explicit permission. Is there a difference? What is required?

webnoob:
You need to tell them cookies are being used. The fact they continue using the site means they agree to that. Legally, we're obliged to display the notification.

Check out: https://cookieconsent.insites.com/app/themes/insites-cookie-consent/examples/example-6-location.html

webnoob:
That being said, I can now see that in Germany (for instance), the choice needs to be revokable which this isn't. That being said, I've never seen one where it *is* revokable ...

docs/quasar.conf.js
```diff
@@ -85,7 +85,11 @@ module.exports = function (ctx) {
         loadingBar: {
           color: 'amber'
         }
-      }
+      },
+
+      plugins: [
```
rstoenescu: Quasar is configured to import everything already. See `all: true` above. This is redundant.

docs/src/App.vue
```diff
@@ -1,11 +1,55 @@
 <template>
   <div id="q-app">
     <router-view />
+
```
rstoenescu: Why all this code when you could use a `this.$q.notify()` call instead?
Cookies are also NOT needed. This can be handled with LocalStorage to avoid increasing the length of each HTTP request (which will include this cookie).

Also, in this example, the QBanner goes under the sidebars and footer. `z-max` would have been more appropriate.

rstoenescu (changes requested):

webnoob:
I'll take these points into account (didn't think of using `$q.notify` - good point).
Noted about local storage.
We've actually opted to make this an app extension now so I'll make the changes in there.
Thanks for the pointers.

rstoenescu:
An app extension might be too much for just a call to this.$q.notify.

nothingismagick:
We were thinking about hooking into the analytics and shutting them down if
no approval (in line with the EU Cookie Law)

rstoenescu:
Ok, let's see it first then we decide.

webnoob:
Ok, no problem. I'm looking at getting it done next week sometime as I'm away this weekend.

webnoob:
Ignore that last commit for QDrawer, I've done something wrong. Working it out now.

webnoob:
I'm closing this PR because I've messed up the commit and want to clear the branch on my tree.
Will do PR's for a Cookie App Ext (TBD) and the Drawer change.

PR #3363, opened 2019-02-22
fix(json-api): Fixes issue with recursive mixins

hawkeye64:
When a mixin references another mixin, then the first mixin is not added to the json api. Only the last mixin makes it in.


panstromek:
Which components are affected by this?

PR #3393, opened 2019-02-25
fix(@quasar/app): Stop overlay affecting visibility state #3381

### LINKED ISSUE #3381: [v1] When setting overlay on a QDrawer it hides it by default
See this codepen:

https://codepen.io/webnoob/pen/gEYwOz?&editable=true&editors=101

If you remove the `overlay` prop, it will show the QDrawer properly based on the initial value of the `v-model`.

The code:
```
showing = this.behavior !== 'mobile' && this.breakpoint < this.layout.width && !this.overlay
        ? largeScreenState
        : false
```

Doesn't really make sense in this instance as overlay should only mean it goes over elements, not affect it's state.

webnoob:

- [x ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ x] No

**Other information:**

fix #3381 

I considered changing the docs to note this change but I figured it's such an edge case that it'd be self explanatory. If you set `overlay` on the drawer, it's going to overlay as the name implies.

Also worth noting I was considering not adding the PR for this as it now causes the mini drawer to overlap the content if the initial value is `true` but at least this way, it's up to the developer. It's not imposed on them that setting `overlay` means the drawer is hidden by default.

I've tested this vs old functionality and can't see any unintended side effects.

### DIFF
--- a/quasar/src/components/layout/QDrawer.js
+++ b/quasar/src/components/layout/QDrawer.js
@@ -65,7 +65,7 @@ export default Vue.extend({
       largeScreenState = this.showIfAbove || (
         this.value !== void 0 ? this.value : true
       ),
-      showing = this.behavior !== 'mobile' && this.breakpoint < this.layout.width && !this.overlay
+      showing = this.behavior !== 'mobile' && this.breakpoint < this.layout.width
         ? largeScreenState
         : false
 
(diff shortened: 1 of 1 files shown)


webnoob:
Might be worth noting, I've found away to achieve the same result without this PR so if you're worried about knock on effects then it can be worked around.

I personally don't think a prop like `overlay` should affect element position but it's up to you.

Note, the work around would be setting `:overlay="!miniState"` so it only goes into overlay mode when it's being hovered.

PR #3396, opened 2019-02-25
[Feature] Added defaultView prop to QDate

stefanvanherwijnen:

- [x] Feature

- [x] No

Similar to the prop that was available in v0.17. When selecting a birth date it is easy to forget to change the year after selecting a day and month. For that reason, it is better to start with selecting the year.

I could not get the docs to run locally so I was not able to check them (exports is read-only error...). But the functionality in npm run dev works fine.


rstoenescu:
Thank you for contributing!
This will be available in beta.11.

Issue #3478, opened 2019-03-05
[v1] QDialog: Toggling 'maximized' property while hidden still locks scroll


abruere:
Thanks for Quasar!

This is a minor issue related to maximized dialog. I’m submitting a PR too.

This is loosely related to #3310.

### quasar-v1.0.0-beta.8

Operating System - Linux(4.4.0-142-generic) - linux/x64
NodeJs - 8.11.2

Global packages
  NPM - 5.6.0
  yarn - 1.12.3
  @quasar/cli - 1.0.0-beta.2
  cordova - Not installed

Any other software related to your bug:

### JsFiddle

Codepen: https://codepen.io/anon/pen/YgGMqe?editors=1010

### What did you get as the error?

Toggling `maximized` on a hidden QDialog locks body scroll.

### What were you expecting?

Scroll lock should only apply when QDialog is visible. Maximized can be toggled in background depending on layout or screen size.

### What steps did you take, to get the error?

Please have a look at the codepen, where I reproduce using:
- Screen size based `maximized` property
- Manual toggle

nothingismagick:
Can you confirm that this is still happening with Quasar v1.0.0-beta.8 ?

>

abruere:
Yes I’ve just updated the codepen (forked from components docs) from v1.0.0-beta.7 to v1.0.0-beta.8, thanks.

rstoenescu:
@abruere Thx for working on a PR! Will be available in beta.9

abruere:
Thanks for building Quasar @rstoenescu!

PR #3479, opened 2019-03-05
Fix: #3478 Toggling hidden QDialog 'maximized' property locks scroll

abruere:
Scroll lock now only applies when QDialog is visible so that `maximized` can be toggled in background depending on layout or screen size.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

#3478 [codepen](https://codepen.io/anon/pen/YgGMqe?editors=1010#0)

### DIFF
--- a/quasar/src/components/dialog/QDialog.js
+++ b/quasar/src/components/dialog/QDialog.js
@@ -88,12 +88,14 @@ export default Vue.extend({
     },
 
     maximized (newV, oldV) {
-      this.__updateState(false, oldV)
-      if (this.value) this.__updateState(true, newV)
+      if (this.showing === true) {
+        this.__updateState(false, oldV)
+        this.__updateState(true, newV)
+      }
     },
 
     seamless (v) {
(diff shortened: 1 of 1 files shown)


nothingismagick:
Thanks for the PR - I haven't tested it yet but my main comment is that this behaviour is something that should definitely be mentioned in the documentation if not also in the API.

abruere:
Thanks for feedback, then I’ll also try to add the checkbox toggle in the docs too (as in codepen) to make it easier to test.

If some users might rely on this always-lock feature currently (although a hidden element’s property affecting whole page scroll is not something we usually expect and sounds like a bug to me), do you want me to mark this as a breaking change?

nothingismagick:
There may not be breaking changes. You will have to restructure the code modification so that this behaviour is optional.

rstoenescu:
Good point with the PR. Thx for it!

PR #3514, opened 2019-03-08
fix(quasar): Fix wheel event beeing passive by default in chrome

pdanpdan:
See wheel scroll over body on v1 docs site when left drawer is open.


euskalhenriko:
Hi, I'm using the v0.17.20 of framework and when I have a q-modal opened, I have the same issue.
Is solved only in the v1?
If yes, is possible to solve in the previous version too?
Thank you very much!

PR #3634, opened 2019-03-20
Proof of concept panel keep alive

pdanpdan:


diegomagikal:
Thanks. I need this. This works great for prevent losing files in q-uploader queue, when the user changes the tab without send the files.

PR #3636, opened 2019-03-20
feat(docs): Custom hover effects example

hawkeye64:
Based on codepen I wrote: https://codepen.io/Hawkeye64/pen/OqBjxM


rstoenescu:
Clicking on such a QBtnDropdown will simply hide then show the dropdown.

PR #3649, opened 2019-03-22
[v1][QDate] Fix default view with null model in Persian calendar mode

behnoodk:
When model is `null`, default view should be current month and year. Current Persian month and year can safely be inferred from current Gregorian full date and not from first day of current Gregorian month.

- [x] Bugfix

- [ ] Yes
- [x] No


rstoenescu:
Thanks!
Will be available in beta.12

Issue #3686, opened 2019-03-26
Feature Request: Allow classes in QTable to be dynamically derived


infintesimal:
Hi,
I would like to be able to dynamically derived additional classes for table columns as follows
```
<template>
  <div class="q-pa-md">
      <q-table
        title="Results"
        :data="fontData"
        :columns="fontColumns"
        row-key="fontData.name"
        binary-state-sort
        dense
      ></q-table>
    </div>
</template>

<script>
export default {
  name: 'FontVoteTally',
  data () {
    return {
      fontColumns: [
        {
          name: 'name',
          required: true,
          label: 'Font Name',
          align: 'left',
          field: row => row.name,
          format: val => `${val}`,
          sortable: true,
          style: 'width: 200px; font-size: 2em;',
          classes: ['text-weight-bold', row => row.specialClassName1]
        },

```
Currently, this is not possible (at least) because of the direct translation of col.classes here
[table-body.js](https://github.com/quasarframework/quasar/blob/dev/quasar/src/components/table/table-body.js#L43)
[table-header.js](https://github.com/quasarframework/quasar/blob/dev/quasar/src/components/table/table-header.js#L58)

I am submitting a pull request that addresses this shortly.

infintesimal:
Convincing reasons:
1) The code works
2) There are plenty of developers who have code where desired css classes can be inferred from dynamically return data. E.g. the data my return a class name, an array of class names or something derivable, such as
```font-name: Foo``` and the developer maintains a contract with stykesheets such that the class that specfies ```font-face: Foo``` always has a name like ```.font-foo```. Thereafter, a quasar developer can easily add a function on QTable that would be like ```getClasses(row) { return row.className }``` or ```getClasses(row) { return ArrayOfClasses }```

rstoenescu:
Hi,

First of all, thank you for providing a PR! Much appreciate it.

Unfortunately, it's not really optimal. There's all sorts of slots that you can use in order to customize the QTable in case there is a scenario that can't be covered by the default set of options. Of course, I can add this PR, but the concern is that it adds overhead for 99% of the use cases. We need to keep this component to render as fast as possible hence for custom scenarios there are the slots.

Looking forward to other PRs from you though.

PR #3687, opened 2019-03-26
fix: 3686

infintesimal:
… https://github.com/quasarframework/quasar/issues/3686

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Here is a diff patch for the documentation update
```diff --git a/source/components/datatable.md b/source/components/datatable.md
index 044276d..f3a3d1e 100644
--- a/source/components/datatable.md
+++ b/source/components/datatable.md
@@ -167,6 +167,10 @@ columns: /* array of Objects */ [
 
     // v0.17.9+; if using scoped slots, apply this yourself instead
     style: 'width: 500px',
+    //  classes: can be any of:
+    //  classes: 'class1 class2 ...'
+    //  classes: ['class1', 'class2', row => this.fxReturnsClass(row) ... ]
+    //  classes: row => this,fxReturnsClassOrArrOfClasses(row)
     classes: 'my-special-class'
   },
   { name: 'calories', label: 'Calories', field: 'calories', sortable: true },
```

### DIFF
--- a/quasar/src/components/table/table-body.js
+++ b/quasar/src/components/table/table-body.js
@@ -40,7 +40,7 @@ export default {
                   : h('td', {
                     staticClass: col.__tdClass,
                     style: col.style,
-                    class: col.classes
+                    class: this.getClasses(col, row)
                   }, this.getCellValue(col, row))
               })
 
@@ -114,6 +114,31 @@ export default {
     getCellValue (col, row) {
       const val = typeof col.field === 'function' ? col.field(row) : row[col.field]
       return col.format !== void 0 ? col.format(val) : val
+    },
+
+    getClasses (col, row) {
+      let colClasses = []
+      switch (typeof col.classes) {
+        case 'object':
+          if (Array.isArray(col.classes)) {
+            col.classes.forEach(function (elem) {
+              switch (typeof elem) {
+                case 'function':
+                  colClasses.push(elem(row))
+                  break
+                default:
+                  colClasses.push(elem)
+              } 
+           })
+          }
+          break
+        case 'function':
+          colClasses.push(col.classes(row))
+          break
+        default:
+          return col.classes
+      }
+      return colClasses
     }
   }
 }


rstoenescu:
Hi,

First of all, thank you for providing a PR! Much appreciate it.

Unfortunately, it's not really optimal. There's all sorts of slots that you can use in order to customize the QTable in case there is a scenario that can't be covered by the default set of options. Of course, I can add this PR, but the concern is that it adds overhead for 99% of the use cases. We need to keep this component to render as fast as possible hence for custom scenarios there are the slots.

Looking forward to other PRs from you though.

infintesimal:
Hi Razvan,
I am of course disappointed that you decided to close this so quickly.

I take issue with the "concern" of overhead. Is there a set of objective benchmarks that you use to determine overhead? If so, what are they? That way I could test my proposed changes against said benchmarks.

From my point of view, for a internet connected app, there are multiple points of overhead:
1) Too many or unnecessary bytes moving across a network wire/connection
2) Too many CPU op codes invoked, slowing down rendering in case of client side JS
3) Too much data stored in memory, creating unwanted read/write malloc and calloc times
4) Too much disk IO
All of these categories apply to both server side and client side computing. Which one did I violate?

I should point out that CPU throughtputs for contemporary mobile devices is around 1GHz; Network throughput for the typical consumer is about 5Mb/second (download). If we stretched our imaginations and thought of the cost of adding one byte one the network wire vs. adding one op code on a cpu, I could easily argue that the "penalty" of an extra byte of network traffic is no different than that of 160 CPU cycles.
Yes, the slots are available, but it also means I am pushing many more bytes of template code to achieve the same result.

My ask, if you do truly want future contributions from me, is that you create a set of objective performance benchmarks that any PR has to satisfy. That way, a rejection won't fee so capricious.

Thanks for your consideration.
Steve

rstoenescu:
It's about the Vue render speed of QTable, not about disk IO or transferring more bytes over the network.
QTable operates with rows, so every extra operation becomes costly when the set of data is large. Your additions might not matter so much on latest and greatest CPU on a desktop, but for low end phones it matters quite a lot. That extra function will get executed on each QTable render -- which happens a lot, on any kind of change to the Vue VM. My point is that this feature will be disregarded by 99% of devs while it will add this render speed overhead for 100% of use cases. This is one of the reason why slots were added to QTable --> if you need something custom, then just use the slots. Let's not bloat the column definition when there's a way better method to handle this. The column definitions are a helper for most use cases of QTable but should not replace everything you can do with a slot. Yeah, we can add a lot of stuff to the column definitions, like even "render this column with X component", but the cost of render operation increases. If anything, the render speed should increase, not the other way around it. With slots, the render operation cost is kept to a minimum -- you as a developer know exactly what you want to render, so calling extra functions in QTable's render() and treating lots of "if" cases does not apply.

infintesimal:
If I prove you wrong re the render speed (effectively CPU ops) will you take the PR? :-)I too saw that this is invoked at every render of a column and I had given this thought, to be honest.I thought maybe I should "fix" (meaning to my design liking) the entire abstraction of rows and columns and their attributes/properties.  But then I realized the default pagination is 5, meaning that said render opp would only happen for columns affected * # of rows. Since most mobile devs aren't stacking 20 or 30 columns in, more like 5 or 6, it's really only 100 extra ops (at max) (keep in mind my default is to return col.classes as is). I could make the default test the first, so we'd be talking about the overhead of an if for every cell rendered for what you assert is the dominant use case.
Cheers,Steve
    On Tuesday, March 26, 2019, 9:16:11 AM PDT, Razvan Stoenescu <notifications@github.com> wrote:  
 
 
It's about the Vue render speed of QTable, not about disk IO or transferring more bytes over the network.
QTable operates with rows, so every extra operation becomes costly when the set of data is large. Your additions might not matter so much on latest and greatest CPU on a desktop, but for low end phones it matters quite a lot. That extra function will get executed on each QTable render -- which happens a lot, on any kind of change to the Vue VM. My point is that this feature will be disregarded by 99% of devs while it will add this render speed overhead for 100% of use cases. This is one of the reason why slots were added to QTable --> if you need something custom, then just use the slots. Let's not bloat the column definition when there's a way better method to handle this. The column definitions are a helper for most use cases of QTable but should not replace everything you can do with a slot. Yeah, we can add a lot of stuff to the column definitions, like even "render this column with X component", but the cost of render operation increases. If anything, the render speed should increase, not the other way around it. With slots, the render operation cost is kept to a minimum -- you as a developer know exactly what you want to render, so calling extra functions in QTable's render() and treating lots of "if" cases does not apply.

rstoenescu:
I do not want to encourage this behaviour. It's then going to get requested for style too, then for x, y, z because it becomes inconsistent. Add this then that then that then... In this case, since it's very specific and there are slots designed especially for this scenario, it doesn't makes sense to open Pandora's box. Been there a lot of times. Saying that the default rows per page is 5 has no meaning when dealing with a framework source-code. It's not gonna get used only with 5 rows per page. It can be any number of rows (even all). The code needs to run as fast as possible.
I've merged a LOT of PRs into the framework and I'm really glad whenever I see new PRs, however if you need help with slots I can create a Codepen for you with this specific use case.

infintesimal:
Razvan,First and foremost, I appreciate Quasar Framework. It's made my life much easier. Thank you and your team for all the hard work.
I understand the code performance concerns. I understand bloat concerns. All really good frameworks end up in the bloat heap eventually (that's where good and useful code goes to die). Django comes to mind as an example. So does angular and to a lesser degree, react. Will likely happen to Vue one day too. :-(
I also understand the "one purpose." I disagree that my proposed change was single purpose but don't give a crap about the disagreement, I can always maintain my fork as I see fit.
What I do give a crap about is data-driven decisions. If you are going to reject PRs on performance concern, how about having some performance benchmarks all code needs to pass? We could cobble together a reference site and then measure the performance profile pre and post a change. I've worked at companies before where any new feature had to be cost-neutral. In other words, any new feature had to evict an old feature if the performance time increased (now the way that was carried out in practice was that as features cumulated up to 250ms people would start searching for what to prune). Now an early stage open-source project cannot and should not be that austere, but having CI/CD performance profiling steps certainly enables it (and thus increasing the time to the eventual bloat heap of history). My further point on performance profiling was that looking at just things like "paint time" "render time" "recalculation time" "garbage collection time" is a narrow view of performance. It assumes that network time is free. And that is false. If a given feature adds, let's say, 10ms of client side compute and render time but saves 100ms of network delivery time, then the end user is 90ms net positive.
5Mbs/seconds = 625,000 bytes/sec = 1 byte for every 1.6 * 10^-6 seconds (pretending no network latency). We could, to make the math simple, call that 2 bytes per microsecond. If a punting on a particular feature results in an extra 100 bytes of code (enumerating trs and tds with named slots, e.g.) that has to travel over a network wire (again, we'll treat local bus traversal as approximating 0) means there is a net hit of 50 microseconds.
For what it is worth, I did some testing on my machine and here is what I got on the compute side:No dynamic rendering of classes: Avg render time of a 5x4 table: 590 microseconds, s.d. 105 microsecondsDynamic rendering of classes: Avg render time of a5x4 table: 813 microseconds, s.d. 127 microsecondsSample size: 20 in each category
I'll do a version with slots and look at Network time too to figure out the net. I can probably squeeze more microseconds out of the code. :-)
Cheers-s

    On Tuesday, March 26, 2019, 2:09:22 PM PDT, Razvan Stoenescu <notifications@github.com> wrote:  
 
 
I do not want to encourage this behaviour. It's then going to get requested for style too, then for x, y, z because it becomes inconsistent. Add this then that then that then... In this case, since it's very specific and there are slots designed especially for this scenario, it doesn't makes sense to open Pandora's box. Been there a lot of times. Saying that the default rows per page is 5 has no meaning when dealing with a framework source-code. It's not gonna get used only with 5 rows per page. It can be any number of rows (even all). The code needs to run as fast as possible.
I've merged a LOT of PRs into the framework and I'm really glad whenever I see new PRs, however if you need help with slots I can create a Codepen for you with this specific use case.

rstoenescu:
We're talking about totally different things :) I am talking about the Vue render speed -- the speed with which Vue handles the "render()" method. It's got nothing to do with the Network tab (or Performance tab) in dev tools -- you got a correct analysis there, but it doesn't applies to our main issue here. Maybe we can connect next week on Discord.

PR #3727, opened 2019-03-28
Button icon should now get padding when there is no label, if there is a slot

ymhr:
**Description**
Currently, the `q-icon` in the `q-btn` component will only get the `on-left` class if a `label` property is present. In my application (and I am sure many others), the slot is used for the label, rather than this prop, which means that there is no spacing between the icon and the content.

This PR changes that so that if _either_ the `label` prop __**or**_ a slot is present, the icon receives `on-left`.

**Before:**
Screenshot 2019-03-28 at 15 53 20

**After**
Screenshot 2019-03-28 at 15 59 10

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/quasar/dev/components/components/button.vue
+++ b/quasar/dev/components/components/button.vue
@@ -338,6 +338,9 @@
       </div>
       <div class="q-gutter-sm">
         <q-btn :type="tag" color="primary" icon="mail" label="On Left" />
+        <q-btn :type="tag" color="white" text-color="black" icon="mail">
+          On left, slot instead of label
+        </q-btn>
         <q-btn :type="tag" color="secondary" icon-right="mail" label="On Right" />
       </div>
 
--- a/quasar/src/components/btn/QBtn.js
+++ b/quasar/src/components/btn/QBtn.js
@@ -21,6 +21,9 @@ export default Vue.extend({
   computed: {
     hasLabel () {
       return this.label !== void 0 && this.label !== null && this.label !== ''
+    },
+    hasSlotContent () {
+      return this.$slots && this.$slots.default && this.$slots.default.length > 0
     }
   },
 
@@ -112,7 +115,7 @@ export default Vue.extend({
     if (this.icon !== void 0) {
       inner.unshift(
         h(QIcon, {
-          props: { name: this.icon, left: this.stack === false && this.hasLabel === true }
+          props: { name: this.icon, left: this.stack === false && (this.hasLabel === true || this.hasSlotContent) }
         })
       )
     }


rstoenescu:
Hi,

Thank you for contributing!

There are a few problems though:
1. You should never use a computed property to determine if a slot is present or not. `$slots` is not reactive.
2. QIcon has `left` and `right` Boolean props specifically for this reason. Quasar should not interfere with the custom content of a slot. It may add good functionality for some, but break some custom use case for others.

```html
<q-btn ...>
  <q-icon left name="map" />
  My label
</q-btn>
```

ymhr:
@rstoenescu thanks for the quick reply:
1. Understood, could easily run that logic inside the render function instead
2. My change would not affect the code example you have there, take this example:
```vue
<q-btn icon="mail" label="test"/>
```
In this example, the `icon` gets the `on-left` class to provide padding automatically
```vue
<q-btn icon="mail">test</q-btn>
```
in this example, the icon does _not_ get the padding, and therefore does not look right. Since the default slot is essentially a label slot, I think this is consistent behaviour.

Your example is a different way to work around this issue, and achieves the same result, but I believe the original behaviour is inconsistent from what is expected.

rstoenescu:
You are completely disregarding point no. 1 (`You should never use a computed property to determine if a slot is present or not. $slots is not reactive.`).

The QBtn is designed to **either** work with icon/label props or by using the default slot. The default slot can also be used to inject a QTooltip for example, so it's not a reliable method to detect if it contains a label.

ymhr:
I am not disregarding point #1, I have pushed an update and even addressed it in my comment. I am aware that it does not show the updates here so perhaps there is something I missed. My update has moved it to a method instead.

Understand your point about inserting other things into the slot, is there a different solution that can cover all of these bases perhaps? What about a new prop on the button to add left padding to an icon, e.g. `icon-left` so that it can support both use cases?

rstoenescu:
What's wrong with the current design? Either use icon/label props or slot content with a QIcon and text. Let's follow the KISS approach. It's almost impossible to take into account all possible use cases of the default slot (look in docs, there's some examples of custom default slot content, and that's just the tip of the iceberg) -- and in doing so, we'd add a lot of complexity which is simply unneeded and would also decrease the Vue render speed for one of the most used components in an app. We must ask ourselves: is it worth sacrificing performance and risk of introducing bugs just so that we can write: `<q-btn icon="map">Map</q-icon>` instead of `<q-btn> <q-icon name="map" left /> Map </q-btn>` ? I think not.

PR #3756, opened 2019-03-31
clearable icon: close-cirle  (suggestion)

mesqueeb:
Just a preference suggestion:
image

I feel like the circle one is a better option.


rstoenescu:
Indeed, the circle variant is specified in Material Guidelines.
Would have preferred all iconSets be changed to circle, but I've merged this and will do the rest.

mesqueeb:
Sorry, I forgot about the other iconSets! if you want I can still do it, let me know!

rstoenescu:
It's almost ready. Will push it in 1 minute :)

PR #3793, opened 2019-04-03
Update vuex-store.md

Phantomkeks:
- [X] Bugfix

- [X] No

**Other information:**
It needs to be 'Store' in line 82 otherwise the example throws an error.


rstoenescu:
Thanks for catching this!

PR #3840, opened 2019-04-09
Update the translation for "Cancel" in Ukrainian

neskhodovskiy:
I am a native Ukrainian speaker. I suggest the word which sounds more authentic to the Ukrainian language, replacing the artificial late borrowing from Russian.


rstoenescu:
Thanks!

rstoenescu:
Will go into beta.16

Issue #3894, opened 2019-04-15
[v1] Quasar initialization problem with test components


Abuntxa:
### Software version
Operating System - Windows_NT(10.0.17134) - win32/x64
NodeJs - 10.15.3

#### Global packages
  NPM - 6.4.1
  yarn - 1.12.3
  @quasar/cli - 1.0.0-beta.4
  cordova - Not installed

#### Important local packages
  quasar - 1.0.0-beta.17 -- High performance, Material Design 2, full front end stack with Vue.js -- build SPA, SSR, PWA, Hybrid Mobile Apps and Electron apps, all simultaneously using the same codebase
  @quasar/app - 1.0.0-beta.17 -- Quasar Framework App CLI
  @quasar/extras - 1.1.1 -- Quasar Framework fonts, icons and animations
  vue - 2.6.10 -- Reactive, component-oriented view layer for modern web interfaces.
  vue-router - 3.0.4 -- Official router for Vue.js 2
  vuex - 3.1.0 -- state management for Vue.js
  electron - Not installed
  electron-packager - Not installed
  electron-builder - Not installed
  @babel/core - 7.3.4 -- Babel compiler core.
  webpack - 4.30.0 -- Packs CommonJs/AMD modules for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand. Support loaders to preprocess files, i.e. json, jsx, es7, css, less, ... and your custom stuff.
  webpack-dev-server - 3.3.1 -- Serves a webpack app. Updates the browser on changes.
  workbox-webpack-plugin - 4.2.0 -- A plugin for your Webpack build process, helping you generate a manifest of local files that workbox-sw should precache.
  register-service-worker - 1.6.2 -- Script for registering service worker, with hooks

#### Quasar App Extensions
  @quasar/quasar-app-extension-testing-unit-jest - 1.0.0-beta.22 -- A Quasar App Extension for running Jest tests
  @quasar/quasar-app-extension-testing - 1.0.0-rc.5 -- A Quasar App Extension for managing Test Harnesses
### Repository reproducing the problem
[https://github.com/Abuntxa/quasar-testing](https://github.com/Abuntxa/quasar-testing)
It's a repository with the minimal case.
* It's a basic bootstrapped project with the unit testing extension (jest)
* It has been modified src/pages/Index.vue to make use of the plugin localStorage
* There have been added two exactly indentical tests of the component
* Before each test the component to be tested is recreated and destroyed

[...]

PR #3895, opened 2019-04-15
fix: Correct component test instantiation (fix: #3894)

Abuntxa:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
The reproduction steps are explained in the bug.
It's been tested that after the correction the bug is not reproduced and the spa application works properly.
The fix makes publicly available the flag of instalation of Quasar through the Vue instance that might not be the best solution, therefore supporting multiple instalations of Quasar might be a better solution.

### DIFF
--- a/quasar/src/install.js
+++ b/quasar/src/install.js
@@ -17,8 +17,8 @@ export const $q = {
 }
 
 export default function (Vue, opts = {}) {
-  if (this.__installed) { return }
-  this.__installed = true
+  if (Vue.__quasarInstalled) { return }
+  Vue.__quasarInstalled = true
 
   const cfg = opts.config || {}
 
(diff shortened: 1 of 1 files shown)


nothingismagick:
Hey, thanks for the proposal - and really digging in. But we will need to make sure that this has no unintended side-effects, especially with SSR - but also with Cordova and Electron.

Abuntxa:
Thanks for looking into it.
I also saw that there were some other plugins that stored the same way their installation state (when I did a global search of __installed)

rstoenescu:
Ah, sorry for not merging this earlier. It will be available with "quasar" v1.5.3.

PR #3974, opened 2019-04-25
Update mask.json

olegvakulenko:
Add more mask examples.
Is exist more options for this?


rstoenescu:
Thanks for this.
Yes, there are more. Added and merged. https://github.com/quasarframework/quasar/blob/dev/quasar/src/mixins/mask.js

PR #4001, opened 2019-04-30
fix: #3999 bug in infinite scroll behaviour

### LINKED ISSUE #3999: [V1] improve animation on infinite QCarousel
### JsFiddle
https://codepen.io/modchen/pen/LvwqMJ

### What steps did you take, to get the error?

Just swiping always to the next slide...

### What were you expecting?

When swiping "next" on the last slide (3), the transition get inverted to "prev", so it "jumps" back to (1) instead of continuing the transition to next slide. 

This breaks the infinite feeling, and it looks like, 1 is not next to 3 but before...

https://github.com/quasarframework/quasar/blob/a762d771505bdd99b0f014fdb491088cfad736f4/quasar/src/mixins/panel.js#L67

This line of code needs to take care of the "infinite" option

m0dch3n:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

More information about the problem can be found in 

https://github.com/quasarframework/quasar/issues/3999

### DIFF
--- a/quasar/src/mixins/panel.js
+++ b/quasar/src/mixins/panel.js
@@ -62,11 +62,16 @@ export const PanelParentMixin = {
           : -1
 
       if (this.animated) {
+        const oldWasFirstPanel = this.__getPanelIndex(oldVal) === 0,
+          oldWasLastPanel = this.__getPanelIndex(oldVal) === this.__getAvailablePanels().length - 1,
+          newIsFirstPanel = index === 0,
+          newIsLastPanel = index === this.__getAvailablePanels().length - 1
+
         this.panelTransition = validNewPanel === true && this.panelIndex !== -1
           ? 'q-transition--' + (
             index < this.__getPanelIndex(oldVal)
-              ? this.transitionPrev
-              : this.transitionNext
+              ? this.infinite && oldWasLastPanel && newIsFirstPanel ? this.transitionNext : this.transitionPrev
+              : this.infinite && oldWasFirstPanel && newIsLastPanel ? this.transitionPrev : this.transitionNext
           )
           : null
       }


rstoenescu:
Hi,

Thank you for contributing.
There are some serious performance problems with this approach:

* You do not guard oldVal with __isValidPanel (it might be not!)
* You call multiple times `this.__getPanelIndex(oldVal)` and `this.__getAvailablePanels()`. You could have stored the result in a variable rather than executing them multiple times.
* You always determine oldWasLastPane, newIsFirstPanel, newIsLastPanel even though there are cases where you might not need them.
* But the biggest problem is that calling this.__getAvailablePanels() is costly and breaks down performance.

Sorry, but will close this PR as a solution needs another approach.

m0dch3n:
@rstoenescu 

Can you check, if this satisfying you, before I do a PR ?

https://github.com/quasarframework/quasar/compare/dev...m0dch3n:dev

- oldVal is now guarded
- vars, instead of multiple calls
- I only determine everything now in `this.infinite && validNewPanel && validOldPanel && index > -1 && oldIndex > -1`
- `this.__getAvailablePanels()` does not cost more then `this.__getPanelIndex`

PS: If this is still not ok, I think and it's better that someone else, is correcting https://github.com/quasarframework/quasar/issues/3999 because, as I already mentioned there, I don't know the quasar framework enough to fix the issue by myself...

m0dch3n:
BTW

is there a reason, why the transition direction needs to be determined in the `watch` of `value` ? 

I'm asking this, because my solution only works for 3 or more panels. If you have only 2 panels in total,  in the infinite mode, you can't determine the direction only by the indexes...

You can get from panel 1 to 2, by swiping left **AND** right

So maybe panel.js should be refactored completly to not determine the transition by indexes, but at a point, where we still have the direction information....

How about simply puting the transition direction in the __go method, like here ?

https://github.com/quasarframework/quasar/compare/dev...m0dch3n:transitionTest

m0dch3n:
I did some more commits on the refactoring approach branch, I think that's now, the best I can do in

https://github.com/quasarframework/quasar/compare/dev...m0dch3n:transitionTest

Issue #4002, opened 2019-04-30
QSlider circle marker missing when `label-always` enabled


2x2xplz:
This was mentioned on Discord, and it was requested to open a new issue.

When `label-always` is enabled, the circle marker on the QSlider is not visible "at rest" or while sliding. It appears briefly after letting go of the slider, then it disappears again. 

This behavior is viewable in the docs page for `QSlider`, in the section for "always display label." 

(On Chrome/Android, the marker doesn't even show up briefly, in my test.)

### Software version

Quasar: 1.0 beta 19
@quasar/app (v1+ only):
OS: 
Node:
NPM:
Browsers: Chrome, Firefox on Windows, Chrome on Android
iOS:
Android:
Any other software related to your bug:

rstoenescu:
Hi, this is per design and as per Material Guidelines at some point (I see no mention of this in the guidelines at the moment I am writing though).

rstoenescu:
The change will be available in Quasar beta.21. Thanks for doing a PR!

drogerie21:
Hi, 
need this functionality. we want to display the thumb always, wheater the thumb is moved or not! Above the thumb-always option is mentioned but in quasar 2 this option is not documented and it doesn´t work!
Any chance this feature will be implemented?

rstoenescu:
@drogerie21 Please check the QSlider API card on the docs. I think that you are looking for label-always prop. There are examples of it throughout the QSlider page too.

drogerie21:
> @drogerie21 Please check the QSlider API card on the docs. I think that you are looking for label-always prop. There are examples of it throughout the QSlider page too.

I configured my slider with the label-always property. The problem i have is that the thumb disappears when i click on the thumb and move it. This is - for me - an unwanted behaviour. So i want to have an option like "thumb-always" that doesn't hide the thumb when moving it. in quasar v2 there is no such option, altough it says above that this has been added.

MichaelUray:
The point at the slider just disappears if it gets clicked/dragged by the mouse.
![ezgif-2-304d0b0f64](https://github.com/quasarframework/quasar/assets/25169478/296862d5-2a07-4bd8-a46c-5f4ef58e3b2f)

I don't think that this is the desired behavior.

rstoenescu:
This is exactly what Material Design Guidelines wanted....

MichaelUray:
> This is exactly what Material Design Guidelines wanted....

What is actually the benefit of this behavior if it is wanted?

[...]

PR #4008, opened 2019-04-30
added a `thumb-always` option (issue #4002)

2x2xplz:
The default behavior is to hide the slider thumb (the circular marker) whenever a label is present. This can be confusing when the `label-always` option is active, because the thumb is never visible, and it doesn't give the user the expected visual cue that sliding is possible. Even if the user knows, the lack of a thumb makes the visual target (for the mouse, or finger) seem much smaller. Having a `thumb-always` option will keep the thumb visible even with a label, if the developer wants it.

- [ ] Bugfix
- [x ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x ] No

- [x ] It's submitted to the `dev` branch and _not_ the `master` branch
- [x ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)

**Other information:**

https://github.com/quasarframework/quasar/issues/4002

There seems to be confusion on whether the hiding the thumb while a label is shown complies with Material Guidelines. @rstoenescu says the invisible thumb is compliant. @smolinari (on Discord) 
says it is a bug. This PR lets the developer decide, by enabling the `thumb-always` option. If the option is not specifically enabled then this PR will have no effect.

### DIFF
--- a/quasar/src/components/slider/QSlider.json
+++ b/quasar/src/components/slider/QSlider.json
@@ -60,6 +60,12 @@
       "category": "behavior|content"
     },
 
+    "thumb-always": {
+      "type": "Boolean",
+      "desc": "Always display the thumb - the default is to hide it when a label is visible",
+      "category": "behavior|content"
+    },
+
     "markers": {
       "type": "Boolean",
       "desc": "Display markers on the track, one for each possible value for the model",
--- a/quasar/src/components/slider/QSlider.json
+++ b/quasar/src/components/slider/QSlider.json
@@ -60,12 +60,6 @@
       "category": "behavior|content"
     },
 
-    "thumb-always": {
-      "type": "Boolean",
-      "desc": "Always display the thumb - the default is to hide it when a label is visible",
-      "category": "behavior|content"
-    },
-
     "markers": {
       "type": "Boolean",
       "desc": "Display markers on the track, one for each possible value for the model",
--- a/quasar/src/components/slider/slider-utils.js
+++ b/quasar/src/components/slider/slider-utils.js
@@ -55,7 +55,6 @@ export let SliderMixin = {
 
     label: Boolean,
     labelAlways: Boolean,
-    thumbAlways: Boolean,
     markers: Boolean,
     snap: Boolean,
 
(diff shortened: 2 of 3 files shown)


smolinari:
Quasar: 

image

Material spec: 

image

Scott

rstoenescu:
`<rant>`
Yeah, good luck keeping up to date with Material Guidelines **which get changed weekly**. And good luck with the guidelines which don't take into labels bigger than 3 chars long.
Rather than releasing v1 final I'll just spend time redoing this component a million times.
`</rant>`

Will analyze this as soon as possible.

smolinari:
Even Google doesn't follow their own spec all the time. 😄 So, for example, not making the label look like a teardrop I wouldn't call out as not following the spec (and not my intention with the images). The tear drop is just how they made it look in the docs. And as we all know, the design or "the looks" of components are supposed to be customizable (see the "Themes" shown at the bottom of the spec doc pages, for example). 

What I was trying to point out was, I think the "dot" (or as @2x2xplz puts it, the thumb) is missing underneath the label on the initial render of Quasar's component. 

Just my 2 cents. 😄 

Scott

rstoenescu:
Yeah, and I'm trying to point out that that dot was explicitly missing in earlier guidelines. Now they suddenly added it. Maybe you can understand my frustration that this cannot continue. I've changed this component a million times due to specs changing. It's enough. Need to focus on rc.1 and release final 1.0 asap.

In any case, I'll change this PR a bit and merge.

smolinari:
Ah. Ok. Then your frustration is understood. That is ridiculously silly from Google to be making those kinds of changes to the spec. It's like adding a major backwards compatibility breakage in some code. It shouldn't be done without something like a major version change. I'm going to look at where I can suggest they stop messing around like that. Maybe open their eyes to the issues such changes cause.  

Scott

smolinari:
I sent feedback on the MD site. Let's see if I get any answer. :smile:

Scott

2x2xplz:
Just wanted to say I agree with the confusion and frustration regarding the Material specs, and appreciate that ultimately you guys respect the developer first and give them the opportunity to decide. Thanks for accepting this PR and generally, thanks for your dedication to building an incredible framework. Looking forward to the future of Quasar.

rstoenescu:
@2x2xplz thx for making a PR! ultimately, it's our developer's needs that must come forward and we listen to them.

PR #4010, opened 2019-05-01
installation & usage info for outside of Vue files 💚

mesqueeb:
There was no info anywhere on this page how to install and use outside of a vue file.


rstoenescu:
Thanks!

mesqueeb:
thanks for your edit!

I wonder where the convention comes from to write `(Object)` in front of the function call.

And sometimes I'm afraid that newer coders might copy the whole line including `(Object)` to then run into syntax errors 😅

Issue #4016, opened 2019-05-02
[V1 request] Dropdown field


mesqueeb:
We currently have a dropdown _button_.
I'm in the market for a dropdown **input.**

For inspiration here's the [official Material design dropdown input](https://material-components.github.io/material-components-web-catalog/#/component/select).

mesqueeb:
My mistake, we already have it, it's called "SELECT"

https://v1.quasar-framework.org/vue-components/select#Installation

I'll update the docs

PR #4017, opened 2019-05-02
Add links between QButtonDropdown and QSelect 🔗

mesqueeb:
I just added a link to QSelect on QButton and the other way around.

### Motivation
Some people (like me) might find one component or the other with a keyword like "dropdown" and do not know where to look for the other.

#4016


rstoenescu:
Nice. Thx.

PR #4053, opened 2019-05-06
Props casting

andreiTn:
https://vuejs.org/v2/guide/components-props.html#Prop-Casing-camelCase-vs-kebab-case

> HTML attribute names are case-insensitive, so browsers will interpret any uppercase characters as lowercase. That means when you’re using in-DOM templates, camelCased prop names need to use their kebab-cased

- [x] Refactor

- [x] No


rstoenescu:
Thanks for catching this.
Was due to a copy-paste :)

andreiTn:
My pleasure. Thank you for merging.

PR #4085, opened 2019-05-10
fix(select) with fillInput render Label and not Value

H3r3zy:
Hello,

I think all is in the title ;)

- [x] Bugfix

- [x] No


rstoenescu:
Thanks for catching this and supplying a PR!

rstoenescu:
Will be in beta.24

PR #4107, opened 2019-05-13
fix(QDrawer): QDrawer with overlay and value(true) will now be opened at render

H3r3zy:
Today, I realized that our QDrawer was in mode 'Open by default' but was closed.

I then did tests and realized that it was because of the overlay mode.

So I have to look at the QDrawer code and modify it for it to work.

- [x] Bugfix

- [x] No


pdanpdan:
This has changed a lot and should now cover this case also.

PR #4135, opened 2019-05-16
fix(GoBack): Fixes the single modification

bloodf:
Fixed the single modification not working on the browser correctly with named paths.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
The GoBack directive was having some problem with the named paths and wasn't working correctly with the single modification.

### DIFF
--- a/ui/src/directives/GoBack.js
+++ b/ui/src/directives/GoBack.js
@@ -16,7 +16,7 @@ export default {
         if (ctx.single){
           vnode.context.$router.go(-1)
         } else {
-          vnode.context.$router.push(ctx.value)
+          vnode.context.$router.replace(ctx.value)
         }
       }
     }
--- a/ui/src/directives/GoBack.js
+++ b/ui/src/directives/GoBack.js
@@ -13,7 +13,11 @@ export default {
     }
     else {
       ctx.goBack = () => {
-        vnode.context.$router.go(ctx.single ? -1 : ctx.value)
+        if (ctx.single){
+          vnode.context.$router.go(-1)
+        } else {
+          vnode.context.$router.push(ctx.value)
+        }
       }
     }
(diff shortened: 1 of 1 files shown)


ui/src/directives/GoBack.js
```diff
@@ -13,7 +13,7 @@ export default {
     }
     else {
       ctx.goBack = () => {
-        vnode.context.$router.replace(ctx.value)
+        vnode.context.$router.go(ctx.single ? -1 : ctx.value)
```
rstoenescu: $router.go() receives only a Number as param, while ctx.value is a String. So this generates error if not using the `single` modifier.

rstoenescu (changes requested):

rstoenescu:
Hi @bloodf 

Are you implying `v-go-back="{ name: 'someRouteName' }"` does not works on desktop?

bloodf:
I'm developing an Cordova/web app that when used the v-go-back.single without any parameters it crashes the Vue router because on the web it passes directly the router replace, and that is not a history change, but a history state mutation. So when I did this change the directive was working fine. Will do more case checking to see, with more parameters. But was tested on all quasar modes.

ui/src/directives/GoBack.js
bloodf: made a new commit, checking for the single to do the go function, or the push function.

I've recheck the vue router docs, and if we use the replace, we will replace the history on the browser, but I don't think it's a good approach, but if you think the replace it's better, can be changed.

rstoenescu:
Thanks for your time!
Can you also check my commit above? Simplified it a little, but want to make sure it still works for you. Thanks!

PR #4189, opened 2019-05-23
Testing a promise queued approach for QImg .decode

pdanpdan:


rstoenescu:
Let's try an inlined web workers approach too.

PR #4259, opened 2019-05-30
Fixing QDialog basic confirm example

Menighin:
The confirm dialog for the basic example did not have the "Cancel" button


rstoenescu:
Thanks for catching this!

PR #4301, opened 2019-06-04
fix(Api): add more precision in the options-cover props with use-input of QSelect

H3r3zy:
We can't use options-cover and use-input simultaneously.

`cover: this.optionsCover === true && this.noOptions !== true && this.useInput !== true,`


rstoenescu:
Thanks for contributing!

PR #4302, opened 2019-06-04
feat(fullscreen) escape key exits fullscreen mode

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/mixins/fullscreen.js
+++ b/ui/src/mixins/fullscreen.js
@@ -1,4 +1,5 @@
 import History from '../history.js'
+import EscapeKey from '../utils/escape-key.js'
 
 export default {
   props: {
@@ -43,6 +44,10 @@ export default {
         return
       }
 
+      EscapeKey.register(this, () => {
+        this.exitFullscreen()
+      })
+
       this.inFullscreen = true
       this.container = this.$el.parentNode
       this.container.replaceChild(this.fullscreenFillerNode, this.$el)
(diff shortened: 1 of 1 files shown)


hawkeye64:
I am not sure if this is a good idea - I'm a bit conflicted over it. What hapens if a QDialog or QMenu is displayed in fullscreen mode. Then you have two things where 1 esc key works, but which one? It needs to be a choice or called from dev land.

lucasfernog:
The esc key that works should be the one targeting the latest shown component. It is done [here](https://github.com/quasarframework/quasar/blob/fc79d6bffe602c496be5d84093948197a9531631/ui/src/utils/escape-key.js#L13) so i believe everything should work

lucasfernog:
I believe this EscapeKey and Fullscreen mixin could be documented if its not... if a user wants to handle esc key properly to dismiss something, he can use the EscapeKey helper and if wants a fullscreen component, could use the Fullscreen mixin (however i don't know if Quasar wants to expose those two)

lucasfernog:
@hawkeye64 I misread your comment.. that is really a valid point. I think if it is open directly in fullscreen mode, ESC should close it. If if is open normally and then gets maximized, then it should exit fullscreen mode on ESC key. But now I get the point and I understand if this gets closed. However, exiting fullscreen mode on ESC key is really an expected behavior everywhere.

nothingismagick:
Escape inception?

rstoenescu:

PR #4377, opened 2019-06-14
Enable publishing with electron-builder via CLI

2x2xplz:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Quasar does a great job integrating with electron-builder to create executables. However there currently is no ability to use electron-builder's `publish` features. Further, because electron-builder looks in the project's `package.json` for configurations, not `quasar.conf.js` , it is almost impossible to even manually publish without overwriting files that will interfere with future quasar builds.

This PR adds a new `--publish` flag (or `-P`) to the CLI that gets passed through to electron-builder. It is ignored under any other build mode, or when using electron-packager. The user remains responsible for correctly preparing his configuration, and using a `--publish` value that corresponds to a valid option (https://www.electron.build/configuration/publish)


app/lib/quasar-config.js
```diff
@@ -749,6 +749,7 @@ class QuasarConfig {
             arch: cfg.ctx.archName,
             config: cfg.electron.builder
           }
+          if (cfg.ctx.publish != null && cfg.ctx.publish.length > 0) cfg.electron.builder.publish = cfg.ctx.publish
```
nothingismagick: !==

app/lib/electron/index.js
```diff
@@ -110,6 +110,11 @@ class ElectronRunner {
         log(`Bundling app with electron-${bundlerName}...`)
         log()
 
+        if (cfg.ctx.debug) {
+          console.log(`DEBUG MODE - bundlerConfig:`)
+          console.log(bundlerConfig)
```
nothingismagick: use log not console.log please

nothingismagick (changes requested):
This is interesting. Have you tested on Mac, Windows and Linux?
I marked a couple things that would be good to fix and we'll need full documentation at quasar.dev when this feature lands. Your comments in the commits looks like a good start though. 

Awesome initiative!

app/lib/electron/index.js
2x2xplz: I started with just plain `log` (yes, that is a Ren & Stimpy reference) but that outputs:
```
app:electron DEBUG MODE - bundlerConfig: +2ms
app:electron [object Object] +2ms
```

while `console.log` actually displays the JS object's contents. I also don't think the `app:electron` prefix is accurate, because the display is triggered from the `debug` flag, not something in electron or any of its builders.

nothingismagick: well, personally I think `log` is better than bad, its good - so why not just destructure your object with JSON.parse?

nothingismagick: Or just use console.log for the bundlerConfig and log for line 113. This seems like it could be valid, and it won't break the universe. Of course, its just my opinion. Razvan will have to sign off on it anyway.

2x2xplz:
Have not tested on Mac or Linux. It's a CLI option, not really sure how it could fail on a specific platform. The only thing this PR does is pass the `--publish` flag and the user-supplied value, as-is, to electron-builder. We don't do any checks or verification within quasar. If the build fails, it's due to the user not properly configuring publishing in electron-builder, and they can ask questions or raise issues there. If we tried doing verification on the quasar side then we would have a problem in the future when electron-builder adds a new option.

This PR is highly necessary because currently, doing a manual publish step requires devs to manually update their quasar project's `package.json` file, which can cause all kinds of problems. Technically you could run `.\node_modules\.bin\electron-builder --config .\dist\electron\Packaged\builder-effective-config.yaml  --publish always` but that seems less likely to be cross-platform than this PR, and way more error-prone. It also repeats the build step which wastes time.

2x2xplz:
As far as documentation, I already updated the docs within `quasar build --help` which is the source of documentation at https://quasar.dev/quasar-cli/cli-documentation/commands-list

I will be happy to make a few edits to https://quasar.dev/quasar-cli/developing-electron-apps/build-commands as well

nothingismagick:
https://github.com/quasarframework/quasar/edit/dev/docs/src/pages/quasar-cli/cli-documentation/commands-list.md

This file is the source for that, it is not generated from the CLI.

And it would be imperative to add to the build-commands as well.

Thanks!!!

2x2xplz:
Docs are updated. Thanks!

nothingismagick:
Thanks for seeing this through!

rstoenescu:
Thanks for contributing!

2x2xplz:
Thanks a lot @rstoenescu and @nothingismagick !

PR #4435, opened 2019-06-21
Add Example - debounce / throttle on methods.

webnoob:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

**Other information:**

### DIFF
--- a/docs/src/pages/quasar-utils/other-utils.md
+++ b/docs/src/pages/quasar-utils/other-utils.md
@@ -56,6 +56,10 @@ created () {
 }
 ```
 
+::: warning
+Debouncing your functions using a method declaration like `myMethod: debounce(function () { // Code }, 500)` will mean that the debounced method will be shared between *all* rendered instances of this component, so debouncing is also shared. This should be avoided by following the code snippet above.
+:::
+
 There's also a `frameDebounce` available which delays calling your function until next browser frame is scheduled to run (read about `requestAnimationFrame`).
 
 ``` js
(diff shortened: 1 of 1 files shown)


rstoenescu:
This throttled/debounces for ALL instances of that respective component.
Should use a beforeMount hook to initialize those methods instead.

webnoob:
Like this?

```js
methods: {
  myfunction () {
    // Run some code.
  }
},

bedforeMount () {
  this.myFunction = debounce(this.myFunction, 500)
}
```

Also, could you explain:

> This throttled/debounces for ALL instances of that respective component.

Do you mean the timeout will be shared among all rendered versions of that component so if you have 6 on the page, you can only run the function once between the all the components every XXX ms?

rstoenescu:
@webnoob 

Why also defining it in methods? Just this will suffice:

```
beforeMount () {
  this.myFunction = debounce(this.myFunction, 500)
}
```

The Vue component definition is exactly what gets instantianted. So if you define a method with the result of the debounce function (methods: { doSmthng: debounce(...) }) you're essentially telling the instantiation Object to have that method pointing to SAME function. So it gets shared by all rendered instances.
When you instantiate a component, the instantiating object is already created (it doesn't gets evaluated again). Does this makes sense to you?

webnoob:
Your explanation makes perfect sense, thanks.

I also put it in methods as I personally find it tidier - the method is still defined in methods (where you would expect it) but if you want it debounced, you just add a line in `beforeMount`.

If I understand you correctly if I didn't put it in methods, I'd need:

```js
beforeMount () {
  this.myFunction = debounce(function () {
    // Do my code here.
  }, 500)
}
```

Which way would you like it documented? Given the confusion I've clearly had with this, I think it's imperative we at least document it somehow.

rstoenescu:
Bulletproof approach:

```
methods: {
  myMethod () { .... }
},

created () {
  this.myMethod = debounce(this.myMethod, 500)
}
```

I just realized there are devs that might not know how to use arrow functions yet.. long story short, document like above.

webnoob:
Ok, will sort it now.

webnoob:
I've also added a warning in there because examples I've seen online demonstrate doing it the way I did initially but without the warning you gave me.

p.s Perhaps we should be educating people about useful things like arrow functions etc in a tips doc.

rstoenescu:
Thanks for your time on this one!

PR #4443, opened 2019-06-23
QTable & QTree define slote-scope props enumerable and configurable.

AjiTae:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

Just make props defined with Object.definePropety enumerable and configurable. 
Main use-case: to pass slots trough wrapper component (v-bind does not bind non-enumerable props), but it can be useful in some other cases. 

- [ ] Yes
- [x] No


rstoenescu:
Thanks for contributing!
Careful with the `configurable` though!

This will go into final 1.0.0

Issue #3945, opened 2019-04-22
[v1][Feature Request] Mini QDrawer open as overlay


smakinson:
I was experimenting with an idea based on the example in the docs where the left drawer goes from mini to not mini on hover. I was able to have it change to overlay also on hover, but when I use this method to accomplish it, that min mode width ( default 60px ) is lost when it switches to overlay mode. I'd like to be able to keep that spacing to prevent movement of the main content while also having it "open" ( from mini to not mini ) as overlay, also to prevent movement of the main content. I may be going about this in the wrong way ( also I'm still on 0.17 ) so I thought I'd start a discussion about it.

Codepen showing what happens now:
https://codepen.io/smakinson/pen/VNBxXy?editors=1010

A side note to this ( and maybe another discussion ) is I have also been experimenting with multiple drawers on the right side and trying to have one overlay the other that would be a mini to not mini expanded. One would be a navigation type drawer and the other potentially a chat area. Anyway, not to talk too much on that one other than to see if I should open a discussion or if the idea of multiple drawers on one side is an unexpected move.

devel-pa:
Hi,
My fix for this on `style` area:
```
.q-drawer-container aside {
  transform: translate3d(0px, 0px, 0px) !important;
}
```

smakinson:
@devel-pa I added the above css to the pen & I am seeing the same issue.

devel-pa:
@smakinson Sorry, an incomplete `style`, my bad. Tested that adding into your pen. The commented second one is for right drawer.

```
.q-page-container {
  padding-left: 57px;
  /*padding-right: 57px;*/
}
```

rstoenescu:
Fixed through #4500

PR #4500, opened 2019-06-28
feat(mini-to-overlay): allow mini drawer to "open" as overlay

smakinson:
A new property mini-to-overlay is added to allow letting the QDrawer be mini and move to full size without moving the page content, as an overlay. The drawer can still be hidden via v-model.

The reason behind this feature is I was trying to toggle overlay in combination with turning mini off for the QDrawer and it lead to trying to override css with !important and I thought it might be something others would like as well.

This adds a new property `mini-to-overlay` to QDrawer to allow it to move from mini=true to mini=false without moving the page content and overlay the page. The drawer can still be hidden via v-model and I didn't touch the behavior or breakpoint props. It should not be used in combination with overlay.

I also added new toggles to test it in the dev layout page and fixed the selects for the behavior. Set Left Mini to Overlay on and then toggle the mini property.

What do you think? Did I miss anything? Do I need to update docs somewhere?

### DIFF
--- a/ui/dev/components/layout/layout.vue
+++ b/ui/dev/components/layout/layout.vue
@@ -342,7 +342,7 @@
             <q-toggle dense v-model="leftMiniToOverlay" label="Left Mini to Overlay" />
           </div>
           <div>
-            <q-select v-model="leftBehavior" :options="drawerBehaviorOptions" />
+            <q-select emit-value v-model="leftBehavior" :options="drawerBehaviorOptions" />
           </div>
           <div>
             <q-input type="number" align="right" prefix="Bkpt" placeholder="Bkpt" v-model="leftBreakpoint" />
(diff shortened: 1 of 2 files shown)


smakinson:
Relates to this feature request: https://github.com/quasarframework/quasar/issues/3945

rstoenescu:
Excellent idea! Thank you for your time.
I'll try my best to add this to the final 1.0.0 build. There's a few minor bugs that need to be squashed.

rstoenescu:
Fixed:
* Setting v-model to "false" does not free up layout space
* Force drawer into fixed position otherwise it breaks header/footer if view is along the lines of `hHh lpr fFf`.

Thank you for contributing!

miladmeidanshahi:
Nice feature, @rstoenescu When do you want to release?

miladmeidanshahi:
The overlay prop on drawer doesn't work, Is this PR fix this issue?

rstoenescu:
@miladmeidanshahi `overlay` prop is one thing, and `mini-to-overlay` another

PR #4528, opened 2019-07-02
Marketing

nothingismagick:
Landing
- Add several more claims
- Update claims for readability
- Add email icon as link to /newsletter

Footer (in Docs)
- Add email icon as link to /newsletter

/newsletter
- Add sendinblue iframe
- New copy
- New left sidebar menu listing

Note:
retain function by allan for future implementation with our own proxy server instead of the lame iframe


lucasfernog:
I had to use v-if in order to remove the right drawer without page floating when changing routes (maybe there is a way to remove the QDrawer animation?) and for it work properly when loading the newsletter page directly...

PR #4539, opened 2019-07-04
cordova: update commands showing extra params

mpacary:
Looks like the sample commands are missing something (see diff).

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: doc update suggestion

- [ ] Yes
- [x] No


rstoenescu:
Hi,

Thanks for making a PR.
But in this case, it's correct. "quasar dev|build -m android|ios" is an alias to the longer form. Both forms work.

mpacary:
Thanks for your quick reply & sorry for not knowing the short form.

rstoenescu:
It was added recently, so no worries. What I appreciate is your willingness to improve docs, which deserves respect.

PR #4566, opened 2019-07-06
feat(proton) watch for rust files on dev

lucasfernog:
I can't restart the application properly (only after the new binary is compiled) because to do that I would need to: 
1- Replace `cargo run` with `cargo build` followed by running the binary myself (and that includes reading the binary name), or 
2- Read the `cargo run` output and wait for the message "Finished compiling", both not good solutions. 

So right now it closes the application, then compiles and runs it again.

### DIFF
--- a/app/lib/proton/index.js
+++ b/app/lib/proton/index.js
@@ -1,15 +1,15 @@
 const
   fs = require('fs'),
   fse = require('fs-extra'),
-  readline = require('readline')
+  readline = require('readline'),
+  chokidar = require('chokidar'),
+  debounce = require('lodash.debounce')
 
 const
   log = require('../helpers/logger')('app:proton'),
   { spawn } = require('../helpers/spawn'),
   onShutdown = require('../helpers/on-shutdown'),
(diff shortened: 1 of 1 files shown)


nothingismagick:
This is sort of how quasar manages electron too.

>

app/lib/proton/index.js
```diff
@@ -36,10 +38,33 @@ class protonRunner {
...
-    )
+    const startDevProton = () => {
+      return this.__runCargoCommand(
+        ['run', '--features', 'dev'],
+        args
+      )
+    }
+
+    // Start watching for proton app changes
+    chokidar
```
rstoenescu: Should stop the watcher on protonRunner's stop() method being called.

rstoenescu (changes requested):

app/lib/proton/index.js
lucasfernog: Done here #e85aaba52bc5cff3f12903374faacefa916f290c

PR #4597, opened 2019-07-09
update title for better contrast with first section

mesqueeb:
currently it just says "best practices" and it's unclear that this represents the part "upgrading from 0.x to v1".

As opposed to the first section which says: "Upgrading from older v1 to latest v1"


rstoenescu:
Thanks!

PR #4629, opened 2019-07-11
fix:4616 - Add computed property support to API Doc and TS Types

### LINKED ISSUE #4616: Missing and Wrong typedef in validation of form components
**Describe the bug**
The  validate.json file for the validate mixin is assuming the prop error as the test result. Following the code in the component, the prop error and errorMessage should be used for external validation and the computed hasError returning the validation result, both external and internal. The code in the js file it's ok, but in the json that generates the typedef files and docs it's wrong.

Also in the first examples of QInput usage in the documentation, it shows the use of hasError as the only way to validate the fields without QForm. So probably this computed was forgotten when QForm was created.

**Expected behavior**
The computed hasError should be in the json as the return of validation tests. And error and errorMessage as the prop used to input the external validation data.

outofmemoryagain:
Add support for Computed properties in the API documentation and generated typescript types.  This resolve #4616 by adding hasError to the validation mixin API JSON file and adding support in the doc ApiRows to render computed properties.  Also updated the Typescript generation script to generate a readonly property for computed properties.

- [X] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

### DIFF
--- a/docs/src/components/ApiRows.js
+++ b/docs/src/components/ApiRows.js
@@ -233,6 +233,18 @@ export default {
       return child
     },
 
+    computed (h, computed) {
+      const child = []
+
+      for (let propName in computed) {
+        child.push(
+          this.getProp(h, computed[propName], propName)
+        )
+      }
+
+      return child
+    },
+
     slots (h, slots) {
       const child = []
 
--- a/docs/src/components/DocApi.vue
+++ b/docs/src/components/DocApi.vue
@@ -119,7 +119,8 @@ export default {
         props: null
       },
       filter: '',
-      filteredApi: {}
+      filteredApi: {},
+      api: {}
     }
   },
 
--- a/ui/build/build.api.js
+++ b/ui/build/build.api.js
@@ -35,9 +35,9 @@ function getMixedInAPI (api, mainFile) {
 }
 
 const topSections = {
-  plugin: [ 'injection', 'quasarConfOptions', 'props', 'methods' ],
-  component: [ 'behavior', 'props', 'slots', 'scopedSlots', 'events', 'methods' ],
-  directive: [ 'value', 'arg', 'modifiers' ]
+  plugin: [ 'injection', 'quasarConfOptions', 'props', 'methods', 'computed' ],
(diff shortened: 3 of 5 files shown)


outofmemoryagain:
Not really sure why this one was close, I think I was cleaning up branches and deleted this one, and it was closed as a result.

rstoenescu:
Hi,

This should be done some other way. We can't just start to put computed props into the API card. If you put one, then it's gonna be inconsistent not to put all of them (for all components). But 99.999% of them are "private". If this is an exception (and it indeed seems to be), then we should make an "exception" logic in build.types.js

outofmemoryagain:
I can look into doing it in the way that you suggest as a hard-coded type definition.  But, I disagree that it is inconsistent to document a public computed property while not document all of them if the rest are indeed private.  Only public properties should be documented.  The fact that this property is a computed property from a consumption standpoint is an implementation detail and we could also consider treating it as a read-only property that we document as a property using the existing approach in the JSON files instead of adding a new category for computed properties in the json.

rstoenescu:
Yes, but as long as someone sees "Computed Property" section for this component, it will make him think that all such computed properties are described... and that "look, there are other computed properties, why are these not documented?"

So what I'm saying is that we need a new (ingenious) way to deal with this situation. It's the only computed property that has any meaning to devland (from all computed properties of all components)... and this one was not meant to be "public"... it was discovered by a Quasar user (by looking at Quasar source) and it made sense for him to use it... but I did NOT advertise for it and nor should it be "public information"...

outofmemoryagain:
I completely understand, and I wondered myself if this was the right approach, but I decided this would be a good start to the discussion.  

So the following questions should be clarified: 
1. Do we want `hasError` to be documented?
2. Do we want `hasError` to be publicly typed?

If both yes, I think I can do it with just adding it as a property to the JSON documentation.  If 1. no and 2. yes, I will have to either extend the json or hardcode the api.

PR #4644, opened 2019-07-12
fix(docs): spelling

hawkeye64:


wtoalabi:
Good catch!
The language in use (context) is  American English however.
Which means, words like favourite (is favorite), colour (is color), behaviour (is behavior), etc
There is an exhaustive list here: http://www.tysto.com/uk-us-spelling-list.html

nothingismagick:
@wtoalabi - why is US English the default? Hey

nothingismagick:
Cuz if it is then `grey !== gray`

PR #4651, opened 2019-07-14
feat(tauri) read templates from @quasar/tauri package

lucasfernog:


nothingismagick:
The templates are located at @quasar/proton/templates

Justification for this is as follows:
- It allows us to iterate quickly
- It keeps core "pure" from non-node components
- Enables easy integration of other bindings

i.e. A separation of concerns.

Please see: https://github.com/quasarframework/proton/tree/dev/templates

rstoenescu:
The reason I asked for separate proton repo is to use it for forking cargo packages.
The template that gets into the user space for proton should stay in the CLI (I'll be regularly checking PRs to it and merging immediately during my vacation). 
We shouldn't create a (possibly temporary) @quasar/proton package to wrap cargo packages. Instead, use the "git" approach for these packages within the template from CLI.

Issue #4603, opened 2019-07-10
Position fixed bug on ios


Evertvdw:
**Describe the bug**
I had an issue with a fixed header which became invisible after I used a fullscreen view of QCarousel. This happened only on IOS with safari (IOS 12.3.1). 

After some googling I found this [this](https://stackoverflow.com/questions/19325052/disappearing-position-fixed-header-in-ios7-mobile-safari) thread on stackoverflow.

I tried the suggested fix and that worked. Can this maybe included somehow in the quasar package itself? Because I can imagine this can be quite frustrating for others if this happens.

**Codepen/jsFiddle/Codesandbox (required)**
I was not able to reproduce the issue in a codepen with safari IOS. Stackoverflow suggests that it might be related to 'a lot going on' on a page, which might be the reason why I had the issue in my project and not in an isolated example on codepen.

**Fix**
The fix includes two lines of css: 
```
  transform: translateZ(0)
  -webkit-transform: translateZ(0)
```
which could be added to the header (and other components?) using fixed positioning.

**Quasar version**
Using Quasar v1.0.3

rstoenescu:
This should no longer happen in v1.1.0.

PR #4660, opened 2019-07-15
fix(quasar): Try to overcome iOS bug with fixed position

pdanpdan:
ref #4603


rstoenescu:
A few problems:

1. Should be added to critical.styl, otherwise it will override custom css classes in components.
2. Doesn't actually solves iOS's problems :(

PR #4662, opened 2019-07-15
problem with double usage of single line `

mesqueeb:
Not a complete fix, but I can't come on a better solution.
There's double usage of ` on the
second line of Misc: https://quasar.dev/start/upgrade-guide#Misc

image


rstoenescu:
Thanks for catching this!

PR #4666, opened 2019-07-15
Improve date checking

daniel-rau:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Regex has been tested on https://regex101.com/**
**Regex has been tested on https://regex101.com/**

**Comments**

Right now your regex expression for date does not actually check for valid dates.  It only checks for valid month, day and year options in a general sense without considering how many days can be in each particular month.  It also does not consider how to handle leap years.

For example:
"2019/02/31" is not a real date, but the old regex would accept this value.

The code revision above should fix these issues.  It is a modified version of a regex expression that I found on StackOverflow (https://stackoverflow.com/questions/15491894/regex-to-validate-date-format-dd-mm-yyyy). 

Note: I used Jason Greanya's regex answer farther down the page (currently has 9 upvotes) as he has a version that is in the correct format (yyyy/MM/dd).  The only changes that I made to his regex is that I required there to be a "/" between year, month and day. (Example: 2019/05/16 = valid; 20190516 = invalid; 2019-05-16 = invalid).

Hope this helps!  As a developer, I really appreciate all that you are doing with quasar.  Congrats on 1.0!


daniel-rau:
I took a look at the old regex statement this morning and noticed that you currently have a "-?" at the beginning of the old statement.  I don't know why you would expect a potential "-" at the beginning of a date, but in case that is important for some other functionality that I am unaware of, feel free to add it to my regex statement if needed.

nanov:
@daniel-rau I guess it is in order to support BC dates. ( ie. year 1000 before Christ, or  in other words -1000 ).

daniel-rau:
@nanov If that is the case (and we are wanting to support dates prior to 1000AD) then my Regex will need to be updated to one of the options as follows...

If we want a date like February 26th 22BC to look like "-22/02/26"... 
Then the Regex should be:
`-?([0-9]?[0-9]?[0-9]?[0-9][\/]((0[13-9]|1[012])[\/](0[1-9]|[12][0-9]|30)|(0[13578]|1[02])[\/]31|02[\/](0[1-9]|1[0-9]|2[0-8]))|([0-9]{2}(([2468][048]|[02468][48])|[13579][26])|([13579][26]|[02468][048]|0[0-9]|1[0-6])00)[\/]02[\/]29)`

- Note: The option above is closer to resembling what the original regex was like

If we want that same date to instead look like "-0022/02/26" with leading zeros...
Then the Regex statement should be:
`-?([0-9]{4}[\/]((0[13-9]|1[012])[\/](0[1-9]|[12][0-9]|30)|(0[13578]|1[02])[\/]31|02[\/](0[1-9]|1[0-9]|2[0-8]))|([0-9]{2}(([2468][048]|[02468][48])|[13579][26])|([13579][26]|[02468][048]|0[0-9]|1[0-6])00)[\/]02[\/]29)`

Jason Greanya's warnings from his stack overflow answer that I used ([here](https://stackoverflow.com/questions/15491894/regex-to-validate-date-format-dd-mm-yyyy)) should also be noted if we are going to support older calendars...

"Other than order, these all are accurate to the Julian Calendar (leap year every four years) until 1700, when the Gregorian Calendar diverges from the Julian. It has two issues:

1. It accepts the year 0000, which doesn't exist in many, but not all, standards. Note that ISO 8601 does accept year 0000 (equivalent to 1 BCE).

2. It doesn't skip the 10-13 days which were lost when the Gregorian Calendar came into use. This varies by locality though. For example, the Roman Catholic Church skipped 10 days, October 5th through October 14th, 1582, but Greece (the last to switch) skipped February 16th through the 28th of 1923, 13 days, having to take into account the leap years of 1700, 1800, and 1900.

This has been tested against Java's calendar implementation from the year 0001 until the year 9999 with the only discrepancy being the abovementioned 10 days in 1582."

Having said all that, at this point we are probably going far more in depth with q-date than will ever be actually be required by the large majority of users and developers (as you have to click backwards 50 times before you can reach such low years in q-date).  If there are developers who need to worry about the locality of the extra days missed in 1582 or how leap-year was different in ancient calendars, they will most likely want to write their own regex rule for checking those dates anyways.  At this point we should probably just pick one of the two Regex statements that I posted above and call it good for now.  It is still far better than what it used to be.

daniel-rau:
Regex statement has been updated following conversation above.  This should be ready to be accepted in quasar (pending review of course).

pdanpdan:
You cannot create a regex that is correct no matter what year. There are a lot of exceptions for long ago years.

ui/src/utils/patterns.js
```diff
@@ -8,7 +8,7 @@ const
   rgba = /^rgba\(((0|[1-9][\d]?|1[\d]{0,2}|2[\d]?|2[0-4][\d]|25[0-5]),){2}(0|[1-9][\d]?|1[\d]{0,2}|2[\d]?|2[0-4][\d]|25[0-5]),(0|0\.[0-9]+[1-9]|0\.[1-9]+|1)\)$/
 
 export const testPattern = {
-  date: v => /^([0-9]{4}[\/]((0[13-9]|1[012])[\/](0[1-9]|[12][0-9]|30)|(0[13578]|1[02])[\/]31|02[\/](0[1-9]|1[0-9]|2[0-8]))|([0-9]{2}(([2468][048]|[02468][48])|[13579][26])|([13579][26]|[02468][048]|0[0-9]|1[0-6])00)[\/]02[\/]29)$/.test(v),
+  date: v => /^-?([0-9]+[\/]((0[13-9]|1[012])[\/](0[1-9]|[12][0-9]|30)|(0[13578]|1[02])[\/]31|02[\/](0[1-9]|1[0-9]|2[0-8]))|([0-9]{2}(([2468][048]|[02468][48])|[13579][26])|([13579][26]|[02468][048]|0[0-9]|1[0-6])00)[\/]02[\/]29)$/.test(v),
```
nothingismagick: `-0/01/01` works according to your regex, as does `-9996/02/29` and the absolutely insane:

`-2000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000000000000000000000000/01/01`

nothingismagick: tested at https://regex101.com

nothingismagick (changes requested):
The 29th of February was not introduced until the middle ages.
https://en.wikipedia.org/wiki/February_29

There is no -0 year.

daniel-rau:
@nothingismagick @pdanpdan You both bring up valid points that we discussed in part before the most recent commit above.  While you are right that there is no regex statement that will perfectly be able to replace this line of code for all calendars, this regex statement is still a great improvement of what it used to be (see original commit).

The original regex statement already has all of the issues you mentioned above and allows for strange dates such as 2019/19/19 and 2019/08/38.  The regex that I have provided should be accurate until around the 1580's and then will have the occasional issue with things like leap year before that.

At the point where developers are wanting to handle ancient calendars, they should be creating their own rules for q-date as you start dealing with crazy things such as calendar locality.  

I know this isn't perfect and I am open to suggestions but this is a vast improvement to what currently stands in the code.

daniel-rau:
The origional regex statement: 
`-?[\d]+\/[0-1]\d\/[0-3]\d`

ui/src/utils/patterns.js
daniel-rau: @nothingismagick I am happy to change the regex back to only allowing a year value with a max of 4 digits if that is how we want to proceed.  The original regex statement allowed crazy long years however, which I why I currently have included that in my regex.

daniel-rau: The origional regex statement:
`-?[\d]+\/[0-1]\d\/[0-3]\d`

pdanpdan:
Yes, and that allows any valid date, and just a few invalid ones, and it's short and quick. The date can be further validated and rejected if invalid. In your case some valid dates will not pass validation and some invalid ones will pass, and the good ones cannot be accepted. It's specific vs sensitive.

daniel-rau:
The original regex statement accepts a lot of invalid date possibilities every single month.  
For example: 
- When you enter a month that has a value > 12
- When you enter a day that is between 32-39
- When you enter a day with 29, 30 or 31 that is not actually valid for that month. (like 2019/09/31)

That is over 100 possible invalid dates that could be entered for any given year that the old regex would not catch (and that is just including the second and third points I made).  I think if you are trying to have a default rule for date's then the old regex honestly becomes pretty useless.

My regex would catch all of these and only starts to have issues when you start dealing with ancient calendars prior to the 1580's.  For the purposes of having a default rule that developers can choose to use on q-date input fields (which is where this is mainly used) this should be more than a sufficient regex to use.

Again, if you have suggestions as to how I could improve this statement, please let me know and I can work on revising it.

daniel-rau:
@pdanpdan @nothingismagick 
I want to apologize for being so wordy and defensive in my comments above.  I have really appreciated all that you and everyone else has done to make such a fantastic framework that I have truly enjoyed using.

I know that it is impossible to have a perfect regex statement to match every calendar throughout history, but I am hoping to at least add an improvement to this date checking.  What specifically would you like me to fix so that we can move forward with this?

pdanpdan:
This is a can of worms, sorry. We only need a quick regex validation to exclude some common mistakes.
A real validation is almost always needed.

PR #4669, opened 2019-07-16
fix(docs): wrong formatting for table

mesqueeb:
Was something wrong with the table:

image


rstoenescu:
Thanks for catching this!

PR #4045, opened 2019-05-04
feat(quasar): add docsUrl to API definition

lucasfernog:
**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [x] It's submitted to the `dev` branch and _not_ the `master` branch
- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)
- [ ] It's been tested on a Cordova (iOS, Android) app
- [ ] It's been tested on a Electron app
- [x] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar-framework.org/tree/dev/source) (for faster update click on "Suggest an edit on GitHub" at bottom of page) or explained in the PR's description.

If adding a **new feature**, the PR's description includes:
- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

The ```quasar docs``` command receives a Quasar component/directive/plugin and opens the docs page. It also supports app extensions. It also has a --api option, which uses the API table anchor as URL.

Credits to @hawkeye64 for the idea.

EDIT: added options to open sections (components, plugins, directives and extensions), e.g. ```quasar docs components```

PR #4749, opened 2019-07-27
feat(IDE) generate web-types (helper file for WebStorm)

panstromek:
Initial work

This PR should improve completion and Quasar support in JetBrains editors. Webstorm already uses those files but JetBrains generates them themselves and publish to @web-types scope. They support linking custom web-types linked through package.json, though.

It would be more appropriate if we generated these files during build so that we don't have to wait until WebStorm team catches up after every release. We can also be responsible for its content and make sure it's correct and complete. We already generate Vetur helpers, that's a similar story.

More information:
https://github.com/JetBrains/web-types/
https://youtrack.jetbrains.com/issue/WEB-31682#focus=streamItem-27-3538850.0-0

### DIFF
--- a/ui/build/build.utils.js
+++ b/ui/build/build.utils.js
@@ -2,7 +2,8 @@ const
   fs = require('fs'),
   path = require('path'),
   zlib = require('zlib'),
-  { green, blue, red, cyan } = require('chalk')
+  { green, blue, red, cyan } = require('chalk'),
+  kebabRegex = /[A-Z\u00C0-\u00D6\u00D8-\u00DE]/g
 
 function getSize (code) {
   return (code.length / 1024).toFixed(2) + 'kb'
@@ -63,4 +64,11 @@ module.exports.rollupQuasarUMD = function (config = {}) {
       }
     }
   }
+}
+
+module.exports.kebabCase = function (str) {
+  return str.replace(
+    kebabRegex,
+    match => '-' + match.toLowerCase()
+  ).substring(1)
 }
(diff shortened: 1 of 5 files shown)


nothingismagick:
AWESOME

panstromek:
There are problems to solve though:
 - [ ] Links to docs don't have any canonical structure so I can't generate them correctly everytime (this is similar for Typescript definitions, where links to docs could also be nice to have)

     We could either add the link into JSON api or create a canonical link structure (which could just redirect links to corrent pages, so that we don't break existing links on the web).. not sure what's the best option here. Or I can just find a clever way how to pull out links from the docs folder.. seems kinda hacky, though

- [ ] Same problem is with source file links - At least in their case I can look them up in a filesystem during build, though.

- [ ] Problems I asked Piotr from JetBrains about, I will update the PR based on his response

lucasfernog:
Doc links on our API was initially done here: #4045

panstromek:
This is ready to merge as initial work. 

Some types are a bit lacking but I will need to share code with TypeScript generator to implement them. Since there are already few more pending TypeScript PRs and I don't want to introduce too many conflicts, I will do it after it all get merged.

Other thing missing are docs links - but this question is still unresolved -- see above.

We also need full component/directive descriptions in JSON API - if we then put it in typescript and web-types, WebStorm can display it directly in the IDE which is very handy. In extreme case, we could even show a big part of documentation just on `ctrl+q` - that would speed up at least my workflow quite a bit 😉

rstoenescu:
Excellent work!!! @panstromek

Issue #4497, opened 2019-06-28
prevent scroll will not work on Iphone


jehadja:
prevent scroll class which used on open dialog for example have overflow:hidden ; 
which is quit correct for browser and windows .
unfortunately its will not work for iPhone 9 + 
its a famous bug 
reference
https://stackoverflow.com/questions/3047337/does-overflowhidden-applied-to-body-work-on-iphone-safari
```
.q-body--fullscreen-mixin, .q-body--prevent-scroll {
overflow: hidden !important;
}
```
i just add fixed for layout on open dialog etc... 
and it works perfectly  
```
.q-body--prevent-scroll
  .q-layout
    position: fixed

```
i actually need it for putting carousel inside dialog which was not working correctly because scroll can go x,y 
![image](https://user-images.githubusercontent.com/17106547/60370525-83e45400-99ff-11e9-958f-98cbd3754add.png)
now every thing work correct hope it will help !

rstoenescu:
Fix will be available in 1.0.6

Issue #4127, opened 2019-05-15
[v1] Scroll issues with containerized Layout inside Dialog on iOS


tegola:
I'm using a `<q-layout container>` inside a `<q-dialog>` on a Cordova iOS app, and I noticed a couple bad scrolling issues.

After investigating a little bit I noticed that even the demo on the docs has the same issues: https://v1.quasar-framework.org/vue-components/dialog#Example--With-containerized-QLayout

Here's a screen cap: https://imgur.com/a/TAr14xf

As you can see, the toolbar is cut off and, when at start/end of the scroll view, the view behind is scrolled instead.

On a desktop browser and on Chrome on Android there are no such issues.

fergusmeiklejohn:
I also see this issue

lxmfly123:
The background content should not scroll when a dialog popped.

lxmfly123:
The solution below fix the problem on iOS Safari.

https://github.com/willmcpo/body-scroll-lock

tegola:
Thanks, I'll try it ASAP, I'm actually busy with another section of my app.

@rstoenescu are you aware of this? Is it something you want to include in quasar?

rstoenescu:
Fix will be available in 1.0.6

rstoenescu:
Sorry, one more problem needs fixing. Reopening.

summerinsect:
I got this issue on Android

rooneyl:
Wondering if we have a workaround for this until it gets fixed?

ibrainventures:
to get / work there some progress .. only for the "records":

flickering: (all from the quasar doc pages):

  containerized qlayout :  .> inside the lorem ipsum content area
  not flickering : inside the qdrawer element

not flickering also on basic scroll and fixed size dialogs

tested on ios 12.1x and 12.2  / ipads and iphones

9527Lin:
## An easy solution to the jitter problem at the head and tail of the layout layout in the dialog:

* 1: Register the following code as a global directive

* 2: Add a **div** under the **q-page** element and add the **v-fixLayoutJitter** 

### example

[https://youtu.be/mxFkz0KqX4s](url)

### html template：
```html
    <q-layout view="hHh Lpr fFf" container>
        <q-header>...header content</q-header>
        <q-footer>...footer content </q-footer>
        <q-page-container>
            <q-page>
                <div v-fixLayoutJitter="{bottom:64}">
                  ...page content
                </div>
         </q-page>
        </q-page-container>
[...]

PR #4786, opened 2019-08-01
feat(quasar): Prevent scroll on all platforms (including iOS overflow)

pdanpdan:
ref #4497, ref #4127


rstoenescu:
Reference for later on the problem when fixed positioning the doc: https://stackoverflow.com/questions/37586286/prevent-overflow-rubberband-scrolling-on-ios

PR #4798, opened 2019-08-02
feat(QItem): Adds transition-show & transition-hide to QItem

nachodd:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

The main idea of this change is to add the possibility of using transition-show and transition-hide to QItem, so when you show or hide the item, it animates. It has to be used with v-show or v-if.
This PR doesn't affect the documentation. If it is merged, I will make the necessary changes to it.

### DIFF
--- a/ui/src/components/list/QItem.js
+++ b/ui/src/components/list/QItem.js
@@ -3,11 +3,12 @@ import Vue from 'vue'
 import { RouterLinkMixin } from '../../mixins/router-link.js'
 import slot from '../../utils/slot.js'
 import { stopAndPrevent } from '../../utils/event.js'
+import TransitionMixin from '../../mixins/transition.js'
 
 export default Vue.extend({
   name: 'QItem',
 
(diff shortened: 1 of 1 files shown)


rstoenescu:
Hi,

Thank you for wanting to contribute!
This may seem like a good idea, but it is detrimental to the component Vue render speed.

If you need this then a wrapper can be created.

nachodd:
> Hi,
> 
> Thank you for wanting to contribute!
> This may seem like a good idea, but it is detrimental to the component Vue render speed.
> 
> If you need this then a wrapper can be created.

After checking on the transition mixin in the source code, I was thinking... 
What do you think on creating a transition-group mixin and applying it on QList, in order to use transition-show and transition-hide directives on it?
If you think it's a good idea, maybe I can help you with that.

Issue #4874, opened 2019-08-12
QInput with mask, fill-mask and clearable always show clearable icon


hawkeye64:
**Describe the bug**
When QInput is used with properties clearable, mask, and fill-mask the "clearable icon" is always displayed. 

**Codepen/jsFiddle/Codesandbox (required)**
https://codepen.io/Hawkeye64/pen/gVVNWL?editors=1010

This is happening because the QField and mask mixins do not know about each other. So, in QField, this code:
```js
    hasValue () {
      const value = this.__getControl === void 0 ? this.value : this.innerValue

      return value !== void 0 &&
        value !== null &&
        ('' + value).length > 0
    },
```
Will cause the clearable icon to be displayed because of the "fill-mask" property.

Somewhere, there needs to be a a check: `this.fillMask !== this.innerValue` to verify no additional data has been entered by the user.

I suspect there will be needing something like this:
QField.js
```js
    hasValue () {
      const value = this.__getControl === void 0 ? this.value : this.__getInnerValue === void 0 ? this.innerValue : this.__getInnerValue()

      return value !== void 0 &&
        value !== null &&
        ('' + value).length > 0
    }, 
```
Notice the `this.__getInnerValue`, then in...
QInput.js
```js
    __getInnerValue () {
      if (this.hasMask === true && this.fillMask !== false) {
        return this.__unmask(this.innerValue)
      }
      return this.innerValue
    },
```

If this looks ok, I can PR and test it.

hawkeye64:
any traction on this?

hawkeye64:
Tested with bleeding edge and this is no longer an issue.

Issue #4875, opened 2019-08-12
Unmasked QInput text overlays


hawkeye64:
![image](https://user-images.githubusercontent.com/10262924/62894018-19fee000-bd09-11e9-8c7e-6b6f9b5beb62.png)

size for first one is: 244x15
size for second one: 244x20

The seond one is missing reference to this class:
```
.q-field--float .q-field__label {
    transform: translate3d(0,-40%,0) scale3d(0.75,0.75,0.75);
    right: -33.33333%;
}
```
This is the css needed.
```css
    &.q-field--float .q-field__label
      transform translate3d(0, -30%, 0) scale3d(.75, .75, 1)
```
QField.js (see last item):
```js
    classes () {
      return {
        [this.fieldClass]: this.fieldClass !== void 0,

        [`q-field--${this.styleType}`]: true,
        'q-field--rounded': this.rounded,
        'q-field--square': this.square,

        'q-field--focused': this.focused === true || this.hasError === true,
        'q-field--float': this.floatingLabel,
...
```

hawkeye64:
Disregard the reference above. I messed up and used the wrong one.

hawkeye64:
This next reference is good.

hawkeye64:
any traction on this?

rstoenescu:
Hi,

I'm a bit confused by this, because bleeding edge works correctly. Investigated a bit and the floating label logic does not seem to need any fixes.

`floatingLabel` computed prop looks at (amongst other things):

```
(
          this.inputValue !== void 0 && this.hideSelected === true
            ? this.inputValue.length > 0
            : this.hasValue === true
        )
```

In this case, it takes `hasValue`, which in turn looks at `innerValue` which is correctly filled.

Can you re-test with bleeding edge pls? What am I missing?

hawkeye64:
Will do.

rstoenescu:
Thx. I mean, nothing changed to this respect into the code for some time. So wondering if I am missing something.

hawkeye64:
Either way, I just built with bleeding edge (yarn build in ui, copy to node_modules/quasar/dist in docs, quasar dev, etc) and it no longer appears to be an issue:
![image](https://user-images.githubusercontent.com/10262924/63637191-f86df480-c636-11e9-9c6e-0125fc59c251.png)

PR #4879, opened 2019-08-12
fix: QInput with mask, fill-mask and clearable always show clearable icon #4874

hawkeye64:
Tested, added doc examples


hawkeye64:
Ignore the "fix: Unmasked QInput text overlays #4875" I just labeled the commit wrong and then tried to change it to the proper text.

hawkeye64:
ref: https://github.com/quasarframework/quasar/issues/4874

hawkeye64:
Tested with bleeding edge and this is no longer an issue.

PR #4931, opened 2019-08-20
Fix: correctly applies content-style on QScrollArea root elements

nachodd:
Add's static style (content-style prop) to the root element, when it's on desktop. Otherwise, style cannot be applied to the root element.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/ui/src/components/scroll-area/QScrollArea.js
+++ b/ui/src/components/scroll-area/QScrollArea.js
@@ -263,6 +263,7 @@ export default Vue.extend({
 
     return h('div', {
       staticClass: 'q-scrollarea',
+      style: this.contentStyle,
       on: {
         mouseenter: () => { this.hover = true },
         mouseleave: () => { this.hover = false }


rstoenescu:
Hi,

Thank you for contributing!

There's a bit of misunderstanding with this PR. The whole reason for this prop is that it gets applied to either the root (on mobile) or the underlying correct node when on desktop (through mainStyle computed property). Notice there is "content-style" and "content-active-style" props, and their explanation in the API card is: "Object with CSS properties and values for styling the **container** of QScrollArea" and "Object with CSS properties and values for styling the **container** of QScrollArea when scroll area becomes active (is mouse hovered)"; the container differs and the reason for having content-style is exactly the fact that there exists a "content-active-style" too, so it auto-switches between the two based on the state.

Should you wish to apply some style to the root node, regardless if on mobile or desktop, then directly use Vue's style, which applies style to your desired node, similar to what you are doing on this PR.

Issue #4687, opened 2019-07-18
The public path setting with history mode


muei:
```
Operating System - Linux(5.0.0-21-generic) - linux/x64
NodeJs - 12.6.0

Global packages
  NPM - 6.10.1
  yarn - 1.16.0
  @quasar/cli - 1.0.0
  cordova - Not installed

Important local packages
  quasar - 1.0.5 -- Build high-performance VueJS user interfaces (SPA, PWA, SSR, Mobile and Desktop) in record time
  @quasar/app - 1.0.4 -- Quasar Framework local CLI
  @quasar/extras - 1.2.0 -- Quasar Framework fonts, icons and animations
  vue - 2.6.10 -- Reactive, component-oriented view layer for modern web interfaces.
  vue-router - 3.0.7 -- Official router for Vue.js 2
  vuex - 3.1.1 -- state management for Vue.js
  electron - Not installed
  electron-packager - Not installed
  electron-builder - Not installed
  @babel/core - 7.5.5 -- Babel compiler core.
  webpack - 4.35.3 -- Packs CommonJs/AMD modules for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand. Support loaders to preprocess files, i.e. json, jsx, es7, css, less, ... and your custom stuff.
  webpack-dev-server - 3.7.2 -- Serves a webpack app. Updates the browser on changes.
  workbox-webpack-plugin - 4.3.1 -- A plugin for your Webpack build process, helping you generate a manifest of local files that workbox-sw should precache.
  register-service-worker - 1.6.2 -- Script for registering service worker, with hooks

Quasar App Extensions
  @quasar/quasar-app-extension-dotenv - 1.0.0-beta.10 -- Load .env variables into your quasar project
```

This issue is related to https://github.com/quasarframework/quasar/issues/4460 ,  https://github.com/quasarframework/quasar/issues/4451 .
My frontend structure like this:
1. `quasarexample.com` is the root host.
2. Below this host, there are not one frontend project. For example,
- `quasarexample.com/site1`
- `quasarexample.com/site2`
- `quasarexample.com/about`
3. **site1**, **site2** are stand alone project, and written by quasar. And proxy by nginx or aws cloudfront.
4. All static files like *css*, *js*, *images*..., are stored in cloud, and **CDN** like `https://cdn.quasarexample.com/js/xsdfdfsdfs.js`.
[...]

PR #4962, opened 2019-08-23
fix(devserver) Remove "prod" requirement for using publicPath

### LINKED ISSUE #4840: Exposing build.appBase variables for user-defined configuration
Yes. I'm developing a SPA website. After building, all files (includes `js, css, img, statics, fonts, index.html` and so on) are hosted on a CDN server like `https://public.example.com/**/` However, the address of the website is different from the CDN server, it's like `https://spa.example.com`. So when you want to replace or push a router like `https://spa.example.com/login` with `https://spa.example.com/home`, it will redirect to `https://public.example.com/**/home`, which is wrong.
And it seems caused by some code in [quasar-config.js](https://github.com/quasarframework/quasar/blob/f70bf797301002fc9c0f0403a44c574dc2559062/app/lib/quasar-config.js#L465): 
`cfg.build.appBase = cfg.build.vueRouterMode === 'history'
      ? cfg.build.publicPath
      : ''
`
And it will also cause a `replaceState` error when enable `scrollBehavior` in `vue-router`.

When replacing `cfg.build.publicPath` above with `'/'`, the problem is disappeared.
So currently, I'm using `'history'` in `/router/index.js` like `new VueRouter( mode: 'history', routes: [...]})`, not the example in the `quasar.config.js` like `build: { vueRouterMode: 'history' }`, to avoid `appBase` be assigned to a value like `https://public.example.com/**/` .
However, the appBase is empty `''` now, when website has subroute like `https://spa.example.com/home/subroute`, it crashed.
I also tried `<base href="/">` in `index.template.html`, but it will cause an error when building.
I'm hoping a clearer way to achieve this like `build: { appBase: '/' }` or something better.

Distinguish `publicPath` with CDN and the website's real host. And files may be host on website's domain or CDN or both at the same time.

**Additional context**
In addition, none `statics` file is used in this project, all static file like `img` file is process as `assets`, and is hosted on `https://public.example.com/**/` .

EEParker:
This PR allows `publicPath` and the `base` tag to work on dev builds. It is used in conjunction with webpackDevServer publicPath.

- [x] Bugfix
- [x] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

Fixes #4840, #4687 

**quasar.conf.js**
```javascript
    build: {
      publicPath: '/path/to/app1/', 
      forceDevPublicPath: true, // this allows publicPath to work in dev and prod builds
      vueRouterMode: 'history',
    },

    devServer: {
      publicPath: '/app1/', // this can match the publicPath from above or be part of the subpath. If using a proxy, set it to match.
    },
```

### DIFF
--- a/app/lib/quasar-config.js
+++ b/app/lib/quasar-config.js
@@ -376,6 +376,7 @@ class QuasarConfig {
       webpackManifest: this.ctx.prod,
       vueRouterMode: 'hash',
       preloadChunks: true,
+      forceDevPublicPath: false,
       // transpileDependencies: [], // leaving here for completeness
       devtool: this.ctx.dev
         ? '#cheap-module-eval-source-map'
@@ -474,7 +475,7 @@ class QuasarConfig {
     }
 
     cfg.build.publicPath =
-      this.ctx.prod && cfg.build.publicPath && ['spa', 'pwa'].includes(this.ctx.modeName)
+      (this.ctx.prod || cfg.build.forceDevPublicPath) && cfg.build.publicPath && ['spa', 'pwa'].includes(this.ctx.modeName)
         ? formatPublicPath(cfg.build.publicPath)
         : (cfg.build.vueRouterMode !== 'hash' ? '/' : '')
 
@@ -711,7 +712,7 @@ class QuasarConfig {
       }
     }
     else {
-      cfg.build.env.__statics = `"${this.ctx.dev ? '/' : cfg.build.publicPath || '/'}statics"`
+      cfg.build.env.__statics = `"${((this.ctx.prod || cfg.build.forceDevPublicPath) && cfg.build.publicPath) ? cfg.build.publicPath : '/' }statics"`
     }
 
     appFilesValidations(cfg)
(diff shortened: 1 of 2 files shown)


rstoenescu:
Hi @EEParker ,

Thanks for contributing!

I cannot merge this as it would break some Quasar modes. However, what we could do is setup a new Boolean flag under quasar.conf.js > build, name it "forceDevPublicPath", then scenarios like yours can be covered too.

I'm just wondering over the very large implications of this. Were you able to test the changes in this PR? Need help getting you up to speed on testing CLI changes?

Also, would need some doc updates for the new flag. Again, I can show you quickly how to do it (ping me on Discord) -- if this is unclear how to do. Should be a 1-2 minute thing to add this to docs.

rstoenescu:
Merged it and pushed a subsequent commit to bulletproofed it.

@EEParker Thanks for starting this and taking time for it for a PR! Much appreciate it.
Will be available in quasar/app 1.0.6 shortly.

PR #4978, opened 2019-08-26
feat(docs): enhancements to DocsApi

hawkeye64:
You can pick and choose what you wish to keep.


rstoenescu:
My only concern here is that it adds CodePrism dep into the client bundle, and we want to avoid extra K being transferred over to the browser.

rstoenescu:
`apiCount` seems to no longer be used?

rstoenescu:
How about this (made slight changes):

card

PR #4984, opened 2019-08-28
fix(DocsApi): Don't show badge when 0 items in filter

hawkeye64:
This now behaves the same way as the inner tabs


rstoenescu:
Beautiful!

hawkeye64:
Thanks :)

PR #5017, opened 2019-09-02
fix(DocApi):  top tabs show incorrect data for plugins and directives

hawkeye64:


hawkeye64:
Fixes the following:
image
image

and
image

image

PR #5018, opened 2019-09-02
feat(DocApi): gives a different bg color for each name level

hawkeye64:


hawkeye64:
This PR assigns a different background color for each level of "name" property.
image

PR #5155, opened 2019-09-25
fix(app): incorrect case for Generator

hawkeye64:


rstoenescu:
It's the correct case, cause it's the instance, not the class :)

PR #5185, opened 2019-09-29
QZoom

hawkeye64:
New core component. Added to dev and docs.


hawkeye64:
blah, looks messy because I had to get lastest from dev (sorry about that).

PR #5222, opened 2019-10-04
feat(QSelect): implement inputClass and inputStyle

### LINKED ISSUE #4351: [REQUEST] QSelect text alignment
Could you please add support for right side text alignment for value, label and options?

Demo
https://jsfiddle.net/fvqs1n5y/6/

Version
v1.0.0-rc.4

### LINKED ISSUE #4636: [v1] Change text color of QSelect inner QField

When using a QSelect component, I want to set the color of the inner QField displaying the selected item label and icons. I expected the `color` prop to do this, and it seems to be adding the `.text-<color_variable_name>` class to the inner element of the internal QField, but that class is only affecting the label and the border (if any), and it's effective only when the QSelect is focused. I assume this is the expected behavior.

Adding a prop to the QSelect (and maybe QField?) component to change the color of all text/icons inside its QField. I'm aware I can do that now by using the different slots available. I.e: If I want a QSelect with a prepended icon in which the color of said icon, the selected label, and the dropdown icon is primary, I can do something like this:

```
<q-select
    hide-dropdown-icon
    emit-value
    map-options
    label="Order By"
    stack-label
    v-model="currentSelectedOrder"
    :options="orderingOptions"
>
    <template #prepend>
        <q-icon name="sort" color="primary" />
    </template>
    <template #selected-item="{ opt }">
        <span class="text-primary">{{ opt.label }}</span>
    </template>
    <template #append>
        <q-icon name="arrow_drop_down" color="primary" />
    </template>
</q-select>
```

But I would find it useful to be able to have the same QSelect without redefining the slots. Let's name the prop to do this `text-color`, and the same QSelect could be achieved like this:
```
<q-select
    hide-dropdown-icon
    emit-value
    map-options
    label="Order By"
    stack-label
    v-model="currentSelectedOrder"
    :options="orderingOptions"
    text-color="primary"
/>
```

pdanpdan:
Commented code for strict updating of class

close #4636, close #4351

### DIFF
--- a/ui/dev/components/form/select-part-4.vue
+++ b/ui/dev/components/form/select-part-4.vue
@@ -27,6 +27,70 @@
         </q-select>
       </div>
     </div>
+
+    <div class="q-pa-md">
+      <div class="q-gutter-md row">
+        <q-select
+          filled
+          v-model="model"
+          use-input
+          input-debounce="0"
+          label="Styled input - string"
+          :options="options"
+          @filter="filterFn"
+          style="width: 250px"
+          input-class="text-green text-right"
+        >
+          <template v-slot:no-option>
+            <q-item>
+              <q-item-section class="text-grey">
+                No results
+              </q-item-section>
+            </q-item>
+          </template>
+        </q-select>
+
+        <q-select
+          filled
+          v-model="model"
+          use-input
+          input-debounce="0"
+          label="Styled input - array"
+          :options="options"
+          @filter="filterFn"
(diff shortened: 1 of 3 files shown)


grojguy:
@pdanpdan @rstoenescu 
This is not working for all applicable components, as far as I can tell.  Quasar v1 or v2.  See [this comment](https://github.com/quasarframework/quasar/issues/4351#issuecomment-1507676370).

PR #5291, opened 2019-10-12
fix(docApi): filter gives false-positive for "props" when nothing found

hawkeye64:


hawkeye64:
Before fix:
image

After fix:
image

Fix, with something found:
image

hawkeye64:
ok, I don't know why this did the wrong update - will look into it

PR #5292, opened 2019-10-12
fix(docApi): filter gives false-positive for "props" when nothing found

hawkeye64:


hawkeye64:
Sorry, I branched off of my branch on this one. Let me know if you would like me to redo it.

Before fix:
image

After fix:
image

Fix, with something found:
image

PR #5297, opened 2019-10-13
feat(IDE) Update web-types to the latest spec

panstromek:
Web types spec was changed a little bit (slots and scoped slots were merged), so I updated the generating script.

Reference: https://github.com/JetBrains/web-types/pull/10


rstoenescu:
Thx for the update!

PR #5298, opened 2019-10-13
feat(docs) Tab to search [WIP]

### LINKED ISSUE #5013: [docs] "Tab to search" functionality for docs
This feature allows you search a page directly by just typing its URL and "TAB"  key. It works on Quasar forum for example (but also on YouTube, Facebook, StackOverflow etc.).

This would be very handy, because you can skip two steps when looking up a docs page.

What we need to do to achieve this:
 1. Sync search input with URL query (or route param, but query is probably easier with current setup)
 2. Add a search manifest xml file with url template (it's like 5 lines, very simple file)
 3. Make sure that loaded page shows the results in popup

Advanced:
 4. We could also fetch search results on the server and redirect user directly to the top result (with search popup still open, I guess). That's probably more complicated work, though.

reference: https://www.chromium.org/tab-to-search

panstromek:
Adds "Tab To Search" functionality for docs.

Demo: https://tab-to-search-test.netlify.com - copy the url to search bar and hit tab (You might need to visit the page first)

fixes #5013

TODO
 - [x] Sync search input with URL query (or route param, but query is probably easier with current setup)
 - [x] Add a search manifest xml file with url template (it's like 5 lines, very simple file)
 - [x] Make sure that loaded page shows the results in popup

 - [ ] Move XML to CDN and replace hardcoded statics url with `cdn.quasar.dev` one (This is on Razvan since I don't have rights for the CDN)

Optional
 - [x] ~~Fetch results on server and redirect to the top result directly~~
        (Better to do in separate PR, it's complicated and I don't have time for that)

### DIFF
--- a/docs/src/index.template.html
+++ b/docs/src/index.template.html
@@ -12,6 +12,7 @@
     <meta property="og:image" content="https://cdn.quasar.dev/logo/512/quasar-logo.png" />
     <meta property="og:site_name" content="Quasar Framework" />
 
+    <link type="application/opensearchdescription+xml" rel="search" href="/statics/search_manifest.xml">
     <link rel="icon" type="image/svg+xml" href="https://cdn.quasar.dev/logo/svg/quasar-logo.svg">
     <link rel="icon" type="image/png" sizes="128x128" href="https://cdn.quasar.dev/app-icons/favicon-128x128.png">
     <link rel="icon" type="image/png" sizes="96x96" href="https://cdn.quasar.dev/app-icons/favicon-96x96.png">
--- a/docs/src/layouts/Layout.vue
+++ b/docs/src/layouts/Layout.vue
@@ -299,7 +299,13 @@ export default {
   },
 
   mounted () {
-    this.leftDrawerState = !!this.$route.query.search
+    // If we have a search string in the query (mostly from tab-to-search functionality),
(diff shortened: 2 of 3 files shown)


panstromek:
@rstoenescu I forgot about this PR a bit. 

Would you be ok with this kind of approach? If so, I can resolve conflicts and we can merge it.

It's not perfect but it does the job :wink: and helps a bit when searching.

rstoenescu:
This might just work. Yes, pls do update it. Thanks for taking care of this.

panstromek:
Rebased, fixed some little problems with initial focus and added clarifying comments. 

This should be ready to merge 😉 Except that you will probably want to move the XML file on the cdn (atm it's in statics)


PS: I also updated the test webpage on netlify, but since I had a wrong version on that url for some time and chrome seems to cache that XML file despite disabling the cache, for me it points to `quasar.dev` instead of `tab-to-search.netlify.com` but other than that it works fine.

try to click here [https://tab-to-search-test.netlify.com/start/pick-quasar-flavour?search=notify](https://tab-to-search-test.netlify.com/start/pick-quasar-flavour?search=notify) to get the correct behaviour. Tab to search should link you to that url.

rstoenescu:
Awesome work. Will move the xml to CDN.

PR #5356, opened 2019-10-20
fix(QTable): watch for changes when filter is an object

stefanvanherwijnen:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Here is a demo of the problem:
https://codepen.io/stefanvh/pen/RwwoBWg

The changes in the filter object are not detected, thus a new request is never performed and the data is not filtered when using server-side filtering.

### DIFF
--- a/ui/src/components/table/table-filter.js
+++ b/ui/src/components/table/table-filter.js
@@ -13,10 +13,13 @@ export default {
   },
 
   watch: {
-    filter () {
-      this.$nextTick(() => {
-        this.setPagination({ page: 1 }, true)
-      })
+    filter: {
+      handler: function (val) {
+        this.$nextTick(() => {
+          this.setPagination({ page: 1 }, true)
+        })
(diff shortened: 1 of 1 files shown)


rstoenescu:
Hi,

Thanks for contributing!
But why would you not supply `filter="filter.name"` directly to QTable?

stefanvanherwijnen:
Thanks. 
I would like to filter 2 fields. E.g. `filter = { name: '', surname: ''}. This is why filter needs to be an object.

I could also stringify and parse the object but I think it makes more sense to pass the actual object.

rstoenescu:
Ok. Will be available in "quasar" v1.2.5.
Thanks for contributing!

Issue #4930, opened 2019-08-20
Error in callback for watcher "value": "TypeError: Cannot read property 'desktop' of undefined" found in <QMenu>


nikoz84:
I have this problem when import a new QSelect component, 

`
app.js:67460 [Vue warn]: Error in callback for watcher "value": "TypeError: Cannot read property 'desktop' of undefined" found in 
---> <QMenu>
       <QSelect>
         <SearchForm> at resources/assets/js/forms/SearchForm.vue
           <Table> at resources/assets/js/components/Table.vue
             <Listar> at resources/assets/js/components/ListAdmin.vue
               <Admin> at resources/assets/js/pages/Admin.vue
                 <QPageContainer>
                   <QLayout>
                     <Default> at resources/assets/js/layout/Default.vue
                       <Main> at resources/assets/js/layout/Main.vue
                         <Root>
`

I use quasar without quasar CLI in standalone way, my app file look that way

![Captura de tela de 2019-08-20 08-37-28](https://user-images.githubusercontent.com/6708508/63344523-02d27a80-c327-11e9-9f68-80bb8f9ebdb6.png)

And when debug in the transpiled file app.js 

![Captura de tela de 2019-08-20 09-09-16](https://user-images.githubusercontent.com/6708508/63346281-ffd98900-c32a-11e9-8819-a43f45629030.png)

Thanks

yusufkandemir:
@nikoz84 you need to provide more information. Also as you are using the UMD version, you don't need to use `Vue.use(Quasar, //...`, Quasar UMD already calls that for you. If you need to change the `iconSet`, you need to do `Quasar.iconSet.set(Quasar.iconSet.fontawesomeV5/* or something else */)`, MaterialIcons is already the default so you can remove the `Vue.use(Quasar, //...`, part completely. You can check [the docs for icon sets](https://quasar.dev/options/quasar-icon-sets#UMD-Way).

[...]

PR #5366, opened 2019-10-22
fix: Removed dekstop check in escape-key.js #4930

thijs981:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
In some cases, the Platform.js doesn't have the .is property. This will throw an exception and causes dropdowns (QSelect, QMenu, etc.) not to work. With this check removed, the Esc key will always be installed without throwing exceptions. This will also work on mobiles even though there is no Esc key.

### DIFF
--- a/ui/src/utils/escape-key.js
+++ b/ui/src/utils/escape-key.js
@@ -16,18 +16,14 @@ export default {
   },
 
   register (comp, fn) {
-    if (Platform.is.desktop === true) {
-      this.__installed !== true && this.__install()
-      handlers.push({ comp, fn })
-    }
+    this.__installed !== true && this.__install()
+    handlers.push({ comp, fn })
   },
 
   pop (comp) {
-    if (Platform.is.desktop === true) {
-      const index = handlers.findIndex(h => h.comp === comp)
-      if (index > -1) {
-        handlers.splice(index, 1)
-      }
+    const index = handlers.findIndex(h => h.comp === comp)
+    if (index > -1) {
+      handlers.splice(index, 1)
     }
   }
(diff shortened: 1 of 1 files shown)


pdanpdan:
This shouldn't be solved in this way.
If you don't have Platform populated it means you are not initializing Quasar in the right way.

I suppose it's related to #4930.  Do:

```javascript
Vue.use(Quasar, {
  pugins: { Platform, ... },
  ...
})
```

Plugins must be an object, not a list.

shandrolis:
> This shouldn't be solved in this way.
> If you don't have Platform populated it means you are not initializing Quasar in the right way.
> 
> I suppose it's related to #4930. Do:
> 
> ```js
> Vue.use(Quasar, {
>   pugins: { Platform, ... },
>   ...
> })
> ```
> 
> Plugins must be an object, not a list.

Changing it into an object rather than an array doesn't change anything; it's still broken.

Do note that I dont 'new Vue' anything in my component library, and rather export its components directly, which are then imported in my web-apps. So, while my web-app itself does have the Platform plugin available just fine, perhaps that is what causes the exported component to not be able to access it? This was never an issue pre 1.0 though.

rstoenescu:
Platform should also no be required to be specified in the initialization plugin list, because it is automatically installed.

rstoenescu:
At least try to place Vue.use(Quasar, ....) at the very top of your main.js.

Issue #5351, opened 2019-10-19
QDialog [ios] [ipados] Keyboard is overflowing inputfields on Dialogs


ibrainventures:
Dear Team,

that is my major issue on the QDialog. If you open a standard Dialog and there is a Input Field the virtual Keyboard overflow the entry-fields.

So no entry is possible .. Scrolling to the fields else is also prevented.

With the seamless Option, the entries are possible, but the scroll / overall handling of that seamless Dialog is awkward (Background-Scrolling etc.)

For Testing there is a Sandbox:

https://vtft7.sse.codesandbox.io/
e: https://codesandbox.io/s/codesandbox-app-vtft7

If you make a slow motion .. you will see, that standard Dialog scrolls short up - but it snaps down .. 

That **it is possible, to have a working Dialog** with Input on ios Devices, you see on my codepen:

https://codepen.io/ibrainventures/full/eYYBdRE
e: https://codepen.io/ibrainventures/pen/eYYBdRE

(here the Dialog is NOT snapping down - opened in an non seamless Dialog)

May the reason that it is working there (same Dialog Code) is, that the codepen wrapper is blocking some dyamic body-attaches by the quasar fw  .. (i think so) (there are some .js routines).

So the  solution is maybe to make that to the body element dynamic applied styles / classes (more) ios compatible - so that the explained snap-down will not occur...

This issue is really a big problem for me, because all kind of Dialogs (Login, Signup, Edit, etc.) are not usable for any iphone / ipad User.

Affected: All Apple i-devices with safari, IOS: 11, 12, 13, IpadOS: 13

It would be really great to get here a solution. Many thanks to the Team.

pdanpdan:
There is not much we can do on the framework side.
For iOS you have to set some CSS for the dialog, but this CSS is very specific for each use-case.
The basic idea is this:
- detect if the dialog has a field that requires keyboard input focused
- if it has then add a class on the dialog and style it so that it covers at most 50vh and is aligned on the top of the screen

You can check an example here: https://codesandbox.io/s/codesandbox-app-44xoy

ibrainventures:
Hi Dan,

its a little funny to read this bloody workaround which i posted myself a week ago inside the discord components channel: (wednesday , 12:23)

...  make those input fields nearly above the virt.keyboard and  made a resize of the dialog at the focus event (on the input) .. but with landscape  .... 

but i also wrote, that to handle all use-cases is nearly impossible .. :-(

[...]

Issue #4264, opened 2019-05-31
[v0.17][v1] Safari scrolls back to the QInput when focused


jean-moldovan:
Hello!

This one is Safari specific. Tested on Quasar v0.17. Safari version 12.1.

I'm not 100% sure that something is wrong with the QInput. However, was unable to reproduce it for a native input. When you have a bunch of inputs on a page, and one of them is focused, if you scroll down, Safari scrolls back up to that input. The harder you try, the faster Safari goes. A game that you can never win.

https://jsfiddle.net/qnase9dg/1/

Steps to reproduce:
1) Click inside second input
2) Scroll down and wait
3) Safari takes you back to that input
4) Uncomment native input, try again. Works ok.

![safari-q-input](https://user-images.githubusercontent.com/3635352/58693577-c0d61000-8399-11e9-9336-c7bfa1a66218.gif)

Thanks!

jean-moldovan:
Happening on v1 as well https://jsfiddle.net/b3yt60kw/2/.

rstoenescu:
Hi,

Should no longer happen in Quasar v1.1.0. Please reopen if I missed something.

jean-moldovan:
@rstoenescu , I'm still able to reproduce on v1.1.0: https://jsfiddle.net/b3yt60kw/2/
![test](https://user-images.githubusercontent.com/3635352/64940984-157c9880-d86e-11e9-8ffb-4e23075bd575.gif)

jean-moldovan:
![Dec-02-2019 15-28-09](https://user-images.githubusercontent.com/3635352/69959522-679d9380-1518-11ea-8402-42056ced1a18.gif)
Still seeing this on v1.5.2

pdanpdan:
Yes, it's still open, sorry.

rstoenescu:
Fix will be available in "quasar" v1.5.10. Unfortunately, it can only be fixed for iOS 13+

jean-moldovan:
This is still an issue on desktop Safari 13.0.4 and Quasar v1.5.11

![q-safari](https://user-images.githubusercontent.com/3635352/71347065-82dd4b00-257a-11ea-9425-7a9771677226.gif)

PR #5377, opened 2019-10-23
feat(QDialog): Detect iOS soft keyboard show/hide and adjust QDialog

pdanpdan:
ref #4264, ref #5351


ibrainventures:
https://github.com/quasarframework/quasar/issues/5351#issuecomment-545417848

PR #5391, opened 2019-10-24
feat(QTable): support for nested row key

stefanvanherwijnen:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Currently, only a string can be used for the row-key. In some cases the key (ID) is nested inside an object (https://github.com/mrichar1/jsonapi-vuex). This PR provides the ability to use a nested row key similar to the `field` property of a column.

### DIFF
--- a/ui/src/components/table/QTable.js
+++ b/ui/src/components/table/QTable.js
@@ -43,7 +43,7 @@ export default Vue.extend({
       default: () => []
     },
     rowKey: {
-      type: String,
+      type: [String, Function],
       default: 'id'
     },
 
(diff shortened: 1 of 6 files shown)


rstoenescu:
Thank you for contributing!
This will get into "quasar" v1.2.6.

rstoenescu:
Will make some small adjustments though.

stefanvanherwijnen:
Thanks, using a computed property looks much better indeed.

PR #5421, opened 2019-10-29
Refactor all components using target/scroll-target to use the same interface #5906

### LINKED ISSUE #5051: Refactor all components using target/scroll-target to use the same interface
Grouping #4963, #4832. #4695, #3644

- allow passing CSS selector or DOM node in `scroll-target`
- unify code

pdanpdan:
- target accepts (in QMenu, QPopupProxy, QTooltip):
  - true => auto find target based on first parent (starting from parentEl) without .q-anchor--skip class
  - false => no anchor
  - Vue instance => uses $el of that instance
  - CSS selector or DOM element => that element

- scroll-target accepts (in QInfiniteScroll, QMenu, QScrollObserver, QParalax, QPullToRefresh, QTooltip, QVirtualScroll)
  - null or void 0 => auto detect using getScrollTarget
  - Vue instance => uses $el of that instance
  - CSS selector or DOM element => that element
  - in the end, if the scroll target is document, document.body, document.scrollingElement or document.documentElement, window is selected as scroll target
  

close #5051, ref #5474

### DIFF
--- a/docs/src/pages/vue-components/scroll-observer.md
+++ b/docs/src/pages/vue-components/scroll-observer.md
@@ -15,10 +15,7 @@ Scroll this page to see the example below in action.
 <doc-example title="Basic" file="QScrollObserver/Basic" />
 
 ## Determining Scrolling Container
-All components or directives in Quasar have a simple algorithm to determine the container that supports the scroll:
-- if a `scroll-target` property is available on the component then it tries to use it as scroll container
-- then it searches for a parent DOM element which has the `scroll` Quasar CSS Helper class attached to it,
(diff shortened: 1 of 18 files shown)


mesqueeb:
@pdanpdan enlighten me, will the create the ability to have QMenu available as plugin, just like QDialog?

pdanpdan:
@mesqueeb No, why would it do something like this?

mesqueeb:
@pdanpdan Because you said it would close #5474 which was such a request 😅

pdanpdan:
Thanks for pointing it out - it was just related to it.

rstoenescu:
This will get into "quasar" v1.7.5.

PR #5451, opened 2019-11-02
Update notify.md

bartduisters:
Add extra information for the guide on adding the Notify plugin.


rstoenescu:
Hi,

Thanks for contributing!
It was right there in the docs, but a more explicit way works fine too :)

notify

PR #5462, opened 2019-11-04
fix(Notify plugin): actions CSS - sane default color (black)

euphemism:

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Changing a default CSS color

- [ ] Yes
- [x] No

**Other information:**
The default text color for the notification's actions is an odd purple. I figure having the default be black makes more sense.

### DIFF
--- a/ui/src/plugins/Notify.sass
+++ b/ui/src/plugins/Notify.sass
@@ -50,7 +50,7 @@ body.q-ios-padding .q-notifications__list
     padding: 8px 0
 
   &__actions
-    color: #c581ff
+    color: #000000
 
   &--standard
     padding: 0 16px
--- a/ui/src/plugins/Notify.styl
+++ b/ui/src/plugins/Notify.styl
@@ -50,7 +50,7 @@ body.q-ios-padding .q-notifications__list
     padding 8px 0
 
   &__actions
-    color #c581ff
+    color #000000
 
   &--standard
     padding 0 16px


rstoenescu:
Hi,

Thank you for contributing!
The current color is per Material guidelines though...

snackbar

euphemism:
Hi Razvan, thank you for the prompt response. I have not been able to find anything on [material.io](https://material.io/) that actually dictates that the default color of the action is supposed to be this particular purple. I see things like the following:

* To distinguish the action from the text label, text buttons should display colored text.
* The text label shouldn’t share the same color as the text button.
* Don’t use a filled or elevated button in a snackbar, as it draws too much attention.

The specifications site is using a range of purples as the primary color, but there is nothing to indicate that this is expected to be the default. Looking at the snackbar's spec for color, it only specifies the color for the bar itself, not the action button: 
material_design_snackbar_color

rstoenescu:
But regardless, pure black is not a good color for it...

notify-change

euphemism:
I would agree with that. For some reason I was thinking of a dialog with a white background. I noticed the purple on a notification that I had changed the default background color of, and it didn't occur to me when I made this quick change that Quasar's snackbars are a dark grey by default. If I change this to a lighter grey or white would that be acceptable?

euphemism:
It might make sense to have this default to the primary color.

Issue #5562, opened 2019-11-12
Datepicker: navigate with touch events (swipe)


getriebesand:
**Is your feature request related to a problem? Please describe.**
Changing month on smartphones is hard only with the arrow buttons.

**Describe the solution you'd like**
I want to swipe left and right on calendar to change month (like other apps with date picker on android)

**Describe alternatives you've considered**
I could implement it by my self, but i think it will be a nice feature for quasar with focus on mobile apps

**Additional context**

pdanpdan:
I think that the best way is to implement it on the user side - I added an example here (uses a few internals but should be safe): https://codepen.io/pdanpdan/pen/qBdExzK?editors=1010

PR #5585, opened 2019-11-14
feat(QDate): Add swipe navigation and fix RTL transitions #5562

pdanpdan:
- in Calendar view - change month
- in Months view- change year
- in Years view - change group of years


rstoenescu:
Please include only the RTL fix, but without the TouchSwipe.

Issue #5595, opened 2019-11-15
Bug Drag&Drop in Mozila browser


Teffety:
**Describe the bug**
`MouseEvent.mozPressure is deprecated. Use PointerEvent.pressure instead.`

**Codepen/jsFiddle/Codesandbox (required)**
https://codepen.io/pen/?&editable=true&editors=111

**To Reproduce**
u can see error in your application in your https://codepen.io/pen/?&editable=true&editors=111;

**Platform (please complete the following information):**
Browsers: Mozila

**Additional context**
when using  `mutation drag&drop` i see this warring. And function drag&drop doesn`t  work after "onDragStart"

Teffety:
Mutation Drag&Drop bug in browser Mozila

pdanpdan:
You have linked to 2 empty codepens :)
But thanks to you I caught an error in some codepen examples that were using self closing divs.

Now coming back to Mozilla - I cannot reproduce it and I cannot find any refference to mozPressure in our code. Also you don't have a version of browser, and the fact that you didn't see the console full or red errors mentioning that vue cannot compile the template makes me think you were not checking too much :)

The fixed codepen https://codepen.io/pdanpdan/pen/YzzRXQQ?editors=1010 (with divs closed correctly) shows no refference to mozPressure.

Teffety:
Ok. Now a load screens with bug.
this is start example page in two browsers: Opera and Mozila.
![mozzila](https://user-images.githubusercontent.com/25255599/68951048-69542100-07ce-11ea-8373-ac99108e9bef.png)
![mozilaAbout](https://user-images.githubusercontent.com/25255599/68951050-69542100-07ce-11ea-90c9-4a5548019ffb.png)
![BrowserOpera](https://user-images.githubusercontent.com/25255599/68951051-69542100-07ce-11ea-8c5e-7eb2730f1886.png)

this is my code template in Vue.js and same methods;
![dragMethods](https://user-images.githubusercontent.com/25255599/68951098-77a23d00-07ce-11ea-81a0-c92e947494c9.png)
![code](https://user-images.githubusercontent.com/25255599/68951099-77a23d00-07ce-11ea-80ba-1c16b5ffce5c.png)

and when i start dragging in browser opera, i see that message(i know, that i  used console.warn() ) in console panel
![operaDraggingItem](https://user-images.githubusercontent.com/25255599/68951256-b932e800-07ce-11ea-8a04-c59bcedbf6ca.png)

[...]

PR #5597, opened 2019-11-15
fix(docs): Replace self closing <div/> tags in codepens with <div></div> because it breaks them #5595

pdanpdan:


rstoenescu:
This should be fixed instead of the examples: https://github.com/quasarframework/quasar/blob/dev/docs/src/components/Codepen.vue#L73-L79

PR #5608, opened 2019-11-15
feat(QTable): Skip adding __index to each row

pdanpdan:


rstoenescu:
This should be fixed instead of the examples: https://github.com/quasarframework/quasar/blob/dev/docs/src/components/Codepen.vue#L73-L79

rstoenescu:
What's the motivation behind it, besides a perf improvement? It is currently very useful with the slots usage and will be a breaking change if removed.

pdanpdan:
It mutates data retrieved from store.

PR #5620, opened 2019-11-17
docs(quasar-cli): Add webpack magic comments

ejez:
Add webpack magic comments `webpackInclude` and `webpackExclude` to 'lazy loading' doc


rstoenescu:
Thanks for the additions!

Issue #5476, opened 2019-11-05
[docs] Navigation bug


mseneshen:
I've found a bug that I'm able to reproduce on multiple browsers.

## Docs URL
https://quasar.dev/quasar-cli/quasar-conf-js

## Symptom
Clicking certain links on the page cause the navigation system to fail.

## Steps to reproduce
1. Navigate to https://quasar.dev/quasar-cli/quasar-conf-js
2. Click "Handling Webpack" or "Handling process.env" at the bottom of the page
3. Note that the current page address changes, but the page content remains the same
4. Clicking any internal navigation link on the page or in the sidebar after this point does not navigate away from the page.

smolinari:
I could at one point reproduce the issue. However, I refreshed my browser window and the problem went away. 

Scott

rstoenescu:
Hi,

Thanks for reporting.

Must have been a browser caching issue.
Cannot be a bug in docs.

mseneshen:
Hi @rstoenescu and @smolinari,

I'm able to reproduce this on every attempt in Chrome and Edge on desktop, and freshly reset Chrome and Firefox on Android, on separate networks too. I really don't think this could be a caching bug on my end.

Could you please try to reproduce this again?

smolinari:
What OS are you using on Desktop?

Scott

mseneshen:
Windows

Here is a reproduction using Browserling: https://imgur.com/OKrGsOE

You can test here: [Browserling](https://www.browserling.com/browse/win/10/chrome/56/https%3A%2F%2Fquasar.dev%2Fquasar-cli%2Fquasar-conf-js)

smolinari:
I got the problem again too. It seems like, when it happens, the different pages are loading, but the `quasar.conf.js` page is stuck showing its content. Did a Ctlr + F5 refresh, and it started working again. 

Scott

mseneshen:
It only starts working once the second page's URL is in the address bar and then you refresh into that new page, but not when refreshing the `quasar.conf.js` URL.

smolinari:
It does refresh on my machine when refreshing on the same page. The URL changes, the right side ToC changes, but the content on the page doesn't change. It's like it gets stuck.

Scott

mseneshen:
Should this be re-opened since the bug still exists?

ejez:
Issue traced down to having 2 quasar components in the same line:
changing:
`docs/src/pages/quasar-cli/quasar-conf-js.md` line 163:
[...]

PR #5637, opened 2019-11-18
fix(docs): Fix site navigation issue

ejez:
Fix router view not rendering in page https://quasar.dev/quasar-cli/quasar-conf-js

See https://github.com/quasarframework/quasar/issues/5476


rstoenescu:
Thanks for investigating!

PR #5670, opened 2019-11-21
fix(ui): table sort issue

hawkeye64:


hawkeye64:
https://cdn.discordapp.com/attachments/616163166677893148/647145867303845908/unknown.png
Table getting errors in docs
https://cdn.discordapp.com/attachments/616163166677893148/647150386767200267/unknown.png

https://cdn.discordapp.com/attachments/616163166677893148/647152089591447574/unknown.png

Just to keep backwards compatibility I left check for null, but appears it would just return `undefined`

Issue #5684, opened 2019-11-24
q-drawer fires x-hundred mouseout/mouseover/moseleave/mouseenter .. events on little mousemove


ibrainventures:
**Describe the bug**
I have two normal qdrawer with some text-divs, images etc, on both-sides.

At any minimal mousemovement between the elements / above the qdrawer elements there are many events fired. My usage is with desktop-mode / no breakpoint, and no mouseover-controlled-elements.

**Screenshots**

![grafik](https://user-images.githubusercontent.com/8803626/69501423-103e6880-0f05-11ea-8b66-5d4afd294ca9.png)

Is there a easy way to stop / remove those listeners?

Firefox Browser
Windows

rstoenescu:
Enhancement will be available in "quasar" v1.5.0

ibrainventures:
thank you and @pdanpdan for looking and solving this issue very much 🥇

PR #5706, opened 2019-11-27
feat(QDrawer): Only re-emit native mouse events as Vue events if there are listeners for them #5684

pdanpdan:


rstoenescu:
$listeners is not reactive

pdanpdan:
It is checking the existence of the listener before emitting, not on adding the native listeners.

It is just avoiding re-emitting the event.

PR #5725, opened 2019-11-28
feat(typescript): enhance typings 1

IlCallo:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

_In this batch_
Only simple changes to `utils` helpers signatures.

_Details_
- `debounce` and `throttle` now return input function type, instead of a generic untyped function
- `extend` signature has been made a bit more helpful, even if there is no way into TS to model an indefinite number of arguments of different shape. Now it's possible to explicitly state the result type at least

---

- `getBrand` return type now can also be `null`, to reflect scenarios where no color of given name is found on specified element

---

- `isValid` and `isBetweenDates` input types has been broadened, because they are wrapped anyway into a `new Date()` before usage
- `inferDateFormat` is now stricter
- `getDateBetween` reflect the possible absence of `min` and `max` parameters
- `isSameDate` last parameter got a stricter signature, its type was already present
- `clone` implementation is open to a lot of edge cases, because it will return as-is any value which is not a Date (including Objects, Arrays, etc which are passed by reference). I put some strict typings (still looser than current one), but a signature more adhering to implementation should be `function clone<D>(date: D): D;` which is much broader.

---

- `css` and `cssBatch` now have more typesafe signature

---

- as suggested by Kerry, all `Element` as been updated to `HTMLElement` into scroll methods
- `getScrollTarget` apparently had parameter and return types switched: `Window` doesn't have a `.closest` method, which is used by the implementation

### DIFF
--- a/ui/types/utils.d.ts
+++ b/ui/types/utils.d.ts
@@ -7,7 +7,15 @@ export * from './utils/event';
 
 // others utils
 export function openURL(url: string): void;
-export function debounce(fn: Function, wait?: Number, immediate?: boolean) : Function
-export function extend(...args: any[]): any;
-export function throttle(fn: Function, limit: Number): Function; 
+export function debounce<F extends (...args: any[]) => any>(
+  fn: F,
+  wait?: number,
+  immediate?: boolean
+): F;
+export function extend<R>(deep: boolean, target: any, ...sources: any[]): R;
+export function extend<R>(target: object, ...sources: any[]): R;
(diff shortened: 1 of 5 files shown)


rstoenescu:
Awesome!

PR #5729, opened 2019-11-28
feat(typescript): enhance types 2

IlCallo:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

Depending on the choices taken, the implementation of some functions will need to be updated slightly.

**Other information:**

_In this batch_
Changes to `utils` helpers signatures which **would require an implementation update too**.
Those implementation changes are not committed into the PR, just proposed into the following section.
This can also be a good moment to decide if TS signatures should be as much adherent to the implementation as possible, or if they should represent the "ideal" scenario (like functions had been written with TS) and be stricter and cleaner than their actual JS implementation.

I can give a try to update the implementation, after your feedback, but without automatic tests covering my back and most context still unclear, I'd prefer not to.

_Details_
- `BuildDateOptions` and `ModifyDateOptions` differ only by the `date`/`days` property, but inside `getChange`, `days` key name seems to be transformed into `date` anyway, so this can be probably reduced to a common `DateOptions` for all methods. Funny enough, it seems to me that if you provide a `date` key to `getChange` it should work anyway. To me, keeping a different APIs for very similar options isn't ideal and only one should be chosen and used
- `getDateDiff` return type was `Date`, while apparently a number is always returned. `unit` argument was `string` type, but this could cause an uncaught error if someone inserts a random string, because the inner switch doesn't have a default case (while it probably should), making the function effectively returning void in those scenario. I put `DateUnitOptions` to narrow thre type, **but this require to update switch case conditions** which atm are all plurals, while `DateUnitOptions` are all singular. I'd advocate to use the same set of singular words everywhere, if possible.
- `formatDate` `format` argument should be optional, looking at the implementation. `date` should be optional by implementation too, meaning that if you pass an invalid value the function will accept it and just return void. An `Error` should be thrown in those cases instead IMO, to fail fast. I don't think is possible to model current behavior in a meaningful way in TS.

---

- `ready` was missing any real typing. The first signature cover the case where a function is provided and we return its result, while the second covers all cases where executions stops immediately because provided parameter isn't a function. IMO, the second signature should be omitted in order to allow usage only when a function parameter is provided.

---

- Proposed `ListenOpts` version is more adhering to the current implementation, ~~but depending on where it used and how, you could still prefer previous one. I still don't have all use cases in mind, but I understood that the complexity of this type is given by the browser support or lack of support of providing an options object when defining a listener. @GordonBlahut you already had issues with this API, would this typings work for you or the more general ones are better?~~
After fixing usage with `removeEventListener`, seems like the new version is actually better than the old one. _This update should not be a breaking change_.
- an `event.getEventKey(evt)` is mentioned into documentation but I cannot find any other reference of it in the codebase, maybe it must be removed?

---

-  `between` return type seems to be always a number, never a boolean. What is the actual expected behaviour of this function? Current signature suggest that it should check if a value is into an interval, but documentation and usage in the codebase explains it differently. Documentation refers to it as if it was `normalizeToInterval` function and of course it seems it's used like that. This is very confusing. The implementation of the two is pretty similar too.

### DIFF
--- a/ui/types/utils/date.d.ts
+++ b/ui/types/utils/date.d.ts
@@ -1,18 +1,9 @@
-export interface BuildDateOptions {
+export interface DateOptions {
   milliseconds?: number;
   seconds?: number;
   minutes?: number;
   hours?: number;
   date?: number;
-  month?: number;
-  year?: number;
-}
-export interface ModifyDateOptions {
-  milliseconds?: number;
-  seconds?: number;
-  minutes?: number;
-  hours?: number;
-  days?: number;
   month?: number;
   year?: number;
 }
@@ -29,23 +20,23 @@ export type DateUnitOptions = "second" | "minute" | "hour" | "day" | "month" | "
 export namespace date {
   function isValid(date: number | string): boolean;
   function extractDate(str: string, mask: string, locale?: DateLocale): Date;
-  function buildDate(options: BuildDateOptions, utc?: boolean): string;
+  function buildDate(options: DateOptions, utc?: boolean): string;
   function getDayOfWeek(date: Date | number | string): number;
   function getWeekOfYear(date: Date | number | string): number;
   function isBetweenDates(date: Date | number | string, from: Date | number | string, to: Date | number | string, opts?: { inclusiveFrom: boolean; inclusiveTo: boolean }): boolean;
-  function addToDate(date: Date | number | string, options: ModifyDateOptions): Date;
-  function subtractFromDate(date: Date | number | string, options: ModifyDateOptions): Date;
-  function adjustDate(date: Date | number | string, options: ModifyDateOptions, utc?: boolean): Date;
+  function addToDate(date: Date | number | string, options: DateOptions): Date;
+  function subtractFromDate(date: Date | number | string, options: DateOptions): Date;
+  function adjustDate(date: Date | number | string, options: DateOptions, utc?: boolean): Date;
   function startOfDate(date: Date | number | string, option: DateUnitOptions): Date;
   function endOfDate(date: Date | number | string, option: DateUnitOptions): Date;
   function getMaxDate(date: Date | number | string, ...args: (Date | number | string)[]): Date;
   function getMinDate(date: Date | number | string, ...args: (Date | number | string)[]): Date;
-  function getDateDiff(date: Date | number | string, subtract: Date | number | string, unit?: string): Date;
+  function getDateDiff(date: Date | number | string, subtract: Date | number | string, unit?: DateUnitOptions): Date;
   function getDayOfYear(date: Date | number | string): number;
   function inferDateFormat(date: Date | number | string): "date" | "number" | "string";
   function getDateBetween(date: Date | number | string, min?: Date | number | string, max?: Date | number | string): Date;
   function isSameDate(date: Date | number | string, date2: Date | number | string, unit?: DateUnitOptions): boolean;
   function daysInMonth(date: Date | number | string): number;
-  function formatDate(date: Date | number | string | undefined, format: string, locale?: DateLocale, __forcedYear?: number): string;
+  function formatDate(date: Date | number | string | undefined, format?: string, locale?: DateLocale, __forcedYear?: number): string;
   function clone<D extends Date | number | string>(date: D): D;
 }
--- a/ui/types/utils/dom.d.ts
+++ b/ui/types/utils/dom.d.ts
@@ -10,5 +10,6 @@ export namespace dom {
   function width(el: Element): number;
   function css(el: Element, css: Partial<CSSStyleDeclaration>): void;
   function cssBatch(elements: Element[], css: Partial<CSSStyleDeclaration>): void;
-  function ready(fn: Function): any;
+  function ready<F extends (...args: any[]) => any>(fn: F): ReturnType<F>;
+  function ready(fn?: any): void;
 }
--- a/ui/types/utils/event.d.ts
+++ b/ui/types/utils/event.d.ts
@@ -1,7 +1,14 @@
+// Allow using `passive` and `notPassive` with `removeEventListener`
+// See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#Matching_event_listeners_for_removal
+// See https://github.com/quasarframework/quasar/pull/5729#issuecomment-559588257
+interface RemoveEventListenerFix {
+  capture: undefined;
+}
+
 export interface ListenOpts {
   hasPassive: boolean;
-  passive: undefined | { passive: true };
-  notPassive: undefined | { passive: false };
+  passive: undefined | ({ passive: true } & RemoveEventListenerFix);
+  notPassive: undefined | ({ passive: false } & RemoveEventListenerFix);
   passiveCapture: true | { passive: true; capture: true };
   notPassiveCapture: true | { passive: false; capture: true };
 }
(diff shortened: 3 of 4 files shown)


GordonBlahut:
@IlCallo the proposed event typings look fine to me. They look like a more accurate representation of the actual implementation.

IlCallo:
Could you do a small check on your code where you're using those typings and see if anything breaks?

Just create a `shims.d.ts` in `src` folder with this content and it should temporarily patch that interface in your project :)

```ts
import 'quasar';

declare module 'quasar' {
  interface ListenOpts {
    hasPassive: boolean;
    passive: undefined | { passive: true };
    notPassive: undefined | { passive: false };
    passiveCapture: true | { passive: true; capture: true };
    notPassiveCapture: true | { passive: false; capture: true };
  }
}
```

Unluckily it's not possible to augment constants type (or I haven't yet found the way), so if you are using it via `listenOpts` (highly probable), you should cast it like `(event.listenOpts as ListenOpts).<the-option-you-want>`.

GordonBlahut:
> if you are using it via `listenOpts` (highly probable), you should cast it like `(event.listenOpts as ListenOpts).<the-option-you-want>`.

The cast is needed only to see the shim, right? Adding the event namespace to the shim as well should avoid the need to cast.

I tried out the proposed change and didn't run into any issues but my usage is not exhaustive.

IlCallo:
I think adding the namespace won't work unluckily, it's a bit more complicated...
I tested with casting and it worked in a simple test for me

GordonBlahut:
I found a minor issue, but it's not really the fault of your proposed types.

`ListenOpts.passive` and `ListenOpts.notPassive` are not compatible with `removeEventListener`'s `boolean | EventListenerOptions | undefined` because they have no properties in common with `EventListenerOptions`. 

They work fine with `addEventListener` because `AddEventListenerOptions` has a `passive` property.

```typescript
// no problem
window.addEventListener("scroll", this.onScroll, (event.listenOpts as ListenOpts).passive)

// no overload matches this call
window.removeEventListener("scroll", this.onScroll, (event.listenOpts as ListenOpts).passive) 
```

In `removeEventListener`, only the `capture` property really matters, so it doesn't really make sense to use `listenOpts.passive` or `listenOpts.notPassive` in `removeEventListener`.

e.g. one could use instead:
```typescript
window.addEventListener("scroll", this.onScroll, (event.listenOpts as ListenOpts).passive)
window.removeEventListener("scroll", this.onScroll, false) 
```
Though this does lead to some awkwardness between adding and removing an event handler due to the non-matching options.

[MDN also notes this about options for `removeEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#Matching_event_listeners_for_removal):
> It's worth noting that some browser releases have been inconsistent on this, and unless you have specific reasons otherwise, it's probably wise to use the same values used for the call to `addEventListener()` when calling `removeEventListener()`.

So it's not really "wrong" to use something like `{ passive: true }` in `removeEventListener`, but it's not really supported by TypeScript either.

I think your definitions are an improvement but I also think someone should be able to pass the same options object used in `addEventListener` to `removeEventListener`. 

I don't really have a solution for that.

Quasar does contain internal `removeEventListener` calls with `listenOpts.passive`, not that this is affected by the typing.

IlCallo:
Related https://github.com/microsoft/TypeScript/issues/32912
I don't see many options except a rollback or a global-level patch (not really a good idea to me).
I'll think a solution about this in a few days

rstoenescu:
Guys, please ping me when a consensus has been reached on this and it needs merging.
If there is any breaking change, drop me a message about it. Thanks for the hard work!

IlCallo:
@GordonBlahut I think I found a workaround.

See if this new shim work for the scenario you brough up, I did a small test myself and it seems to work

```ts
import 'quasar';

declare module 'quasar' {
  // Allow using `passive` and `notPassive` with `removeEventListener`
  // See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#Matching_event_listeners_for_removal
  // See https://github.com/quasarframework/quasar/pull/5729#issuecomment-559588257
  interface RemoveEventListenerFix {
    capture: undefined;
  }
  interface ListenOpts {
    hasPassive: boolean;
    passive: undefined | ({ passive: true } & RemoveEventListenerFix);
    notPassive: undefined | ({ passive: false } & RemoveEventListenerFix);
    passiveCapture: true | { passive: true; capture: true };
    notPassiveCapture: true | { passive: false; capture: true };
  }
}
```

GordonBlahut:
@IlCallo That seems to work without any type issues.

Just curious, is there a benefit to the additional interface instead of just directly adding the undefined capture directly?
```typescript
declare module 'quasar' {
  // Allow using `passive` and `notPassive` with `removeEventListener`
  // See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#Matching_event_listeners_for_removal
  // See https://github.com/quasarframework/quasar/pull/5729#issuecomment-559588257
  interface ListenOpts {
    hasPassive: boolean;
    passive: undefined | { passive: true; capture: undefined };
    notPassive: undefined | { passive: false; capture: undefined };
    passiveCapture: true | { passive: true; capture: true };
    notPassiveCapture: true | { passive: false; capture: true };
  }
}
```
and/or `undefined |` instead of `?:`
```typescript
declare module 'quasar' {
  // Allow using `passive` and `notPassive` with `removeEventListener`
  // See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#Matching_event_listeners_for_removal
  // See https://github.com/quasarframework/quasar/pull/5729#issuecomment-559588257
  interface ListenOpts {
    hasPassive: boolean;
    passive?: { passive: true; capture: undefined };
    notPassive?: { passive: false; capture: undefined };
    passiveCapture: true | { passive: true; capture: true };
    notPassiveCapture: true | { passive: false; capture: true };
  }
}
```

IlCallo:
For the first, code clarity :)
The interface is an excuse to semantically group the code needed for the workaround and the name is self-explanatory for why that strange code is needed (you'd never define an object with only an undefined property). It also provide an immediate visual hook to which I can add comments with additional context.

Without the interface it would be more difficult to understand which part of the code is the actual fix and to which lines comments refer. Next persone checking it would spend a lot of time just to understand those two things or just blindly remove the fix.

For the second, the `?:` notation usually means that a property is optional.
In this case the property is not optional, it is an object or is not set.
For the compiler there is no difference, but for the programmer this is a subtle (but important) distinction: it means that `undefined` carries an inner value and invites the reader to dig further to understand why it has been put like this. Like you did :P
An explanation comment would be better, but in this case it is a JS-related behaviour, not TS-related, so the comment should probably go near the JS code.

Also, it keeps it visually similar to other properties

GordonBlahut:
Thanks for the clarification. Makes sense, I'll have to keep that in mind when I create my own types.

mesqueeb:
@IlCallo I'm currently doing this as workaround:
```ts
(getDateDiff(date1, date2, unit) as unknown) as number
```
😆

rstoenescu:
Hi,

I see this PR is in "Draft" state. Is it ready for merging though?

IlCallo:
It's in draft state because it suggests breaking changes to the implementation under `other information` paragraph, I need your feedback on those.

If you could tell me what do you think about the changes I propose, I can implement the one you likes and discard the one you don't like (I only put the typings updates for now, not the implementation changes) and remove the draft flag

IlCallo:
If you prefer I can break up this PR in smaller ones, but being a bunch of possible breaking changes I think it's better to commit them in a single batch.

PR #5730, opened 2019-11-28
feat(docs): filter deprecated app extensions

hawkeye64:
The code added, filters discovered app extensions to make sure they are not deprecated before showing them to the user.


rstoenescu:
Cool!

PR #5735, opened 2019-11-28
feat(typescript): enhance types 3

IlCallo:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

**All typings here are purely additive and descriptive**, they don't introduce breaking changes, it should be safe to merge the PR whenever you like.

_In this batch_
These types are meant to be used later on as building blocks for more complex types (eg. `QuasarConfiguration`) and to provide a better access to language and iconset, both while testing or when accessing them via global Quasar object (`$q.lang`, `$q.iconSet`).
They are the "brute force" ones, because I had to get them by copy/pasting stuff around or manually checking plain JS objects.
If the codebase switch to use TS, most of these would be inferred automatically or defined once and applied everywhere to enforce consistency, but right now they **may** be a pain in the ass to maintain.
Seeing they changed really rarely in the last year and that they can be easily fixed if someone create an issue, it's probably a burden which can be taken.

_Details_
- `extras` options could be derived at build time, but they are so few they aren't really worth an automation script.
- `StringDictionary` is a small helper to keep the code a bit more concise and avoid adding tons of `[propertyName]: string`.
- `QuasarLanguageCodes` contains all currently available languages. If keeping it in sync turns out to be a problem, it could be generated at build time from files names into `ui/lang`. 
- `QVueGlobals` has been enhanced to provide intellisense both when accessing language/iconset and when setting it.
- `QuasarPluginOptions` has more precise typings now. This change proves useful when writing unit tests with TS.


rstoenescu:
Should this be on the /extras package (`@quasar/extras`) directly instead?

IlCallo:
I thought so, but currently all types are in `ui` folder, in the monorepo, so I guessed that all typings should go there.
Also, this is a WIP, there are other types I have to commit on this PR which are related to `@quasar/extras`.
I'll ping you for review once it's complete and if needed I can just move everything into the `/extras` folder

ui/types/extras/icon-set.d.ts
```diff
@@ -0,0 +1,92 @@
...
+type QuasarIconSetType = "positive" | "negative" | "info" | "warning";
+type QuasarIconSetArrow = "up" | "right" | "down" | "left" | "dropdown";
+type QuasarIconSetChevron = "left" | "right";
+type QuasarIconSetColorPicker = "spectrum" | "tune" | "palette";
+type QuasarIconSetPullToRefresh = "icon";
+type QuasarIconSetCarousel = "left" | "right" | "navigationIcon" | "thumbnails";
+type QuasarIconSetChip = "remove" | "selected";
+type QuasarIconSetDatetime = "arrowLeft" | "arrowRight" | "now" | "today";
+type QuasarIconSetEditor =
+  | "bold"
```
outofmemoryagain: Is the `|` before the first option going to be a problem?  It should probably be removed.

ui/types/extras/icon-set.d.ts
```diff
@@ -0,0 +1,92 @@
...
+type QuasarIconSetExpansionItem = "icon" | "denseIcon";
+type QuasarIconSetFab = "icon" | "activeIcon";
+type QuasarIconSetField = "clear" | "error";
+type QuasarIconSetPagination = "first" | "prev" | "next" | "last";
+type QuasarIconSetRating = "icon";
+type QuasarIconSetStepper = "done" | "active" | "error";
+type QuasarIconSetTabs = "left" | "right" | "up" | "down";
+type QuasarIconSetTable = "arrowUp" | "warning" | "prevPage" | "nextPage";
+type QuasarIconSetTree = "icon";
+type QuasarIconSetUploader =
```
outofmemoryagain: Same as above.

ui/types/extras/icon-set.d.ts
IlCallo: No, it's only a style choice and it will be ignored during TS parsing. Prettier uses this notation for readability, keeping all options aligned

ui/types/lang.d.ts
```diff
@@ -0,0 +1,163 @@
+import { StringDictionary } from "./ts-helpers";
+
+export type QuasarLanguageCodes =
+  | "ar"
```
rstoenescu: Can we do something here to avoid manual updating?

rstoenescu:
@IlCallo Merged, but can we do something with QuasarLanguageCodes so that it automatically gets updated rather than having to manually do it, pls? This will certainly be missed out when people PR new languages.

IlCallo:
Yes, it can be automated when building types, I'll give it a try

PR #5750, opened 2019-11-29
fix(app/cordova): fixed possible race condition

yusufkandemir:

- Bugfix

- No

- It's been tested on a Cordova (iOS, Android) app

**Other information:**
Previously it was async and wasn't awaited, so a possible race condition could produce a bug, solved by making it sync since the containing function is sync.
[Documentation for ensureDirSync](https://github.com/jprichardson/node-fs-extra/blob/2b97fe3e502ab5d5abd92f19d588bd1fc113c3f2/docs/ensureDir-sync.md)

### DIFF
--- a/app/lib/mode/mode-cordova.js
+++ b/app/lib/mode/mode-cordova.js
@@ -85,7 +85,7 @@ class Mode {
   }
 
   addPlatform (target) {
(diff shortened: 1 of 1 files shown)


rstoenescu:
Nice find!

Issue #5766, opened 2019-12-01
[docs] sdkmanager --licenses


mesqueeb:
https://quasar.dev/quasar-cli/developing-capacitor-apps/preparation

here it says:
> Open the terminal and go to the folder where the SDK was installed, in tools/bin and call `sdkmanager --licenses`.

However, after installing Android Studio on macOS:
1. ~~I'm not sure where this "SDK" was installed. I cannot find a folder called `tools/bin` on my mac.~~

→　found in the hint below

2. also my terminal says: `fish: Unknown command sdkmanager`
also on bash:

```
/Users/lucaban/Library/Android/sdk/tools/bin
bash-3.2$ sdkmanager --licenses
bash: sdkmanager: command not found
```

Some extra information on this in the docs would be greatly appreciated!

PR #5767, opened 2019-12-01
hint(docs): explain to read further down for more info

mesqueeb:
the other day I got stuck at the top, because I couldn't find the SDK folder.
This is however, explained further down in a hint in the docs.

A small hint to look further down in the docs would help a lot of people I think!

Partially related to #5766


rstoenescu:
Hi,

Thanks for your awesome contributions!
Will leave this one out unfortunately. It's not clear where "down below" is...

Issue #5828, opened 2019-12-08
"rtl:ignore" comment is ignored when using RTL and SASS


behnoodk:
When using SASS and with RTL enabled. the following class should be left intact:

``` css
.my-class{
  text-align: left; /* rtl:ignore */
}
```
But it gets converted by postcss anyway. This problem does not happen when using Stylus.

tests on:
```
Pkg quasar........ v1.5.5
Pkg @quasar/app... v1.3.3
```

pdanpdan:
I did some research.

SASS removes comments by default, so they don't get processed by postprocessor.
If you want the comments to be kept you have to use this kind of comments (also take care to put the semicolon after the comment):

```sass
.my-class{
  text-align: left #{"/* rtl:ignore */"};
}
```

I'll make the same changes in the main repo now that I found out.

rstoenescu:
Hi,

Thanks for reporting!
Fix will be available in "quasar" v1.5.8.

behnoodk:
@pdanpdan  Thank you. This special comment syntax takes care of the issue.

rstoenescu:
Not ready yet. Build is still broken. Working on it.

rstoenescu:
Will require "quasar" v1.5.8 AND "@quasar/app" v1.3.6.

Issue #5826, opened 2019-12-07
Wrong QSpinner animation when using SASS and RTL


behnoodk:
Today I switched my Quasar V1 app from Stylus to SASS and noticed this problem. This is exactly the same issue as #3736. This can easily be reproduced when a new SASS project is in RTL mode. 

The only difference from #3736 is that the problem is both in `dev` and `build` versions.

tested on quasar `v1.5.5` and `@quasar/app v1.3.3`

rstoenescu:
Hi,

Thanks for reporting!
Fix will be available in "quasar" v1.5.8.

rstoenescu:
Will require "quasar" v1.5.8 AND "@quasar/app" v1.3.6.

PR #5834, opened 2019-12-09
fix(SASS rtl): SASS removes comments so we need to use special comments #5826, #5828

pdanpdan:


rstoenescu:
Does not fixes the build

PR #5850, opened 2019-12-10
feat(colors): add new brightness function

hawkeye64:
Determines brightness of passed in color. Will be usefull for customized dark modes.
Algorithm taken from https://www.w3.org/TR/AERT/#color-contrast


rstoenescu:
Excellent!

Issue #5430, opened 2019-10-31
Ipad 13.1 considered as mac desktop


matthieuowlie:
**Describe the bug**
When using an Ipad in 13.1 version, the Platform plugin of quasar considered the Ipad as a mac desktop

**To Reproduce**
1. Use Platform from 'quasar'
2. Use Platform.is
3 Platform.is.mac is true
4 Platform.is.desktop is true

**Expected behavior**
Platform.is.mac should be false
Platform.is.desktop should be false
Platform.is.ipad should be true

**Screenshots**
![platform](https://user-images.githubusercontent.com/56551343/67949654-5d6e3980-fbe8-11e9-99af-d789b0d65699.png)

**Platform (please complete the following information):**
Node: v10.16.3
NPM: 6.9.0
Browsers: Safari
iOS: 13.1
Platform: iPad

pdanpdan:
Uncheck the option to always request desktop site in iPad.

iperiago:
While @pdanpdan comment is right, this has some implications though : when you use viewport units in your css (like height: 100vh for instance), it's a well known issue that iOS doesn't treat this correctly. If you rely on Quasar Platform to detect iOS and react accordingly, this will create a bug in this particular case as iOS won't be detected. My solution was to check if both Platform.is.mac and Platform.has.touch are true - it's obviously not perfect though.

mrvictorn:
@pdanpdan 

> Uncheck the option to always request desktop site in iPad.
--
This does not work for iPad PWA. No way to uncheck it.
Also asking users to uncheck something is not best way to treat them,
right @rstoenescu ?

PR #5855, opened 2019-12-10
Just for ref - feat(Platform): Improve detection of ios when emulated or used with request desktop site

pdanpdan:
ref #5430

### DIFF
--- a/ui/src/plugins/Platform.js
+++ b/ui/src/plugins/Platform.js
@@ -8,6 +8,10 @@ export const isSSR = typeof window === 'undefined'
 export let fromSSR = false
 export let onSSR = isSSR
 
+function hasTouch () {
+  return 'ontouchstart' in window || window.navigator.maxTouchPoints > 0
+}
+
 function getMatch (userAgent, platformMatch) {
   const match = /(edge|edga|edgios)\/([\w.]+)/.exec(userAgent) ||
     /(opr)[\/]([\w.]+)/.exec(userAgent) ||
@@ -180,6 +184,20 @@ function getPlatform (userAgent) {
   browser.platform = matched.platform
 
   if (isSSR === false) {
+    if (browser.mac === true && hasTouch() === true) {
+      browser.mac = void 0
+      browser.desktop = void 0
+      browser.iosDesktop = true
+      browser.mobile = true
+      browser.ios = true
+      browser.platform = Math.min(window.innerHeight, window.innerWidth) > 414 ? 'ipad' : 'iphone'
+      browser[browser.platform] = true
+    }
+
+    if (browser.ios === true && window.navigator.vendor.toLowerCase().indexOf('apple') === -1) {
+      browser.iosEmulated = true
+    }
+
     if (userAgent.indexOf('electron') > -1) {
       browser.electron = true
     }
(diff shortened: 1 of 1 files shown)


ibrainventures:
hi @pdanpdan , hi @rstoenescu 
i am working with this commit since some days, and the ipados / render as desktop with those additions are acknowledge. May this will go into the master?

rstoenescu:
@ibrainventures chose a different path, but it has the same effect. you'll be happy when you upgrade to v1.7.4 :)

ibrainventures:
hi @pdanpdan  hi @rstoenescu 
i saw today the new , modified version and tested on quasar.dev all features. ! Thanks for all yourwork.
I thought - and hoped - that the qmenu etc. deposition with zoomed / viewport on ios 13+ / ipados - requested as desktop - with this differentiation works .. but it will not work ..

have you got a idea, why this viewport regulation / qmenu positioning - which is according to the sourcecode respected - not working on request as desktop ipados 13 ?

i tested the platform detection on quasar.dev and all vars are showing ios , webkit, ipad - with the request as desktop mode ...

ibrainventures:
may quasar.dev is not completly updated ? .. i have no debug-ios-ipad option actually, so i cant view into the dom about those two vp css root vars ...

pdanpdan:
We just tested, the QMenu/QTooltip is positioned correctly when zoomed.
Can you explain in more details what seems to be wrong?

ibrainventures:
hi @pdanpdan 
ipad ios 13 , request as desktop, qmenu , tooltip, etc. are depositioned - if a little zoomed.

the plattform detection helps maybe - there are some little differences. 
ipad-not-as-dekstop
ipad-as-desktop
depos1
depos2

ibrainventures:
platform detection : some differences in the ranking  - (position of the characteristic)

ibrainventures:
same ipad , same minute, same webside (quasar.dev)

pdanpdan:
We'll check, thanks for reporting.

ibrainventures:
but the icons looks great !! and all other things .. sorry for beeing sometimes the party-crasher ...

rstoenescu:
Not at all! Please do comment. It’s helping Quasar become insanely good ;)

ibrainventures:
good morning @pdanpdan @rstoenescu 

some good and some bad news

i freshly gitcloned the quasar git and started locally a quasar dev inside the doc dir .. there the viewport adaption works fine, and qtooltip and qmenu are on the right place with a zoomed request as desktop.

i updated our webapp with a heavy patched 1.7.4 and there the viewport adaption also works fine.

Now the bad news:

But still on quasar.dev a non good working 1.7.4 .. So i think your quasar.dev deploy is half-failured or some cdn s or too long cache-times for .js are interferring the site actually.

May a complete cleanup on the webspace and a fresh deploy can  solve this "issue" ...

Some pics from the local running /docs :

good1


good2

rstoenescu:
@ibrainventures  Found the culprit. You're safe with the mobile apps. It's only a SSR issue (which I fixed in next v1.7.5 - to be released).

ibrainventures:
@rstoenescu perfect !! have a nice weekend.

rstoenescu:
u too!

ibrainventures:
only for completness

the 1.8.3 deployed quasar.dev is working fine with the viewport zoom. 
All menu dropdowns are on the right place.

PR #5861, opened 2019-12-11
feat(docs) Add warning about missing `.cancel()` when debouncing method the "wrong way"

panstromek:
Warn against doing this:
```js
methods: {
  myMethod: debounce(function(){},500)
},
beforeDestroy() {
  this.myMethod.cancel() // This won't work!!
}
```
This has bitten me today ;)


rstoenescu:
Thanks for this!

Issue #5711, opened 2019-11-27
[typescript] Storage.getItem return type should not be any


Davilink:
**Is your feature request related to a problem? Please describe.**
the return type of Storage.getItem should not be 'any', because it hide the 'null' possibility.

**Describe the solution you'd like**
```ts
type ReturnType = Date | RegExp | number | boolean | string | object;

interface SessionStorage {
 // ...
 /**
     * Get a storage item value
     * @param key Entry key
     * @returns Storage item value
     */
    getItem (key : string): ReturnType | null;
// ...
}
```

**Additional context**
```ts
const jwt = storage.getItem(SessionStorageKey.Token);
```
instead of having this

![image](https://user-images.githubusercontent.com/1498932/69739387-a3de8780-1105-11ea-9b3a-a11aa0a4853a.png)

i would get 

![image](https://user-images.githubusercontent.com/1498932/69740441-64b13600-1107-11ea-9d73-48a0cffd1bef.png)

which clearly indicate that null is a possibility.

panstromek:
@IlCallo you might be interested, not sure if you noticed already ;)

IlCallo:
`this change is not as straightforward as it seems`

Can confirm this, current system doesn't allow to inject arbitrary TS types, I'm thinking a solution capable of covering this need

rstoenescu:
Fix will be available in "quasar" v1.6.2.

PR #5937, opened 2019-12-23
feat(typescript): allow usage of custom TS types for props and methods (fix: #5711)

IlCallo:

- [x] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
As per https://github.com/quasarframework/quasar/issues/5711
Type build system has been extended to allow providing a TS custom interface for a prop (of type Object, Any, Function, MultipleTypes) or a method.

Custom types must be exported by `ui/types/api.d.ts`. Every custom type should be placed inside a file into `ui/types/api/` folder with the same name of the feature they are related to (eg. `ui/types/api/cookies.d.ts`).

The PR includes an example of a prop and a method with custom TS type, taken by two fix requests.

@webnoob can you check if Cookies `get()` now works for you as we discussed last month on Discord?
@Davilink can you check if Storage `getItem()` and `getIndex` now works for you as you asked?

You can do so by
- cloning my quasar fork in a local folder
- changing branch to `typings-7`
- moving into `ui` folder
- running `yarn install`
- running `yarn build`
- moving to your project folder
- run `yarn add file:<relative-path-to-quasar-ui-folder>` (eg. `yarn add file:../quasar/ui`)


Davilink:
Yeah, i can see the different return type, but will it be possible to have something more like : 

// index.d.ts
image

// web-storage.d.ts
image

image
image

IlCallo:
@Davilink better?

Davilink:
Hi, yes and no. 
Yes: 
we can cast the return type by doing:
image

No:
The intellisense doesn't show any thing about the generic, so if you don't know that you can do it, you will not think about using it.
image

and even if i go into the definition i cannot see that it take a generic
image

the default return type should not be string, but any.
image

IlCallo:
The problems you see are due to the fact that to have a simpler types build system we must declare methods signature as types and apply a function type to a property (instead of declaring it as a function from the start). Until core is migrated to TS, I think this is a good compromise.

If you jump to the type declaration (so first you jump to `getItem`, then `WebStorageGetItemMethodType`) you'll see the generic type, you just have to do 2 jumps instead of 1.

Actually, checking again `decode` function, I think my previous implementation is right.
When something gets out of `getItem`, it isn't of type `any`, it is one between the types I put into `WebStorageGetMethodReturnType`.

Why aren't you using it with a type assertion, as it should be in this case?
Like `const value = LocalStorage.getItem('test') as boolean`

Generics here aren't problably the best way, now I remember why I didn't use them in the first place

Davilink:
We could remove the default type 
```ts
export type WebStorageGetIndexMethodType = <
  T extends WebStorageGetMethodReturnType = string
>(
  index: number
) => T | null;
```

to 
```ts
export type WebStorageGetIndexMethodType = <
  T extends WebStorageGetMethodReturnType
>(
  index: number
) => T | null;
```

> Why aren't you using it with a type assertion, as it should be in this case?
Like const value = LocalStorage.getItem('test') as boolean

>Generics here aren't problably the best way, now I remember why I didn't use them in the first place.
Using plain return type and asserting it's type also result in a cleaner generated type.

the reason is because the type `any` result encapsulate the nullable type, i wanted a way to get `T | null`

because the majority of the people will do the type assertion just like you did : 
```ts
const value = LocalStorage.getItem('test') as boolean
```
but in fact it should be : 
```ts
const value = LocalStorage.getItem('test') as boolean | null
```

but we could argue that we should use the hasItem before, but because the LocalStorage and SessionsStorage are shared between multiple context (page/worker), it is possible that the Storage could be emptied or that property could be clear in other context. so the `hasItem` return true and when we reach the `getItem` it return null. It may be a edge case but that the reason behind why i wanted to indicate the nullability of the return type of `getItem`

IlCallo:
```ts
export type WebStorageGetIndexMethodType = <
  T extends WebStorageGetMethodReturnType = WebStorageGetMethodReturnType
>(
  index: number
) => T | null;
```

Putting the return type as default type is probably the best way. I don't want to force devs to specify the generic in that moment, I'll check if it works in a few days

rstoenescu:
This is not needed because current type generating algorithm covers multiple types on "returns" too.
Fix reference:
1. https://github.com/quasarframework/quasar/commit/2968f35af5801e9df0437f97e65a4f2426737ddf
2. https://github.com/quasarframework/quasar/commit/32ddfd5b4a74163b1010b0ec70ae8a604fd91ce1

IlCallo:
Pay attention, this PR is meant to allow to use generics for a whole method (check "CookiesGetMethodType" example), which cover cases not currently feasible using the current system

rstoenescu:
@IlCallo can you explain in-depth what these cases are, pls? With the current changes (not ones from this PR), the types are correct.

Davilink:
@rstoenescu the type of the a `getItem` call will result as a mixed type of  `Date | RegExp | Number | boolean | Function | object | string | null` and one of my fear is the people will do:

```ts
const value = LocalStorage.getItem('test') as boolean
```

but in fact it should be :

```ts
const value = LocalStorage.getItem('test') as boolean | null
```

IlCallo:
Types are correct, but they could be more dev-friendly, precise and concise using a generic type.
They are preferred over type assertions (the `as Something` notation which is usually seen as a type cast) and TS devs expects to have them when dynamic types are used.

I don't know how much do you know about TS, but just to be sure [here](https://www.typescriptlang.org/docs/handbook/generics.html) are explained what generics are in TS and in which cases you could use them.

Simple example with WebStorage (but same goes for `Cookies.get`):
```ts
// With current system
const value1 = LocalStorage.getItem('test');
// 'value1' is of type 'Date | RegExp | number | boolean | Function | object | string | null'
const value2 = LocalStorage.getItem('test') as boolean | null;
// 'value2' is of type 'boolean | null'

// Using generics as shown in this PR
const value1 = LocalStorage.getItem('test');
// 'value1' is of type 'Date | RegExp | number | boolean | Function | object | string | null'
const value2 = LocalStorage.getItem<boolean>('test');
// 'value2' is of type 'boolean | null'
```

The latter is preferred because it's less error prone: you can define the return type, but the `| null` part is always automatically added.

Generics get more useful when complexity increases.
I tried to find a more complex scenario into Quasar realm but apparently I don't know enough about the components/directives/plugins yet.

From my experience (mostly Angular based), I found generics useful when dealing with dynamic components, dialogs/forms creation, functions/callbacks manipulations, etc. 

For example, [`Dialog`](https://quasar.dev/quasar-plugins/dialog) plugin can be modeled like this in case a shortcut to the underlying custom component should be added:

```ts
interface DialogComponent extends Vue {
  show() => void;
  hide() => void;
  // Other mandatory  methods as needed
}

// Typesafe Dialog 'create' version
interface QDialogOptions<C extends DialogComponent | undefined = undefined> {
  // Throw error when mandatory methods are not defined
  component: C;
  // Other options
}
interface Dialog {
  // Here we are using the generic type 
  create<C extends DialogComponent>(opts : QDialogOptions<C>): DialogChainObject & { component: C }
}

const component = {
  someExtraMethod() => string;
  // Dialog component mandatory methods implementation
}

// 'C' generic type gets automatically inferred from the provided value 
const dialog = Dialog.create({ component });
// 'component' here will have the same type of our custom component
dialog.component.someExtraMethod();
```

Putting it as a methaphore: supporting TS without using/supporting generics is like having a toolkit with a manual saw and a chainsaw, but using only the manual saw even when you need to cut down a tree with an half-meter diameter trunk.
You can do it, it's just harder, unsafer etc etc.

---

It's always possible to leave this PR unmerged and check it again when a more complex use case arises, the work has already been done anyway.

PR #5938, opened 2019-12-23
feat(docs): multi-image slide with QCarousel

hawkeye64:
This is an example of how to put more than 1 item in a QCarouselSlide. In this case, multiple QImg's are used.


hawkeye64:
image

hawkeye64:
Merry Christmas! :santa:

PR #5946, opened 2019-12-25
fix(ui): fix attributes not being passed on correctly

hawkeye64:
Be explicit in $attrs inheritance. Vue does not always get it right.
https://cdn.discordapp.com/attachments/596275881907978240/659171134062985226/unknown.png
Also, make sure user doesn't change type as per component ($attrs needs to be first)
As per: https://vuejs.org/v2/guide/components-props.html#Disabling-Attribute-Inheritance


hawkeye64:
Fixes #5963

PR #6015, opened 2020-01-06
feat(api): find as many api errors as possible before exiting build

hawkeye64:
In this regard, it's easier to fix many than one at a time and restarting build each time


rstoenescu:
Gonna make some small tweaks to it. Including the array validation.

hawkeye64:
Nice, thanks for the explanation. :heart:

Issue #5987, opened 2020-01-02
`quasar build -m pwa` failing with latest version


codingfriend1:
After upgrading to the latest version of quasar from these previous versions:

@quasar/extras: 1.3.1 → 1.3.3
 quasar: 1.1.0 → 1.5.11
 @quasar/app: 1.0.6 → 1.4.3
 @quasar/quasar-app-extension-icon-genie: 1.0.2 → 1.1.3

The build command for generating a pwa fails.

quasar build -m pwa

 Build mode........ pwa
 Pkg quasar........ v1.5.11
 Pkg @quasar/app... v1.4.3
 Debugging......... no
 Publishing........ no

 app:extension Running "@quasar/icon-genie" Quasar App Extension... +0ms
 app:quasar-conf Reading quasar.conf.js +14ms
 app:workbox [GenerateSW] Will generate a service-worker file. Ignoring your custom written one. +815ms
 app:webpack Extending PWA Webpack config +22ms
 app:artifacts Cleaned build artifact: "/Users/jonpaulmiles/Documents/Work/prowlers-and-paragons/dist/pwa" +5ms
 app:generator Generating Webpack entry point +0ms
 app:build Extension(@quasar/icon-genie): Running beforeBuild hook... +4ms
 app:build Building... +23ms

 PWA █████████████████████████ [100%] in ~13s

Version: webpack 4.41.3
Time: 12889ms
Built at: 01/02/2020 2:16:15 PM
                                                  Asset      Size  Chunks               Chunk Names
            fonts/KFOkCnqEu92Fr1MmgVxIIzQ.5cb7edfc.woff  19.9 KiB                       
           fonts/KFOlCnqEu92Fr1MmEU9fBBc-.87284894.woff    20 KiB                       
           fonts/KFOlCnqEu92Fr1MmSU5fBBc-.b00849e0.woff  19.9 KiB                       
           fonts/KFOlCnqEu92Fr1MmWUlfBBc-.adcde98f.woff  19.9 KiB                       
           fonts/KFOlCnqEu92Fr1MmYUtfBBc-.bb1e4dc6.woff  19.9 KiB                       
               fonts/KFOmCnqEu92Fr1Mu4mxM.60fa3c06.woff  19.8 KiB                       
                      fonts/fa-brands-400.c5e0f14f.woff  86.4 KiB                       
                     fonts/fa-brands-400.cccc9d29.woff2  73.6 KiB                       
[...]

PR #6051, opened 2020-01-10
Fix: #5987 - webpack conflict for [contenthash]

jeancaffou:
> This appears to be a conflict with webpack 4.3 which introduced a [contenthash] variable of its own: https://github.com/webpack/webpack/releases/tag/v4.3.0

Source: https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/763


jeancaffou:
This is a workaround and not a good solution.

Issue #6080, opened 2020-01-12
[docs] QTable (Synchronizing with Server) sorting issue


hawkeye64:
When you sort by description it does not work. Then if you sort by something else, it takes 2 tries after sorting by desc.

![QTable-binary-state-sort--issue](https://user-images.githubusercontent.com/10262924/72219469-b66d1e80-3503-11ea-9dac-3c6849827e09.gif)

pdanpdan:
Fixed on the docs site.

PR #6086, opened 2020-01-13
docs(QTable): Fix sort in server side simulation example #6080

pdanpdan:


rstoenescu:
@pdanpdan Please check all table examples if they work correctly (on all of their features).

pdanpdan:
The rest of the examples for QTable look good to me.

Issue #2067, opened 2018-05-21
Feature Request: Option to get rid of Checkbox intermediate state or change toggle order


KnowZero:
Quasar: 0.15-0.16

It would be nice to have the option of completely removing the intermediate state in checkboxes or maybe set the toggle order of states.

I know I can set the 'data' to the false-value or true-value but that isn't very convenient when working with dynamic forms.

Even if I set something like indeterminate-value="none" and don't specify a false-value, it still picks the indeterminate-value. (Even though it should technically pick the false value since that is null and not the intermediate value)

I would be able to get around this if I just change the indeterminate-icon="check_box_outline_blank", but then the user has to click twice to turn it on since I can't control the toggle order.

So the only way I can achieve close to the needed functionality is to flip everything around:
false-value="yes" indeterminate-icon="check_box_outline_blank" unchecked-icon="check_box" checked-icon="check_box_outline_blank" keep-color

But having false-value as the true value isn't a very clean solution.

https://jsfiddle.net/wso1av4f/5/

rstoenescu:
Hi,

Maybe I'm missing some point, but are you aware of the Boolean `toggle-indeterminate` prop? If you don't want the indeterminate state, just set it to Boolean "false"..

KnowZero:
@rstoenescu  How do I do that? I tried

toggle-indeterminate="false" and :toggle-indeterminate="false" and they don't seem to work. Am I doing something wrong?

KnowZero:
@rstoenescu From my understanding, toggle-indeterminate is false by default. But even if it is false by default, it still exists until user clicks once. But there should be an option to completely eliminate it. Look at my jsfiddle, the last checkbox is what I would like:

https://jsfiddle.net/wso1av4f/5/

gaby64:
I do need a null-is-false prop

rstoenescu:
@gaby64 `:false-value="null" indeterminate-value="something-else"`

gaby64:
I have tried that, they are indeterminate on render

rstoenescu:
Works fine: https://codepen.io/rstoenescu/pen/YzPJYdv?editable=true&editors=101
Correct the initial value of the model... @gaby64

gaby64:
undefined is my correct inital value for an object array in a v-for

gaby64:
https://codepen.io/gaby64/pen/YzPREwm?editable=true&editors=101


the only change required to make it work is 

`props: {
    value: {
      default: null
    }`

[...]

PR #6146, opened 2020-01-17
fix(QCheckbox) Indeterminate bug

gaby64:
my use case was to allow false to be null/undefined

such as in a v-for of an object array where I want item.selected without having to append a selected property to each item in advance.

`:false-value="null" :indeterminate-value="''"`

`indeterminateValue `is `null `by default so having the `void 0` comparison was redundant and caused the bug

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] It's submitted to the dev branch and not the master branch
- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. fix: #xxx[,#xxx], where "xxx" is the issue number)
- [x] Any necessary documentation has been added or updated in the docs (for faster update click on "Suggest an edit on GitHub" at bottom of page) or explained in the PR's description.

If adding a new feature, the PR's description includes:

### DIFF
--- a/ui/src/components/checkbox/QCheckbox.js
+++ b/ui/src/components/checkbox/QCheckbox.js
@@ -17,7 +17,7 @@ export default Vue.extend({
 
   computed: {
     isIndeterminate () {
-      return this.value === void 0 || this.value === this.indeterminateValue
+      return this.value === this.indeterminateValue || this.value === this.indeterminateValue
     },
 
     classes () {
--- a/ui/src/components/checkbox/QCheckbox.js
+++ b/ui/src/components/checkbox/QCheckbox.js
@@ -17,7 +17,7 @@ export default Vue.extend({
 
   computed: {
     isIndeterminate () {
-      return this.value === this.indeterminateValue || this.value === this.indeterminateValue
+      return this.value === this.indeterminateValue
     },
 
     classes () {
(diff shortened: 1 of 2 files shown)


gaby64:
(fix #2067)

rstoenescu:
Closing as per our discussion.
But please don't be discouraged in PRing.

rstoenescu:
Closing as per our discussion.
But please don't be discouraged in PRing.

gaby64:
so I have to maintain this patch on my end, our discussion was not over

rstoenescu:
This cannot get into Quasar because it would be a breaking change.
Your last argument is that you have a v-for loop and you don't have the models in your data() -- but I showed you how to do it -- and that not having the models is bad practice.

gaby64:
what does it break?

my items are loaded in vuex from an api, I have a checkbox on each item, how would it be bad practice to have a temp property item.checked or item.selected, undefined when not checked, true when checked

gaby64:
why the void 0 comparison? that means undefined is treated as indeterminate and cannot be changed by indeterminate-value

gaby64:
I should be able to have undefined be treated as false

PR #6160, opened 2020-01-20
docs: fix uploader nodejs example

stefanvanherwijnen:

- [ ] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Other information:**
I don't know why it happens, but if you place the nodejs server behind a reverse proxy and send a response outside of `form.parse()`, the upload fails without any error or warnings for larger files. It probably has something to do with buffering of the request, not sure why.
I figured I'd make a PR so no one else runs into the problem with this example.


rstoenescu:
Thanks!

Issue #2780, opened 2018-11-25
[Request] Extended QFab 


nafaabout:
With the current implementation of the QFab, there is no way to add a label to the parent button or its actions. In my current use case, I need to use the Fab component to enable the user to add different types of resources from the same place. So basically all the actions do the same thing but for different types of resources. So it doesn't seem good for me to use only different Icons since the resources are very similar. So I'm afraid that the user would be confused without a label that explains what type of resource action would add. 

Here what I mean from Google's material design site
https://material.io/design/components/buttons-floating-action-button.html#extended-fab

nothingismagick:
This has been resolved in v1.

MadeMyDay:
> This has been resolved in v1.

does it? Can't find a reference. Could you point me to the docs where this is described? Or do you mean just wrap an ordinary button in a q-page-sticky?

pdanpdan:
No, it's not implemented.

rstoenescu:
Extended QFab & QFabActions coming in v1.9.0.

PR #6236, opened 2020-01-27
feat(QFab): Add extended QFab #2780

pdanpdan:


rstoenescu:
Marking as v1.9 material

PR #6288, opened 2020-02-04
fix(docs): Fix issue of anchors getting hidden

ejez:

- [x] Bugfix
- [x] Documentation

- [ ] Yes
- [x] No

**Other information:**
When navigating directly to an anchor, for example: https://quasar.dev/introduction-to-quasar#All-Platforms-in-One-Go , the anchor element gets hidden by the header, this issue can be resolved by adding the css property `scroll-margin-top` to all anchor elements (elements with the `id` attribute) in the layout page.

`scroll-margin-top` is [supported by most browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top#Browser_compatibility).

### DIFF
--- a/docs/src/layouts/Layout.vue
+++ b/docs/src/layouts/Layout.vue
@@ -385,6 +385,6 @@ export default {
   &:hover img
     transform: rotate(-360deg)
 
-.q-page-container [id]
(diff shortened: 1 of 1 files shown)


rstoenescu:
Thanks for the nice enhancement!

Issue #4182, opened 2019-05-22
[V1] [Proposal] clearable & clear-value with model


olegvakulenko:
@rstoenescu 

Before v1 there is exist property `clear-value` and it behavior discussed in issue:
https://github.com/quasarframework/quasar/issues/1994

In v1 when using `clearable` property model set to `null` value only.
For emulate desirable behavior now, if I want to set other clear value, I must add my own action in slot.

When I ask you answered, that `clear-value` property removed for performance reason and I agree with you that permanent checking has some cost. But when I want that kind of opportunity I agreed with performance penalty.

I propose another version of handling `clear-value` property when needed only model for clear value, so if this property defined comparison will be made with two models. When `clear-value` not defined used `null` value.

In other words I propose to make out-of-box solution for this pattern:

```
<q-input v-model="text">
  <template v-if="text !== text$clear" v-slot:append>
    <q-icon name="cancel" @click.stop="text = text$clear" class="cursor-pointer"></q-icon>
  </template>
</q-input>
```

Will be replaced with:

`<q-input clearable clear-value="text$clear" v-model="text">`

or 

`<q-input clearable :clear-value="text$clear" v-model="text">`

(not sure with suitable syntax)
Or use property with name `clear-model` (`:clear-model`).

Same for `QSelect`, where I have custom value for default option.

https://codepen.io/anon/pen/byYmje?editors=1010

What do you think about implementation of this pattern?

pdanpdan:
You can use `@clear` event and set the model to the desired value.
`@clear` event is emitted after the `@input` one that sets the model to null, so the final value of the model will be the one you set on `@clear`.

olegvakulenko:
@pdanpdan 
I understand that with `@clear` event I can set my own default value but in `QSelect` if my clear value not null or not blank, clear icon still visible. In other word I can't get out of box solution when clear icon were hide in case of assign my own default value.
Does my suggestion complicate the logic of work too much or is there a conflict in the settings?

pdanpdan:
Ah, I understand the problem now, sorry - reopening and waiting for @rstoenescu opinion on this.

olegvakulenko:
@pdanpdan 

Hi.
As I can see, PR5756 merged (https://github.com/quasarframework/quasar/pull/5756).
I try to test changes, but see strange behavior. When use clear button on field with slot all work fine, but when use clear button on field with `clearable` property model does't updated correctly, so result of `text !== text$clear` not changed.

[...]

PR #6359, opened 2020-02-12
feat(QField): Add clearVisible prop to force showing or hiding the clear button #4182

pdanpdan:


rstoenescu:
It's nice, but we have the slot. We don't want to start adding props for the default clear...

PR #6386, opened 2020-02-16
feat(QIntersection): allow HTML tag to be used for wrapper element

hawkeye64:


hawkeye64:
This fix now allows things like table rows and columns to be lazy loaded and won't break SSR hydration with extra div.

PR #6393, opened 2020-02-17
feat(model-toggle): add stopPropagation prop

jesusvilla:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
If I want to place QMenu within QExpansionItem or within a selectable QItem I will need to propagate the parent component event, so I have added this new property: stopPropagation
Look: https://codepen.io/jvillanueva/pen/jOPbzjB

### DIFF
--- a/ui/src/mixins/model-toggle.js
+++ b/ui/src/mixins/model-toggle.js
@@ -8,7 +8,8 @@ export default {
     value: {
       type: Boolean,
       default: void 0
-    }
+    },
+    stopPropagation: Boolean
   },
 
   data () {
@@ -29,6 +30,9 @@ export default {
 
   methods: {
     toggle (evt) {
+      // Stops the event propagation of the parent component
+      this.stopPropagation === true && evt !== void 0 && evt.stopPropagation()
+
       this[this.showing === true ? 'hide' : 'show'](evt)
     },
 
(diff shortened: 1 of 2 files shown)


pdanpdan:
Hello. I think  there is no need for this because you can stop the event at parent level. Just add a `@click.stop` on the QIcon.

rstoenescu:
This is a good idea, but I'd like to see it for both show() and hide().

rstoenescu:
And without any prop. We should do this by default.

jesusvilla:
What you tell me is that by default it should be with stopPropagation?, If so, it would be great, I did it with an additional property so as not to affect in other cases.

rstoenescu:
Yes, let's do it by default.

jesusvilla:
But the possibility of propagating the event should be allowed (i still can't find a case), so just add a default value: true

rstoenescu:
I think that we are talking about completely different things.
If all you care is the "click" event which is bound to the toggle() method, then use Vue's ".prevent" modifier. No need for Quasar to handle that. Will make a commit with what I was referring to.

Issue #6352, opened 2020-02-12
How to build SSR with quasar extension using Docker multi-stage build


thr0n:
I'd like to create the SSR production build of my quasar app using Docker multi-stage build (details about this pattern can be found for example here: [https://codefresh.io/docker-tutorial/node_docker_multistage/](https://codefresh.io/docker-tutorial/node_docker_multistage/)). 

So basically this is what I'm doing in my Dockerfile:

First stage - Build the app:
1) Use a node Docker image as builder container
2) Install node_modules using `yarn install`
3) Create the ssr build using `quasar build -m ssr`
4) `cd` into `./dist/ssr` and execute `yarn install` to install dependencies needed for production

Second stage - Serve the app:
1) Use a second node image as production container
2) copy `dist/ssr` contents from the builder container to `/app` inside the production container
3) Run `CMD node index.js` to serve the app

Now here's the issue: I'm using `@quasar/quasar-app-extension-qmarkdown` (added to my project using `quasar ext add @quasar/qmarkdown`) which has been automatically installed as a `devDependency`. Consequently `qmarkdown` and its dependencies (e.g. `markdown-it`) are missing in the production `package.json` file and they cannot be found at runtime:

```
Error: Cannot find module 'markdown-it' from '/app'
    at Function.resolveSync [as sync] (/app/node_modules/resolve/lib/sync.js:81:15)
    at r (/app/node_modules/vue-server-renderer/build.dev.js:9312:44)
    at Object.5d44 (server-bundle.js:1:27965)
    at o (server-bundle.js:1:194)
    at Module.4edf (server-bundle.js:1:16859)
    at o (server-bundle.js:1:194)
    at Object.0 (server-bundle.js:1:1423)
    at o (server-bundle.js:1:194)
    at module.exports.0 (server-bundle.js:1:1384)
    at Object.<anonymous> (server-bundle.js:1:1393) {
  code: 'MODULE_NOT_FOUND'
}
```
Is there a way to treat qmarkdown as a production dependency or to transfer the extension dependencies to the `dist/ssr` directory?

thr0n:
Addition: Of course I can just run `yarn add @quasar/quasar-app-extension-qmarkdown` to add qmarkdown as a production dependency. But shouldn't it be possible to achieve this by using `quasar ext add`? E.g. by the help of a flag?

TobyMosque:
maybe that will help u.:
my Dockerfile:

```docker
# develop stage
[...]

PR #6405, opened 2020-02-18
fix(app-extension): Allow the extension developer to specify the extension install flags, fix: #6352

ejez:

- [x] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

An App  Extension might include code that needs to be executed at the app runtime (ex: boot files provided by the ext), in this case the AE should be added to the app `dependencies` not `devDependencies`, if not, the ssr code resulting from `quasar build -m ssr` when used outside the app directory will not include the AE when it is npm/yarn installed. (see reproduction steps below)

When installing an App extension, quasar currently hardcodes the install flags to `--save-dev` for npm and `--dev` for yarn.

The PR allows the extension developer to decide where his ext should go, based on the code included in it. He can do so by adding the following to the ext `package.json` file:
```json
  "quasar": {
    "extension": {
      "npmInstallFlags": "--save-prod",
      "yarnAddFlags": ""
    }
  }
```
If the above is not provided, then `--save-dev` (or `--dev`) will be used as default to keep the current behavior.

When the PR is accepted/finalized relevant docs will be added.

Note: I wanted to start a new project that will avail quasar docker images, but got into this constraining issue, hence this PR

See #6352 

## Issue reproduction steps
```sh
quasar create testapp
cd testapp
quasar ext add @m8a/apollo
quasar build -m ssr
```
output:
```
...
Tip: The output folder must be yarn/npm installed before using it,
      except when it is run inside your already yarn/npm installed project folder.
```

Now we want to copy and serve the built code from our web server (might be a different machine, but for this example we'll use another directory in the same machine)
```sh
cp -r dist/ssr /path/to/served/app
cd /path/to/served/app
yarn
yarn start
```
Now we get `500 | Internal Server Error`, due missing extension code

@rstoenescu @hawkeye64 @thr0n


hawkeye64:
@ejez Enlighten me, because maybe I am missing something. How is the Quasar extension handler supposed to read the package.json before it's installed?

hawkeye64:
right, `npm view` first to get package.json info

hawkeye64:
At work, after we compile everything, it's copied to a new location, without node_modules and then we install with --production, which does not install anything from devDeps

ejez:
Issue updated with minimal reproduction steps of the problem

ejez:
To quickly test the code introduced by this PR, do the same as the reproduction steps described, and in addition, copy the modified `Extension.js` , and use ext `@ejez/apollo` which has the key `quasar.extension` in package.json

```sh
quasar create ext-test-app
cd ext-test-app
curl -o node_modules/@quasar/app/lib/app-extension/Extension.js https://raw.githubusercontent.com/ejez/quasar/dev/app/lib/app-extension/Extension.js
quasar ext add @ejez/apollo
quasar build -m ssr
cp -r dist/ssr /path/to/served/app
cd /path/to/served/app
yarn
yarn start
```
Now the served app should work fine

ejez:
You can also use the following to test different possibilities of the install flags:

`ext-test.js`

```js
class Extension {
  // extracts extension params defined in its package.json file
  params () {
    // reads package.json key: 'quasar.extension' if defined.
    // execSync returns a buffer
    const paramsBuffer = require('child_process').execSync(
      `npm view --json @ejez/quasar-app-extension-apollo quasar.extension`
    )

    // parse the buffer and return an object
    // if key 'quasar.extension' does not exist we return an empty object
    return paramsBuffer.toString() ? JSON.parse(paramsBuffer) : {}
  }

  __installPackage (nodePackager, params) {
    // const params = this.params()

    // if install flags are not defined, we target devDependencies as default
    const npmInstallFlags = 'npmInstallFlags' in params
      ? params.npmInstallFlags
      : '--save-dev'
    const yarnAddFlags = 'yarnAddFlags' in params
      ? params.yarnAddFlags
      : '--dev'

    // const nodePackager = require('../helpers/node-packager')

    // if install flags are defined as empty strings, we don't concatenate
    const cmdParam = nodePackager === 'npm'
      ? npmInstallFlags === '' ? ['install'] : ['install'].concat(npmInstallFlags)
      : yarnAddFlags === '' ? ['add'] : ['add'].concat(yarnAddFlags)

    console.log(`${nodePackager} command params are: ${cmdParam}`)
  }
}

const ext = new Extension()

const p = [
  ext.params(),
  {},
  { npmInstallFlags: '--save-prod', yarnAddFlags: '' },
  { npmInstallFlags: '--save-dev', yarnAddFlags: '--dev' }
]

p.forEach(params => {
  console.log(params)
  ext.__installPackage('npm', params)
  ext.__installPackage('yarn', params)
  console.log('******************')
})
```

```sh
node ext-test
```

ejez:
Few clarifications about the added code:

`npm view --json` :
- throws error if the extension does not exist in npm registry, `params()` then returns an empty object (`catch` statement)

`JSON.parse()`:
- `JSON.parse(something)` is a shorthand of `JSON.parse(something.toString())` , `toString` gets called implicitly when using `JSON.parse`
Quote from its [spec](http://www.ecma-international.org/ecma-262/6.0/#sec-json.parse):
  > Let JText be ToString(text).

- if ext exists but key `quasar.extension` does not exist, `execSync` returns an empty buffer, which leads to `JSON.parse('')` (parsing empty string) which throws an error (`catch` statement covers this case)

ejez:
In the last commit we use the same install flags for removal also, this is mainly to account for when the user has modified the default npm config (for ex if he has done `npm config set save-dev=true`)

hawkeye64 (approved):
This is looking good. I've wanted this for quite some time, especially for UI-bases app-extensions. At work, we build all of our micro-services and client, them move it all sans node_modules, then run the install again with `--production` which only installs `dependencies`. For app-exts, we have to manually move them from `devDependencies` to `dependencies`. A more automated approach is appealing. :+1:

smolinari (commented):
I'm not enough of a subject expert to offer anything to this that might improve it. Good job nonetheless!

Scott

yunseop-kim:
@rstoenescu When does it merge? I'm looking forward to waiting it. 😭

rstoenescu:

Issue #6358, opened 2020-02-12
arrows props at QPagination


matro-skin:
**Is your feature request related to a problem? Please describe.**
Add icon props for left and right arrows at `QPagination`

**Describe the solution you'd like**
Like `prev-icon` and `next-icon` at `QCarousel`

rstoenescu:
@hawkeye64 We need a `icon-first`, `icon-prev`, `icon-next`, `icon-last` props to be hooked into existing `icons` computed prop.

hawkeye64:
Hope you mean v1.9.4 @rstoenescu :)

rstoenescu:
Yep. Released 1.9.3 already. Couldn't postpone it more. :)

rstoenescu:
Enhancement will be available in "quasar" v1.9.4.

PR #6409, opened 2020-02-18
feat(QPagination): arrows props at QPagination (#6358)

hawkeye64:


docs/src/examples/QPagination/Icons.vue
```diff
@@ -0,0 +1,37 @@
...
+    >
+    </q-pagination>
+  </div>
+</template>
+
+<script>
+import {
+  laAngleDoubleLeftSolid, laAngleDoubleRightSolid, lasAngleLeft, lasAngleRight
+} from '@quasar/extras/line-awesome'
+
```
rstoenescu: Only comment would be to avoid using code that would break on UMD, otherwise codepen will stop working.
Try to use only Material Icons (webfont, not svg). If it's not possible to do something nice, then we drop the example.

rstoenescu (changes requested):

docs/src/examples/QPagination/Icons.vue
hawkeye64: Got it, will make appropriate changes

Issue #6412, opened 2020-02-18
Elevation prop for components


GordonBlahut:
**Is your feature request related to a problem? Please describe.**
For some components, it does not seem possible to change the elevation using `shadow-N` classes because you can't target the correct DOM element (e.g. FAB).

**Describe the solution you'd like**
Elevation prop that allows to specify the elevation which would be applied to the DOM element most suitable depending on the component.

```vue
<q-fab elevation="12" />
```
**Describe alternatives you've considered**
Custom CSS but prone to breaking because undocumented DOM structure may change.

**Additional context**
I've mostly run into this on the FAB which I would like to change the box-shadow but it ends up creating a square box around the fab if you use a `shadow-N` class.

https://codepen.io/gordonblahut/pen/bGdeaaN?editable=true&editors=101

hawkeye64:
@GordonBlahut Here's your codepen fixed: https://codepen.io/Hawkeye64/pen/dyoXJwa?editors=1010
You just need to apply style `border-radius: 50%` to the QFab.
Leaving this open as @rstoenescu may want to do this in the code. I see no point for the fab to be square unless the `square` property is used.

GordonBlahut:
@hawkeye64 yeah that's my current workaround but it's very focused to the specific example as opposed to a general solution for all cases. 

You end up with separate drop shadows on `.q-fab` and `.q-btn__wrapper` but it's not the end of the world.

I guess as long as the shadow-N classes work as expected then a prop is not necessary.

hawkeye64:
@GordonBlahut The fix I did is to add the missing `q-fab--form-rounded` class, which should help out one part of your issue. It'll be up to @rstoenescu if QFab should get an `elevation` prop.

hawkeye64:
@GordonBlahut Ok, that was not a good fix -- I didn't do enough testing. It made the Fabs look like footballs when used with label. So, it's something you would have to do manually. But, don't round it if using labels -- looked horrid.

rstoenescu:
Enhancement will be available on "quasar" v1.9.5.

PR #6414, opened 2020-02-18
fix(QFab): add missing classes used in code

hawkeye64:
ref: Elevation prop for components (#6412)
Class injected by this code was missing:
  computed: {
    formClass () {
      return `q-fab--form-${this.square === true ? 'square' : 'rounded'}`
    },

### DIFF
--- a/ui/src/components/fab/QFab.sass
+++ b/ui/src/components/fab/QFab.sass
@@ -11,6 +11,9 @@
   &--form-square
     border-radius: $generic-border-radius
 
+  &--form-rounded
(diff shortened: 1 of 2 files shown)


pdanpdan:
This would break the QFab with internal label. Why do you need it?

rstoenescu:
Yes, it's not used, but that class is part of the efforts on making CSS customization easy, a.k.a avoid using :not(...) selectors.

hawkeye64:
Right, they look like a football. :disappointed:

PR #6418, opened 2020-02-19
feat(QDate): new date range selection - new props

hawkeye64:


hawkeye64:
image

yusufkandemir:
This is outdated. Date range selection has been implemented: https://github.com/quasarframework/quasar/pull/7177#issuecomment-648719829

Aside from the main functionality, it brings color and text color props that are specific to the range. But, I think it's not in popular demand. Even if it was, it would be best to open a separate PR.

So, I am closing this PR.

PR #6437, opened 2020-02-20
feat(QInput): Add shadow text (fill value)

pdanpdan:


ui/src/components/input/QInput.js
```diff
@@ -267,6 +275,17 @@ export default Vue.extend({
           }
           : this.formDomProps
       })
+
+      if (this.hasShadowText !== true) {
+        return control
+      }
+
+      return [
```
rstoenescu, line 283: Forces Vue to perform an extra step to normalize the vm children.

ui/src/components/input/QInput.js
```diff
@@ -266,7 +274,18 @@ export default Vue.extend({
...
           }
           : this.formDomProps
-      })
+      }
+
+      const control = h(this.isTextarea === true ? 'textarea' : 'input', data)
+
+      return this.hasShadowText !== true
+        ? control
+        : [
```
rstoenescu, line 283: Forces Vue to normalize the vm's children list... Any other way to do this?

PR #6442, opened 2020-02-20
Expose QSelect.getEmitValue(opt)

Propaganistas:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

Depending on the `emit-value` prop, `QSelect` emits a different value. I'm making a component wrapping `QSelect` and want to allow both `emit-value` and `multiple` to be chosen upon use of the wrapper component. Different actions need to be taken based on the `multiple` property. When `emit-value` is present as well, this adds an additional layer of complexity as the value could be in two formats, either the raw value or the specified option-value.

In order to fully rely on the behavior of `emit-value` I think it's good to expose a method that outputs the value that will be emitted in the end. This way the decision logic based on `emit-value` remains encapsulated in Quasar.

### DIFF
--- a/ui/src/components/select/QSelect.js
+++ b/ui/src/components/select/QSelect.js
@@ -331,7 +331,7 @@ export default Vue.extend({
   },
 
   methods: {
-    getEmitValue(opt) {
+    getEmitValue (opt) {
        return this.emitValue === true
          ? this.__getOptionValue(opt)
          : opt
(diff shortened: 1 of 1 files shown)


ui/src/components/select/QSelect.js
```diff
@@ -331,6 +331,12 @@ export default Vue.extend({
   },
 
   methods: {
+    getEmitValue() {
```
rstoenescu: Did u mean `getEmitValue (opt)`?

Propaganistas: Undoubtedly yes! It was a quick 'n dirty fork/edit/PR flow. Updating right now.

rstoenescu:
No reason to use getEmitValue() in places where the value is already computed. Will make some changes then merge.

Issue #6466, opened 2020-02-24
[feat] QTabs/QTab needs text-align prop


hawkeye64:
Currently, QTabs has an `align` prop. This is basically how the actual tabs align in horizontal mode on a tab row. Since vertical tabs were introduced, this property is no longer as clear and people have been using it to try and align text within a tab.

A new `text-align` property would determine how the text within a tab would be aligned. This can be used for both horizontal and vertical tabs, but would be most visible and make most sense for the vertical tabs.

hawkeye64:
or this, to be similar to `inline-label` on QTabs, this could be called `align-label`

tanthammar:
Or/And  add the "align" and "justify" prop to q-tab.
Naming consistency ...

Example:
```html
<q-tabs
   vertical
   align="left" //existing prop to control row justification
>
<q-tab
   align="left || right ..." // new prop, grouped alignment of label and icon. 
   justify="between || around || start || end ..." // new prop, spacing between label and icon
/>
</q-tabs>
```

hawkeye64:
@tanthammar I made further adjustments that cover your scenario

rstoenescu:
Breaking changes... can't merge this one...

hawkeye64:
@rstoenescu Can you tell me what's breaking? Is it fixable?

Jokertwo:
Is there any progress?

hawkeye64:
@Jokertwo Nope. Pretty much dead.

dsl101:
I was about to submit a FR for this. As a really simple option perhaps, if the QTabs is `vertical` and the `align="left"` prop is set, applying `style="justify-content: initial"` to the QTab children achieves the left alignment.

Any chance it could be resurrected?

hawkeye64:
@dsl101 You'd also have to NOT break vertical tabs

dsl101:
Sorry? This is about improving vertical tabs...

PR #6468, opened 2020-02-24
feat(QTabs): new align-label property

hawkeye64:


hawkeye64:
Forgot to reference: [feat] QTabs/QTab needs text-align prop #6466

hawkeye64:
I have made adjustments so that the `align-label` prop only affects `vertical` tabs with `inline-label`

PR #6472, opened 2020-02-25
fix(QIntersection): remove redundant div

hawkeye64:


rstoenescu:
It's not redundant. It's needed when transition is specified (needs to be keyed). In order to maintain consistency with the case when no transition is specified, the wrapper is still being used.

hawkeye64:
I see, thanks for the explanation. I appreciate it.

Issue #6486, opened 2020-02-26
QIntersection "hurts" SSR


eladcandroid:
I'm using the lovely q-intersection with prefetched items but I can't see them on the page view-source.. when I'm just replacing q-intersection element into div, the items appear again... 

Example code:
```javascript
      <q-intersection
        once
        v-for="item in items"
        :key="item.slug"
        class="col-md-3 col-sm-4 col-xs-12 item-card"
      > .... </q-intersection>
```

    Dev mode.......... ssr
    Pkg quasar........ v1.9.4
    Pkg @quasar/app... v1.5.8

pdanpdan:
QIntersection only shows the content when it's visible in the client (browser) viewport, so there is no way to use it with SSR as it is now.

eladcandroid:
@pdanpdan Thanks. Now i'm handling this with q-no-ssr with a placeholder. 

```javascript
<q-no-ssr class="row justify-center items q-gutter-md q-ma-md">
        <q-intersection
          once
          v-for="item in items"
          :key="item.slug"
          class="col-md-3 col-sm-4 col-xs-12 item-card"
        >
          <ItemPreview :item="item" />
        </q-intersection>
        <template v-slot:placeholder>
          <div
            once
            v-for="item in itemsPaginated"
            :key="item.slug"
            class="col-md-3 col-sm-4 col-xs-12 item-card"
          >
            <ItemPreview :item="item" />
          </div>
        </template>
</q-no-ssr>
```

If there's a better way, please let me know.

pdanpdan:
Your solution is very good (I suppose itemsPaginated is a list of above the fold items).
I'll try to find a better solution if I can, so please leave the issue open to track it.

Thank you

pdanpdan:
@rstoenescu Do you think a `on-ssr` prop that would force render on SSR would be a good approach?
It would allow pre-rendering a limited number of items (the above the fold part).

rstoenescu:
If the developer wants to show something regardless of the Intersection observer (which obviously can't run on server-side) then he/she shouldn't use QIntersection for the first N items to begin with. If we add a prop to circumvent the target behavior, then why is the component needed in the first place?

rstoenescu:
After some more discussions --> new Boolean prop "ssr-prerender" will be available in "quasar" v1.9.5.

PR #6499, opened 2020-02-27
feat(QIntersection): add onSsr prop to force pre-render on SSR #6486

pdanpdan:


rstoenescu:
This is redundant. If developer wants to show something regardless of intersection, then he/she shouldn't use QIntersection for the first N items to begin with.

pdanpdan:
I agree if the case is of using `.once` modifier.
But in case of a long list one might want to keep a consistent component rendering (all items are QIntersection) and might want to show/hide the top ones on scroll (with the expected effects)

PR #6550, opened 2020-03-05
feat(utils/date): add arg opts.onlyDate in isBetweenDates

jesusvilla:
The possibility of comparing dates without taking into account the hours, minutes, seconds and milliseconds was added. Taking into account the optimal conditions for this.
https://jsperf.com/datebetween/23

and just to ensure that it is correct only date and not time:
https://jsperf.com/datebetween/1


rstoenescu:
Excellent work. Well written.
Now docs update :)

Issue #4891, opened 2019-08-13
Round ripple not showing properly on iOS safari


rooneyl:
**Describe the bug**
Button with round style should have rounded ripple effect, but on iOS safari it shows as regular square ripple when I press button quickly multiple times. It seems like its happening only for the buttons in toolbar

**To Reproduce**
1. Go to 'https://quasar.dev/vue-components/toolbar#Example--With-Avatar' on *iPhone safari*
2. Click on 'any button on toolbar multiple time quickly'

**Expected behavior**
It should always show round ripple effect with round option

**Screenshots**
Expected:
<img width="236" alt="Screen Shot 2019-08-13 at 11 46 10 AM" src="https://user-images.githubusercontent.com/23136511/62970319-12504180-bdc4-11e9-9378-b8ff2ed9f765.png">
Actual:
![ezgif com-crop](https://user-images.githubusercontent.com/23136511/62970324-17ad8c00-bdc4-11e9-9eac-a0ec41f5d522.gif)

**Platform (please complete the following information):**
iOS: 12.3.1

smakinson:
Noticing a possibly related issue on flat round buttons and the Ripple. You can see the bottom left corner breaks out of the button when clicked on the docs here: https://quasar.dev/vue-components/button#Design

Click either one of the round buttons in the first 2 rows there and watch the bottom left corner.

pdanpdan:
I suppose this was fixed by Safari in the newer releases, because i cannot be reproduced on iOS 13+.
If it is present only on iOS 12 Safari then it's a bug in the browser and we cannot fix it, sorry.

soccer99:
Seeing the same issue as @smakinson on the latest version of google chrome on Mac OSX.  The bottom left of the round ripple shows through the circle.

rstoenescu:
Cannot reproduce with latest Quasar on MacOS or iOS (v13+). Probably Apple fixed this bug.

PR #6552, opened 2020-03-05
feat(Ripple): Use 2 phased approach, prevent repetition, allow overlapping ripples #4891, #7474

pdanpdan:
https://pdanpdan.github.io/quasar-docs/vue-directives/material-ripple

#4891, #7474


pdanpdan:
Examples:

https://codepen.io/pdanpdan/pen/jOWPwWj

PR #6565, opened 2020-03-06
feat(QForm): Enforce DOM order of elements for validation

pdanpdan:


rstoenescu:
It enforces validation by DOM order but it destroys perf...

PR #6648, opened 2020-03-21
feat: app extension renderFile() function

stefanvanherwijnen:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

I needed a way to render a single file into the app with different content. E.g. I make a single page .vue file which is rendered into the app with a different template body for different simple text-only pages. With `render()` this is only possible by first creating all individual files in the template folder, but I wanted  to dynamically create them from JS.

In order to do so I had to pull apart the `render()` function. Let me know if any changes are required.


stefanvanherwijnen:
@rstoenescu 
Any thought on using [consolidate](https://github.com/tj/consolidate.js/) for example instead of lodash templating?
Combined with ESLint the lodash templating works quite poorly because any whitespace in the rendered file will throw an error. Also, the lodash templating syntax is not supported in .vue files in Visual Studio Code.

rstoenescu:
Hi,

Thank you for contributing! This will go into "@quasar/app" v2 (to be released this week). Sorry for the delay (this got under the radar unfortunately).

Regarding "consolidate" package: I think we should keep things simple (at least for now). We'll reopen these talks for "@quasar/app" v3 (along with Quasar v2) which will bring Vue 3. Agreed?

stefanvanherwijnen:
Sounds good. I agree on keeping it simple. I think there should be a decision on which (single) templating language to use, but that's open for debate.

My only issue is that ESLint does not like mixing up syntaxes in one file, but functionality wise the current solution works fine.

PR #6680, opened 2020-03-25
feat(QField): Improve a11y for clear button (make it focusable and work like a button)

pdanpdan:


greseky:
https://quasar.dev/vue-components/input#Clearable

2nd case is without clearable attribute but clear button is shown (and not focusable).

In previous version if there was no clearable prop then there was no clearable icon...
Shouldn't that keep that way?

pdanpdan:
Have you checked the code, especially the `<!-- equivalent -->` part? - it's an example on how to do a clearable by hand.

greseky:
I see now...  thank you :) of course i needed to remove clearable prop and added v-slot append.

PR #6693, opened 2020-03-27
fix(QDrawer): Clear layout instance when side is changed

pdanpdan:


rstoenescu:
This is incomplete. Will push a commit on top of this.

PR #6726, opened 2020-04-02
fix(docs): Try to fix tab-to-search with "title" attr

panstromek:
MDN says that it's required and it must match the "shortname" in xml description file. It's kinda weird because I didn't need it for testing, but since tab to search doesn't work on `quasar.dev` on any machine I tried, let's try if this fixes it.

see [here](https://developer.mozilla.org/en-US/docs/Web/OpenSearch) for more details

Btw. If I add the site manually in `chrome://settings/searchEngines` (which is very tedious), Everything works fine, so I really have no idea what's wrong here. I've seen some other tips to do, but most of them seems fairly outdated (like - "don't serve the xml over https", which is nonsense)

### DIFF
--- a/docs/src/index.template.html
+++ b/docs/src/index.template.html
@@ -12,7 +12,7 @@
     <meta property="og:image" content="https://cdn.quasar.dev/logo/512/quasar-logo.png" />
     <meta property="og:site_name" content="Quasar Framework" />
 
-    <link type="application/opensearchdescription+xml" rel="search" href="https://cdn.quasar.dev/other/search_manifest.xml">
+    <link type="application/opensearchdescription+xml" rel="search" title="Quasar docs" href="https://cdn.quasar.dev/other/search_manifest.xml">
     <link rel="icon" type="image/svg+xml" href="https://cdn.quasar.dev/logo/svg/quasar-logo.svg">
     <link rel="icon" type="image/png" sizes="128x128" href="https://cdn.quasar.dev/app-icons/favicon-128x128.png">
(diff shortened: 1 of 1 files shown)


panstromek:
If this doesn't help, one other troubleshooting tip from MDN is:

`Your server should serve OpenSearch plugins using Content-Type: application/opensearchdescription+xml.`

rstoenescu:
Hope it works. If it doesn't then we'll move the xml file back on docs and I'll program it to be served with that HTTP header.

panstromek:
@rstoenescu This indeed didn't help, so I tried to look at differences between my test netlify deployment and quasar.dev and it seems that the only difference here is that manifest is served from the same domain in my case. Since my test deployment serves this file as  `application/xml`, too, I don't think `content type` is the problem.

So I tried to dig deeper and yes, I've found deep hidden in the internet this bit on wikipedia:

> OpenSearch Description Document must be placed on a web server of the same domain.

[source](https://en.wikipedia.org/wiki/OpenSearch)

It links this info to source repo of [opensearch](https://github.com/dewitt/opensearch) spec repo, but I couldn't find this phrase there, but some hazy bits about same origin policy are there.

From the websites I checked, all of them serve the file from the same domain, so I bet that's our problem here. We should move that file back to the docs. If you don't do it yourself sooner, I'll prepare a PR in the evening 😉 Sorry for the confusion, I should have researched this better in the first place.

rstoenescu:
No worries. Moved it back and will be available on next deployment.
Thanks for your work on this!

panstromek:
tab-to-search

works 😉

rstoenescu:
Great great job @panstromek !

PR #6730, opened 2020-04-03
feat(QDrawer): Prevent getting keyboard focus when hidden

pdanpdan:


rstoenescu:
This breaks the hiding effect... Drawer gets invisible before the hide animation starts

Issue #6768, opened 2020-04-09
Unable to build with capacitor


Dinwy:
**Describe the bug**

I was building app for android by following tutorial and having a trouble with android studio.

**To Reproduce**

1. Follow the tutorial until step 2. https://quasar.dev/quasar-cli/developing-capacitor-apps/preparation#2.-Add-Capacitor-Quasar-Mode

2. Move to step 3, and type. `$ quasar dev -m capacitor -T android`
3. "Cannot open C:\Program" error has been thrown which reported in https://github.com/quasarframework/quasar/issues/5556
  - Able to fix this by editing helpers/open-ide.js
4. After fixing `androidStudioPath` error, type  `$ quasar dev -m capacitor -T android` again. Android stuido has opened and Manifest merge failed error has been thrown.

```java
Manifest merger failed : Attribute application@appComponentFactory value=(android.support.v4.app.CoreComponentFactory) from [com.android.support:support-compat:28.0.0] AndroidManifest.xml:22:18-91
	is also present at [androidx.core:core:1.2.0] AndroidManifest.xml:24:18-86 value=(androidx.core.app.CoreComponentFactory).
	Suggestion: add 'tools:replace="android:appComponentFactory"' to <application> element at AndroidManifest.xml:5:5-44:19 to override.
```

- If you go back to `2.` above, and type `npx cap add android` -> open android studio manually didn't throw the error. And able to build SPA.

**Expected behavior**
- Error should not be thrown when following the tutorial.

**Screenshots**

![image](https://user-images.githubusercontent.com/16696001/78854471-62b79600-7a5c-11ea-8dc9-09ab22a44b80.png)

**Platform (please complete the following information):**
OS: Windows 10 Pro
Node: 12.16.1
NPM: 6.13.4
Yarn: 1.22.4
Browsers: Firefox
AndroidStudio: 3.6.2

JayteeNZ:
Can confirm, I've also been receiving this error. Fixing the open-ide file also worked for opening Android Studio.

Enabling androidX worked for the manifest error, but caused the app not to boot up. Will try using npx now.

JayteeNZ:
I reviewed some commits. The ``open`` package for opening AndroidStudio was updated, and fixed one of it's own bugs which involved the use of double-quotes for paths.

So, I guess, the fix for this is to just remove the double quotes from open-ide.js

@Dinwy

rstoenescu:
Fix will be available in "@quasar/app" v1.7.0

[...]

PR #6788, opened 2020-04-11
fix(#6768): Removed double quotes to allow resolving of Android Studi…

JayteeNZ:
The [open](https://github.com/sindresorhus/open) npm package was updated to 7.0.3, which fixed an internal issue regarding path resolutions with double quotes. Quasar upgraded to this version.

7.0.3 now adds the double quotes.


rstoenescu:
Thank you for taking time for this PR!
It will go in "@quasar/app" v1.7.0 soon.

Issue #6795, opened 2020-04-12
Masonry mode on grid row


PH7-Jack:
**Is your feature request related to a problem? Please describe.**
Um recurso similar ao do [link](https://tobiasahlin.com/blog/masonry-with-css/), para fazer as colunas da grade auto ajustar nos espaços livres

**Describe the solution you'd like**
Grid columns with masonry

pdanpdan:
Hello. This is a nice solution, but it is something highly specific (at most 3 columns) and need css adjustment for each use case, so it cannot be implemented in a framework.
Please use English, thank you.

rstoenescu:
@pdanpdan pls create an example in docs with this technique

PR #6796, opened 2020-04-12
docs(flex): Add examples for using flex row/column break and masonry #6795

pdanpdan:


rstoenescu:
An example on QTable doc page with grid mode with masonry would be kick-ass, if you have time :)

Issue #6771, opened 2020-04-09
[Capacitor] capacitor-runtime not added to index.html when https in on


J3m5:
**Describe the bug**
When running capacitor in dev mode with https enabled, the index.html file does not contain the capacitor-runtime script that add cordova, capacitor and plugins.

**To Reproduce**

1. Generate a new app
2. add capacitor mode
3. In quasar.config.js, set devServer > https to true
4. In src-capacitor directory, set the capacitor dependencies to 1.5.2
5. Run quasar dev -m capacitor -T android
6. Once in android studio, launch app on device
7. Inspect app in chrome remote debugger
8. In source tab, look at index file

  This is the html page:
```html
<!DOCTYPE html>
<html>
  <head>
    <title>Quasar App</title>

    <meta charset="utf-8">
    <meta name="description" content="A Quasar Framework app">
    <meta name="format-detection" content="telephone=no">
    <meta name="msapplication-tap-highlight" content="no">
    <meta name="viewport" content="user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1, width=device-width, viewport-fit=cover">

    <link rel="icon" type="image/png" href="statics/app-logo-128x128.png">
    <link rel="icon" type="image/png" sizes="16x16" href="statics/icons/favicon-16x16.png">
    <link rel="icon" type="image/png" sizes="32x32" href="statics/icons/favicon-32x32.png">
    <link rel="icon" type="image/png" sizes="96x96" href="statics/icons/favicon-96x96.png">
    <link rel="icon" type="image/ico" href="statics/icons/favicon.ico">
  </head>
  <body>
    <!-- DO NOT touch the following DIV -->
    <div id="q-app"></div>
  <script type="text/javascript" src="app.js"></script><script type="text/javascript" src="vendor.js"></script></body>
</html>
```
9. Close dev server
10. In quasar.config.js, set devServer > https to false
11. Run quasar dev -m capacitor -T android
13. Once in android studio, launch app on device
14. Inspect app in chrome remote debugger
15. In source tab, look at index file
  This is the html page:
 ```html
<!DOCTYPE html>
<html>
  <head>
<script type="text/javascript">window.Capacitor = { DEBUG: true };

//# sourceURL=capacitor-runtime.js

(function(win) {
  win.Capacitor = win.Capacitor || {
    Plugins: {}
  };

  var capacitor = Capacitor;

  // Export Cordova if not defined
  win.cordova = win.cordova || {};

[...]

PR #6808, opened 2020-04-13
feat(app/capacitor): upgrade to capacitor 2.0, fixes #5823

### LINKED ISSUE #5823: EnableHttpsSelfSigned.enable prevents Capacitor plugins from working
If `EnableHttpsSelfSigned.enable(findViewById(R.id.webview));` is present in `MainActivity.java`, it will prevent Capacitor native plugins from working, since `EnableHttpsSelfSigned` class overrides the `WebViewClient` that Capacitor uses to inject the javascript that makes plugins work.

nklayman:

- [x] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

This PR will use a new method to enable https support for capacitor. It is only available since version 1.5.1 of capacitor. In addition, new projects created will use capacitor 2.0. Other than the https issue, it will work with existing pre-2.0 capacitor projects. I can add in a check to use the old https method for pre-1.5.1 projects to mitigate the breaking change.

**Other information:**

Dev/build process stays the same. Have not testing for IOS, someone has to before this gets merged. Also need someone to test that native plugins still work with https enabled.


J3m5:
This is great, thank you!

You might want to change the hard coded path [here](https://github.com/quasarframework/quasar/blob/27092c4c23e907a53e5b627ff4884e672d6c8b6b/app/lib/capacitor/capacitor-config.js#L193) to the real ones, because the path is only valid if the capacitor or cordova id has not been modified.
If it has, the patch will no be applied.

nklayman:
@J3m5 My latest commit should have fixed that issue, it uses a glob to automatically find the proper path. If you could test it out that would be awesome.

J3m5:
I'll try to test it today and keep you posted.

IlCallo:
@J3m5 hey there! Did you find the time to test this?

J3m5:
Hey ! 
Unfortunately no, I've been quiet busy in the last few days. 
I'll try tomorrow and let you know.

J3m5:
So, I've tested it, it handles the self signed certificate errors correctly but it still doesn't solve the problem of missing cordova and capacitor runtime in the index.html file that is served to the client.

Like I previously said in issue #6771, I don't think it is strictly related to the ssl errors, but it has more to do with two possible things:

1. The file served with https enabled in not the one that is modified by capacitor ( with the cordova and capacitor scripts injected)

2. The good file is served but the injection didn't happened

J3m5:
I think we could merge this and deal with the ssl error in another PR to resolve the issues caused by the capacitor upgrade problems.

rstoenescu:
Merging, testing, and adapting to q/app v2.

PR #6833, opened 2020-04-15
Pass current route in the btn mixin to router resolve

dmitry:

- [x] Feature

- [x] No

**Other information:**

QRouteTab has the following code: 

```
        current = this.$route,
        { href, location, route } = this.$router.resolve(this.to, current, this.append),
```

It would be great to have the same for the btn when passing to `:to`.

### DIFF
--- a/ui/src/mixins/btn.js
+++ b/ui/src/mixins/btn.js
@@ -97,7 +97,7 @@ export default {
       }
 
       if (this.hasRouterLink === true) {
-        attrs.href = this.$router.resolve(this.to).href
+        attrs.href = this.$router.resolve(this.to, this.$route).href
         attrs.role = 'link'
       }
       else {


dmitry:
@rstoenescu please checkout this change. It's really simple but it allows to see current route in extended router resolver.

dmitry:
@rstoenescu please let me know if anything missing here.

rstoenescu:
Hi @dmitry 

Thanks for contributing!

The [signature](https://router.vuejs.org/api/#router-resolve) of the $router.resolve() method specifies that the second parameter is the current route by default anyway. So wondering why you think it should be explicitly added?

dmitry:
@rstoenescu thank you for the answer.

`QRouteTab` has

```javascript
{ href, location, route } = this.$router.resolve(this.to, current, this.append),
```

I think `q-btn` component should have the same code passing current route into the `resolve` function.

Thanks to your comment, I've fixed an issue without that change. I've extended `$router.resolve()` but forgot to to add `this.history.current` for the current if it's missing.

yusufkandemir:
Thanks for contributing, this is no longer needed. And now, every component that interacts with the router links uses the `useRouterLink` composable. So, they pretty much have a unified behavior. Also, a lot about Quasar, Vue, and Vue Router has changed in 2.5 years. So, I am closing this PR.

Issue #6883, opened 2020-04-25
Need a TypeScript support for the Bex mode


Arsync:
With project, written on TS, got message
```
(TS) Property 'bex' does not exist on type 'QVueGlobals'.
```
for `this.$q.bex` in App.vue

For @webnoob

webnoob:
@Arsync Once the PR's are merged, this should be it.

rstoenescu:
Merged both. Fix will be available in "@quasar/app" v1.7.5.

PR #6887, opened 2020-04-25
fix(bex): Add TS support to BEX mode #6883

webnoob:
FYI: Commented code added to `dom-hooks.js` to avoid issues with TS rule with no empty functions.


webnoob:
Hold this for now, need to improve it.

ui/types/globals.d.ts
```diff
@@ -28,14 +28,21 @@ type GlobalQuasarIconMapFn = (
   iconName: string
 ) => { icon: string } | { cls: string; content?: string } | void;
 
+interface GlobalQuasarBex {
```
IlCallo, line 31: I'd put this into "@quasar/app" types, keeping the indirection level which is in place for "cordova" and "electron" to avoid creating problems for Vue CLI users.
The file used for those is "quasar/app/types/globals.d.ts".
It's mostly a mattery of consistency, your types right now doesn't break anything (but could in the future if you have to reference external types provided by "@quasar/app").
Your choice about actually moving it there.
You could then put it directly into "quasar/app/types/globals.d.ts" or into a standalone "bex.d.ts" (and import it into global declaration files).

PR #6932, opened 2020-05-02
chore(docs): improve right menu and search from url behavior when route changes

pdanpdan:
- show right menu on first load
- perform search from url on route change
- catch error when navigating to sam eroute from search


rstoenescu:
Search from url is a one time startup thing. Not ok to run it on each route change.

PR #6974, opened 2020-05-07
build(typescript): generate narrower types

hamiltoes:
## Summary

Generated type interfaces stand to benefit from narrower types.

Since components/plugins/etc already define `values` in their JSON defs when a prop can only be of a certain subset of values, we can include these in the type interface. This also helps with functions signatures that only accept certain values.

For a basic example, the `Dark` plugin interface in `index.d.ts` currently looks like
```ts
export interface Dark {
    /**
     * Is Dark mode active?
     */
    isActive : boolean
    /**
     * Dark mode configuration (not status)
     */
    mode : boolean | string
    /**
     * Set dark mode status
     * @param status Dark mode status
     */
    set (status : boolean | string): void
    /**
     * Toggle dark mode status
     */
    toggle (): void
}
```
a nicer dev experience would be to have an interface like the following that eliminates a 20 second lookup in the docs for what `string` values are valid.

```ts
export interface Dark {
    /**
     * Is Dark mode active?
     */
    isActive : boolean
    /**
     * Dark mode configuration (not status)
     */
    mode : boolean | 'auto'
    /**
     * Set dark mode status
     * @param status Dark mode status
     */
    set (status : boolean | 'auto'): void
    /**
     * Toggle dark mode status
     */
    toggle (): void
}
```

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No
  - unless a component/plugin/etc.'s `json` def incorrectly defines `values`

- [ ] ~When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)~
- [ ] ~It's been tested on a Cordova (iOS, Android) app~
- [ ] ~It's been tested on a Electron app~
- [ ] ~Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs) (for faster update click on "Suggest an edit on GitHub" at bottom of page) or explained in the PR's description.~

**Other information:**

### DIFF
--- a/ui/build/build.types.js
+++ b/ui/build/build.types.js
@@ -36,39 +36,30 @@ const fallbackComplexTypeMap = new Map([
   ['Object', 'LooseDictionary']
 ])
 
+const dontNarrowValues = [
+  '(Boolean) true',
+  '(Boolean) false',
+  '(CSS selector)',
+  '(DOM Element)'
+]
+
 function convertTypeVal (type, def, required) {
   if (def.tsType !== void 0) {
     return def.tsType
   }
 
-  if (def.values) {
-    const enums = def.values.map(v => {
-      if (v === '(Boolean) true' || v === true) {
-        return 'true'
-      }
-      else if (v === '(Boolean) false' || v === false) {
-        return 'false'
-      }
-      else if (v === '(CSS selector)') {
-        return 'string'
-      }
-      else if (v === '(DOM Element)') {
-        return 'Element'
-      }
-      else if (v === null) {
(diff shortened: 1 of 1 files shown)


rstoenescu:
Beautiful! Thanks for contributing!
This will go into "quasar" v1.11.0.

PR #7007, opened 2020-05-11
Chip RemoveIcon Prop

jjjrmy:
Currently the only way to change the Remove icon on a Q-Chip is to set it like this:
`Vue.prototype.$q.iconSet.chip.remove = 'fas fa-times-circle'`
That sets it globally, but I added a prop so you can set it per instance.

- [x] Feature

- [x] No

### DIFF
--- a/ui/src/components/chip/QChip.js
+++ b/ui/src/components/chip/QChip.js
@@ -34,6 +34,7 @@ export default Vue.extend({
 
     icon: String,
     iconRight: String,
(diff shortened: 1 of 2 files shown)


jjjrmy:
Also added a class to the QSelect if it is using chips or not, this is useful same as if it using the native input or not.

rstoenescu:
Thanks for contributing!

PR #7022, opened 2020-05-13
Update select.md

smolinari:
Hope this is ok. 

Scott

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

**Other information:**


pdanpdan:
It's not good - the difference would be that filtering means selecting one/mode of the options in the list (even that means first adding the new option to the list) while autocomplete means completing the text you write with the text of the option in the list (but you can then delete/add some more text to it and use it as it is).
Autocomplete only works with text models.

smolinari:
Ahh... ok. I'll edit it accordingly.

Scott

smolinari:
@pdanpdan - Can you see if it is correct now? Thanks for helping out.

Scott

pdanpdan:
The intended difference between a filtering QSelect and an autocomplete QSelect is that the former only allows setting the model to options from the list of available options (even if that means first adding the new option to the list), while the latter allows setting the model to something that is not in the list (picking an option from the list only fills the model with the content of the option).
As a consequence of this an autocomplete can only be performed on text models and options.

docs/src/pages/vue-components/select.md
```diff
@@ -140,6 +140,12 @@ You can dynamically load new options when scroll reaches the end:
 
 ## Filtering and autocomplete
 
+### Understanding the difference
+
+The intended difference between a filtering QSelect and an autocomplete QSelect is offering the user a set of options that could be the full intended text to be entered (autocomplete), or removing options in the selection according to the input by the user (filter).
```
geoidesic, line 145: I think this is still confusing. Here's my suggestion:

`The intended difference between a filtering QSelect and an autocomplete QSelect is that autocomplete allows you  the model to accept a value that is not in the list of options, without adding it to the list. Filter on the other hand will result in a null or default value for the model if an option is not selected.`

smolinari: I think that would be good the technical part below. The first paragraph is what the user should expect, The second the developer.

Scott

rstoenescu:

Issue #7012, opened 2020-05-12
Background QPageSticky position shifts when QDialog showing


swallowtail62:
**Describe the bug**
Using `QPageSticky` with `QHeader` in `QLayout`,  when scroll down and then show `QDialog`, the background `QPageSticky` position shifts by the window.scrollY. 

**Codepen/jsFiddle/Codesandbox (required)**
https://jsfiddle.net/oxLn6gvd/

**To Reproduce**
1. Go to https://jsfiddle.net/oxLn6gvd/
![image](https://user-images.githubusercontent.com/45661101/81704738-23d88e00-94a9-11ea-8715-45b5e4bc2e95.png)

2. Scroll down to the `QBtn` which toggle `QDialog`
![image](https://user-images.githubusercontent.com/45661101/81704779-2f2bb980-94a9-11ea-8e1d-8466e6c46c59.png)

3. See error
![image](https://user-images.githubusercontent.com/45661101/81704793-34890400-94a9-11ea-8659-ccb9f55f6000.png)

**Expected behavior**
The background `QPageSticky` position does not shift.

**Platform (please complete the following information):**
OS: macOS Mojave
Browsers: Chrome 80.0.3987.163

pdanpdan:
Thanks for reporting. The fix will be available in Quasar v1.11.2

PR #7025, opened 2020-05-13
fix(QLayout): only update data on scroll when it's in container mode or the body is not scroll prevented #7012

pdanpdan:


rstoenescu:
We can't look into DOM on each scroll event. It's costly and the event is triggered often.

PR #7054, opened 2020-05-15
Massive Rewrite

supersexy:
I am extremely proud to push the Quasar documentation project to a new direction with this massive contribution of changes!


rstoenescu:
The new direction is preposterous, dangerous, and way out of line. You're banned. And your children's children. For 5 minutes.

Issue #7062, opened 2020-05-18
On android, nativ keyboard appears when selecting an item in q-select


Brice155:
**Describe the bug**
Only on mobile device (not reproduced in chrome dev tools), when an item is selected in a q-select, the nativ keyboard appears.

**Codepen/jsFiddle/Codesandbox (required)**
https://codepen.io/Brayce/pen/yLYxqMP

**To Reproduce**
1. Access the link with an android mobile
2. click on the q-select to show the options
3. select an option
4. See the nativ keaybords opening
It also can be reproduce on the quasar q-select home page.

**Expected behavior**
At step 3, the nativ keyboard should not be displayed because no input is requested.

**Screenshots**
No available screenshot

**Platform (please complete the following information):**
OS: 
Node:
NPM:
Yarn:
Browsers: Chrome Mobile v81
iOS:
Android: v10
Electron:

**Additional context**

pdanpdan:
Duplicate of #7061

rstoenescu:
Fix will be available in "quasar" v1.11.4.

Issue #7061, opened 2020-05-18
QSelect is pulling up an on-screen keyboard on capacitor Android


vycoder:
**Describe the bug**
QSelect is pulling up an on-screen keyboard on Android when it shouldn't. Only happens on `behavior="menu"`.

**Codepen/jsFiddle/Codesandbox (required)**
https://codepen.io/yev/pen/rNOQzMq

**To Reproduce**
1, Create a QSelect just like attached codepen above
2. Make sure behavior="menu"
3. Run in capacitor mode android.
4. Choose an option on QSelect

**Expected behavior**
No onscreen keyboard should be displayed.

**Platform (please complete the following information):**
OS: Pop!_OS 20.04
Node: 12.4.0
NPM: 6.14.5
Yarn: 1.22.4
Browsers: N/A
iOS: Not tested
Android: Yes
Electron: N/A
Quasar: 1.11.3

**Additional context**

rstoenescu:
Fix will be available in "quasar" v1.11.4.

slowaways:
Now I can't Select Option in Cordova! The fix generate another bug!

PR #7063, opened 2020-05-18
fix(QSelect): on mobiles the focus must be moved to a not input element #7061, #7062

pdanpdan:


rstoenescu:
Still not solved when first focusing on a QSelect with behavior menu on a mobile.

rstoenescu:
(Second example, select-part-6)

Issue #7129, opened 2020-05-29
QSelect - Unable to select option in mobile design using @filter + behavior="menu"


slowaways:
**Describe the bug**
Unable to select option on mobile design (Cordova, PWA mobile access, etc) when using QSelect with @filter + behavior="menu"

**Codepen/jsFiddle/Codesandbox (required)**
https://codepen.io/slowaways/pen/BaogVae?editors=1010

**To Reproduce**
\* If you are accessing from your mobile phone, skip to step 4.

1. Access the CodePen example on the link above
2. Toggle Device toolbar (CTRL + SHIFT + M) and set any mobile resolution
3. Refresh the page (F5)
4. Click in the QSelect and nothing happens when try to select options

**Expected behavior**
Open the menu with options.

**Platform (please complete the following information):**
OS: -
Node: -
NPM: -
Yarn: -
Browsers: Chrome, Firefox ...
iOS: ?
Android: Cordova
Electron: ?

**Additional context**
Version of Quasar: v1.12.0

I think it's because of this fix:
fix(QSelect): focus issues affect QMenu: https://github.com/quasarframework/quasar/commit/074d48f2eb31beba822cd1a85095acd10298492d

rstoenescu:
Fix will be available shortly in Quasar v1.12.1.

PR #7130, opened 2020-05-29
fix(QSelect): on mobile the keyboard should be hidden by focusing another element (not by blur) #7129

pdanpdan:


rstoenescu:
Gets into the previous loop issue on iOS

PR #7141, opened 2020-06-01
refactor(app): Various refactors

yusufkandemir:

- Refactor

- No

**Other information:**
Some colors might be different in the output, this change normally shouldn't exist in a refactor PR, but since that code is really old(pre-v1) and the code only runs for `Node < 10`, I don't it should be a big issue.

### DIFF
--- a/app/lib/node-version-check.js
+++ b/app/lib/node-version-check.js
@@ -1,4 +1,4 @@
-const { log, warn, fatal } = require('./helpers/logger')
+const { warn } = require('./helpers/logger')
 
 const version = process.version.split('.')
 const major = parseInt(version[0].replace(/\D/g, ''), 10)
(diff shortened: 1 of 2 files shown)


app/lib/node-version-check.js
```diff
@@ -1,13 +1,15 @@
...
 const major = parseInt(version[0].replace(/\D/g,''), 10)
 // const minor = parseInt(version[1].replace(/\D/g,''), 10)
 
 if (major < 10) {
-  console.warn('\x1b[41m%s\x1b[0m', 'INCOMPATIBLE NODE VERSION')
-  console.warn('\x1b[33m%s\x1b[0m', 'Quasar CLI requires Node 10.0.0 or superior')
-  console.warn('')
-  console.warn('⚠️  You are running Node ' + version)
-  console.warn('Please install a compatible Node version and try again')
+  fatal('\x1b[41m%s\x1b[0m', 'INCOMPATIBLE NODE VERSION')
```
rstoenescu: fatal() also exits the process, so anything after it won't still be displayed...

yusufkandemir: I have already noticed and fixed it, so this is outdated.

PR #7183, opened 2020-06-08
fix(bex): Don't minify rendered code.

webnoob:


webnoob:
Cancelled. Moving on to qapp/v2

PR #7192, opened 2020-06-09
add email validator function to validation patterns

erfanSW:
- add email validtor for q-input


pdanpdan:
Hello. This is a can of worms (check google, 30 years of history of discussions about this).
And this specific regex does not even pass the basic formats

```
simple@example.com
very.common@example.com
other.email-with-hyphen@example.com
fully-qualified-domain@example.com
user.name+tag+sorting@example.com
x@example.com
example-indeed@strange-example.com
admin@mailserver1
example@s.example
" "@example.org
"john..doe"@example.org
mailhost!username@example.org
user%example.com@example.org
disposable.style.email.with+symbol@example.com
```

on the last line the regex even goes berserk.

https://regex101.com/r/fo4lbQ/1

PR #7266, opened 2020-06-21
fix(babel-preset-app): Pinned the correct minimum babel versions

yusufkandemir:

- Bugfix

- No

**Other information:**
`bugfixes` option is introduced in Babel v7.9.0, please see this [blog post about the release](https://babeljs.io/blog/2020/03/16/7.9.0). But, even though the current pinned version is 7.5.0(as can be seen [here](https://github.com/quasarframework/quasar/blob/955860e6c5958316bedf6e8353561700a71b4012/babel-preset-app/package.json#L10-L24)), that option is still used(see [here](https://github.com/quasarframework/quasar/blob/955860e6c5958316bedf6e8353561700a71b4012/babel-preset-app/index.js#L73)).

If the package manager decides to keep and use the older versions of these dependencies because it's still valid according to that version specification, this error would happen. It's probably rare, but I stumbled upon it while upgrading to `@quasar/app` v2.
```
Error: [BABEL] /.quasar/client-entry.js: Invalid Option: bugfixes is not a valid top-level option.
Maybe you meant to use 'modules'? (While processing: "/node_modules/@quasar/babel-preset-app/index.js.overrides[0]$0")
```


rstoenescu:
This will go into "@quasar/app" v2.0.1.

PR #7281, opened 2020-06-23
feat(position engine): improve positioning by waiting for the animation/transition to finish

pdanpdan:


rstoenescu:
There's two issues with this:
1. When you don't use any transition (or a bogus transition name) the handler won't be called, but still remain attached. 
2. It seems to me that the position gets updated on transition end anyways.. Am I missing something?

pdanpdan:
Now it's running after 300ms, which is most often too early.

Issue #7290, opened 2020-06-24
feat(QDate): Improve the event highligting system


yusufkandemir:
**Is this feature request related to a problem?**
Currently, highlighting events is limited to just indicating if there is an event or not, and modifying the dot's color. It could be improved to offer more customizations such as changing the class of the dot, adjust the color in place(without another prop), popovers that contain descriptions of events, and more.

**The solution I'd like**
`events` and `eventsFn` props could accept an array of objects while keeping the current behavior(array of strings). Similar approaches are currently being used in other components(e.g. [`q-select` > options](https://quasar.dev/vue-components/select#Options-type)).

**Alternatives I've considered**
This could be done in devland if #7076 is implemented in a way that allows this.

**Additional context**
A Vue plugin which has the described features: [V-Calendar](https://vcalendar.io/attributes.html#highlights)

pdanpdan:
It boils down to a duplicate of https://github.com/quasarframework/quasar/issues/7076

Issue #7076, opened 2020-05-20
QDate: provide slot for individual days in month view


qyloxe:
**Is your feature request related to a problem? Please describe.**
There is a events property which allows to mark individual days in month view with colored underline. Unfortunately I would prefer to have a different mark for the event, Maybe even a badge with number of events in this particular day.

**Describe the solution you'd like**
If there would be available a specific slot in QDate component, for rendering of individual days, I could make my events or days visualisation as I wish. Heck, I could even capture mouseenter events and display some kind of hint, tip i popup when user moves the cursor over specific day. I could even put in some days other QComponents.

**Describe alternatives you've considered**
The only viable alternative as of today to display different day rendering and capturing mouse events on particular days is just using another component from another library or single Vue component. The available app extension QCalendar is too big for using - for example I need to just display different colors for particular days in UMD mode, where every kb counts.

**Additional context**
This doesn't seem as overly complicated, but it could IMHO dramatically improve quality of life and possible use cases for Quasar as a machine for presentation data based on time.

webnoob:
New slot available in 1.12.9

webnoob:
Unfortunately, I'm going to have to revert this commit. I've had a ping from Razvan about it. Will try and get more info going forward.

AruXc:
Any new info about this?

ttquoccuong:
Hi @webnoob ,
Any plan to re-add this.
It's very helpfull.
Thanks!

hawkeye64:
Can this issue be closed?

PetrPy:
Implementing the slot would turn the QDate almost into a full-featured calendar component that is missing in Quasar.

ddenev:
@webnoob @pdanpdan 

is there any progress on this request? I see that you have reverted a commit some long time ago and then this was stalled. Do you think this could be implemented?

Thank you!

akahanek:
@webnoob @pdanpdan 
This feature would be really extremely helpfull. Any progress on this please?

ddenev:
@pdanpdan, there are several hundred commit messages in this thread - does this mean you are working on this request? Thank you!

pdanpdan:
This is done for a long time. But it's here just for reference, it's not going to be merged. If you use Q1 and you need this you can use my builds (https://pdanpdan.github.io/quasar-docs)

ddenev:
@pdanpdan, thanks for the reply!

[...]

PR #7293, opened 2020-06-24
feat(QDate): add event rendering function #7076, #7290

pdanpdan:


rstoenescu:
Sorry, but cannot accept this PR. From a conceptual point of view, this is what slots should be used for. The majority is not well-versed with render functions nor should they be as long as there are .vue files and vue-loader.

pdanpdan:
I agree about slots, but I don't see a good way to create slots for days.

rstoenescu:
It should be slots for the content of days at best. You add the day number, an additional absolute positioned div, a QMenu or QTooltip in it. But that should be it.

PR #7297, opened 2020-06-24
feat(app-extension): ability to exclude ts or js files from rendering

stefanvanherwijnen:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

With the new typescript support, the app extensions will have to also support TS files.
If I am correct, this would require both a .js and .ts file in the app extension templates folder as there is no way to combine these two in a single source file.
The render() function renders all files in the template directory which places both js and ts files in the app, which will result in an error.

Excluding the files with fglob seemed to be the most simple solution, but I am not entirely sure if this is the best approach.

Let me know if this is accepted and I will update the docs accordingly.

### DIFF
--- a/app/lib/app-extension/Extension.js
+++ b/app/lib/app-extension/Extension.js
@@ -55,11 +55,16 @@ async function renderFile ({ sourcePath, targetPath, rawCopy, scope, overwritePr
   }
 }
 
-async function renderFolders ({ source, rawCopy, scope }) {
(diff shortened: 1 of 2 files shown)


stefanvanherwijnen:
https://github.com/quasarframework/quasar/issues/7329

PR #7303, opened 2020-06-25
QSelect: Differentiate between mutiple selection and singular via class

dmitry:
This allows to differentiate between multiple and singular QSelect, so it possible to add styles for a specific case.


rstoenescu:
This will go into Quasar v1.13.2.

Issue #7284, opened 2020-06-23
QStep header-nav always true for current step


mv-go:
**Describe the bug**
Header navigation seems to always be true (active) on a current `q-step`. I attempted to set linear navigation following the example from docs (links below) and found that current step would always have `header-nav` enabled on it, unless `header-nav` is set to `false` on parent `q-stepper` (which disables header navigation for all steps, which defeats the purpose).

**Codepen/jsFiddle/Codesandbox (required)**
Can be found right in the docs here https://quasar.dev/vue-components/stepper#Example--Linear-header-navigation
Direct link to CodePen https://codepen.io/pen/?&editable=true&editors=101=https%3A%2F%2Fquasar.dev%2Fvue-components%2Fstepper

**To Reproduce**
Set `header-nav: true` to q-stepper
Set `header-nav: false` to all q-steps
All steps will have header-nav disabled except for the current one

**Expected behavior**
Header-nav on q-step to always override header-nav on q-stepper.

**Platform (please complete the following information):**
any

**Additional context**
Maybe we could go further and provide some sort of validation Fn to `q-stepper` `header-nav` prop, similar to the `q-input` component approach, like so
`:headerNav="isEnabled"`
...and later in methods...
`isEnabled (step) { return this.currentStep > step }`

webnoob:
I've read this 5 times and I don't understand the issue properly.

Are you saying that the current QStep shouldn't be clickable? What problems is this causing?

mv-go:
My apologies for not being clear on this one.

You are basically correct. Current step (just as any other step) should not be clickable if it has its `:header-nav="false"`.
Current behaviour is, in my humble opinion, somewhat misleading.

Maybe this will help:

Have `q-stepper` with `header-nav`.
Have one of the `q-step`s' `:header-nav="false"` (or any condition that returns false)
This step will still be clickable and have hover effects (so essentially have its header-nav enabled) when it's "current".

Demonstrated in this codepen (slightly modified 'Linear Header Navigation' example from quasar docs). Step 2 has `:header-nav="false"`, and step 3 has `:header-nav="step > 4"` (which is ultimately false as well), but both of these steps are coverable and clickable when they are "current". https://codepen.io/mv-go/pen/xxZPZJr?editable=true&editors=101%3Dhttps%3A%2F%2Fquasar.dev%2Fvue-components%2Fstepper

[...]

PR #7333, opened 2020-06-30
fix(QStep): headerNav=false, current step shouldn't be clickable #7284

webnoob:
I've tested QStepper and QCarousel and this doesn't seem to have any negative effects. I've looked back in the git history it seems it's always been like this.


webnoob:
@pdanpdan Can you take a look at this please and see if you can think of a reason *not* to make this change. I can't see based on the history that it was added for a reason and the way it is in the current version doesn't make much sense to me.

pdanpdan (approved):
I think it's ok. I cannot find something that this should break.

Issue #7350, opened 2020-07-04
Some components force their parent to update if they have @click (or other) listeners attached


ddenev:
I'm not sure if this is a bug but the behavior is strange to begin with.

Some Quasar components force their parent to update (re-render) if they have a `@click` listener attached.

**Codepen/jsFiddle/Codesandbox (required)**
https://jsfiddle.net/ddenev/gj78c3r4/18/

**To Reproduce**
Just reload and observe the console. If you remove the `@click` attribute then the parent is not updated.
Comment/uncomment as necessary to test with different components.
I tested with QBtn, QIcon, QCard and QChip. Strangely - this doesn't happen with QChip but happens with the others.

**Expected behavior**
I expect that the parent component is not updated after its first render just because its template contains a Quasar component that has a `@click` handler (or any other, haven't tested with all) attached.

pdanpdan:
Thanks for catching this.

pdanpdan:
this was fixed a long time ago, forgot to close.

PR #7354, opened 2020-07-05
perf(cache): create the original cached object in data to avoid an update of the parent component #7350

pdanpdan:


rstoenescu:
Enhancement will be available in Quasar v1.13.0.
Thank you for reporting!

PR #7361, opened 2020-07-06
chore(docs): Set default eslint-plugin-quasar version to ^1.0.0

hawkeye64:


webnoob:
@hawkeye64 I assume this new version doesn't contain any breaking changes etc? Is there anything I'll need to test before merging this (other than docs actually loading)

hawkeye64:
@webnoob v1 eslint-plugin-quasar has been out for a while now (May 24). I just forgot to get it updated and wasn't sure if npm/yarn would update it based on semantic rules, if original had "beta" in it. All my app-ext demos use v1.

PR #7386, opened 2020-07-09
fix(QSelect): guard keyboard search for index -1 (initial state when no option selected); allow space in search if it's not first; allow search for any printable char #7382

### LINKED ISSUE #7382: QSelect throws typeError when typing in a focused state
**Describe the bug**
We use a lot of arrays with objects in combination with `map-options` in QSelects, in order do translate them we use `i18n` translation functions. Basically we do this: `:option-label="item => translationFn(item.label)"`.

However this seems to cause some issues, when selecting the select the menu opens. You can type and select an option without problems. The menu will close and the select is still focused, now if you start typing you will get `typeErrors` (can not read property of item).

**Codepen**
https://codepen.io/lajuffermans/pen/PoZeQRe

**To Reproduce**
1. Click the select
2. Start typing, select an option (menu/select will close)
3. Start typing again, menu will open, option won't be selected

**Expected behavior**
No errors and a selected option

**Screen recording**
[Screen Recording 2020-07-09 at 09.38.33.zip](https://github.com/quasarframework/quasar/files/4895238/Screen.Recording.2020-07-09.at.09.38.33.zip)

**Screenshot**
Screenshot 2020-07-09 at 09 40 52

pdanpdan:


webnoob:
Playing with: http://localhost:8080/form/select-part-5-kbd

When tabbing through the new selects you've added, search for a value to pop the options. Hitting enter doesn't select it.

Is this an issue or expected?

pdanpdan:
I'll have to check, it's not ok.

pdanpdan:
Fixed, a missing === 1 :(

webnoob:
Working now, thanks.

PR #7394, opened 2020-07-09
docs(QEditor): Document fixedLabel, fixedIcon and highlight #7380

### LINKED ISSUE #7380: [docs] fixedLabel and fixedIcon not documented
Version: 1.12.8 (latest at this moment)

As just [noted on discord](https://discord.com/channels/415874313728688138/596275881907978240/730500246584426607) I wanted to hide the label in dropdowns for `q-editor` then realizing that `fixedLabel: true` would do the trick. But this solution was based on trying things by looking at existing examples as `fixedLabel` is actually not documented in the API section itself (nor  `fixedIcon` for that matter): https://quasar.dev/vue-components/editor#QEditor-API

>(btw, thanks again for all the great work you all quasarians are doing!)

webnoob:


webnoob:
@smolinari Requested a review for this one as you might have a better way to explain the props.

smolinari (approved):
Looks ok to me. Just commas missing before the "so"s and "when", but that is getting picky. :)

Scott

PR #7434, opened 2020-07-16
fix(app): CopyWebpackPlugin - ignore options not working #7433

### LINKED ISSUE #7433: [CLI] - CopyWebpackPlugin - ignore options not working
When I was a looking for a way how to add custom folder to be ignored when building the app with `quasar build` I discovered that Quasar code handling icons for SPA and Electron doesn't work ([Discord chat](https://discord.com/channels/415874313728688138/601046346602119168/732903815564623872)).

It shouldn't copy icons when ['electron', 'cordova', 'capacitor'] is used. [link to Quasar DEV CLI code- CopyWebpackPlugin - ignore](https://github.com/quasarframework/quasar/blob/6805244a67e539e2b5e1ff1cde74e8db3849e689/app/lib/webpack/create-chain.js#L322)
```
const patterns = [{
        from: appPaths.resolve.app('public'),
        to: '.',
        noErrorOnMissing: true,
        globOptions: {
          ignore: [ appPaths.resolve.app('/**/.*') ].concat(
            // avoid useless files to be copied
            ['electron', 'cordova', 'capacitor'].includes(cfg.ctx.modeName)
              ? [ appPaths.resolve.app('public/icons'), appPaths.resolve.app('public/favicon.ico') ]
              : []
          )
        }
      }]
```
### Result for Electron
```
 Build mode........ electron
 Pkg quasar........ v1.12.11
 Pkg @quasar/app... v2.0.5
 Debugging......... no
 Publishing........ no
```
image

### Result for SPA
```
 Build mode........ spa
 Pkg quasar........ v1.12.11
 Pkg @quasar/app... v2.0.5
 Debugging......... no
 Publishing........ no
```
image

Problem is in path definition (inspiration from [copy-webpack-plugin issue](https://github.com/webpack-contrib/copy-webpack-plugin/issues/498)):
`[ appPaths.resolve.app('public/icons'), appPaths.resolve.app('public/favicon.ico') ]` should be probably changed to ` ['**/public/icons', '**/public/favicon.ico']` - after changes it copy the folders only for SPA.
```
 const patterns = [{
        from: appPaths.resolve.app('public'),
        to: '.',
        noErrorOnMissing: true,
        globOptions: {
          ignore: ['/**/.*'].concat(
            // avoid useless files to be copied
            ['electron', 'cordova', 'capacitor'].includes(cfg.ctx.modeName)
              ? ['**/public/icons', '**/public/favicon.ico']
              : []
          )
        }
      }]
```
And the result:
image

I'll be happy If you could provide an example how to correctly inject my custom ignore folder in `quasar.conf.js` instead of hacking source `create-chain.js` file.

webnoob:


app/lib/webpack/create-chain.js
```diff
@@ -322,7 +322,7 @@ module.exports = function (cfg, configName) {
           ignore: [ appPaths.resolve.app('/**/.*') ].concat(
```
yusufkandemir: This pattern must also be fixed. Also, dotfiles are ignored by default(can be enabled with `dot: true`), so we can drop this pattern instead of fixing it because it is redundant.

app/lib/webpack/create-chain.js
```diff
@@ -322,7 +322,7 @@ module.exports = function (cfg, configName) {
...
             // avoid useless files to be copied
             ['electron', 'cordova', 'capacitor'].includes(cfg.ctx.modeName)
-              ? [ appPaths.resolve.app('public/icons'), appPaths.resolve.app('public/favicon.ico') ]
+              ? [ '**/icons/**/*', '**/favicon.ico' ]
```
yusufkandemir, line 325: This can be simplified. Also, even though the chances are low, deep matching the icons directory could result in unwanted results.
```suggestion
              ? [ 'icons/**/*', 'favicon.ico' ]
```

yusufkandemir (changes requested):
It looks good, I left some comments for little improvements.

app/lib/webpack/create-chain.js
yusufkandemir: Ignore the first sentence, you have already done it 🚀.

webnoob: Good point about `dot: true` - I had forgotten. Going to update as it's a better fix.

app/lib/webpack/create-chain.js
webnoob: That doesn't work. How I have done it is the suggested way I've seen online.

yusufkandemir: See: https://github.com/mrmlnc/fast-glob#ignore

app/lib/webpack/create-chain.js
webnoob: p.s It's `dot: false` not true.

app/lib/webpack/create-chain.js
webnoob: For anyone else who reads: after discussion in discord, it needs to be as it is now. Glob options are strange with copy-webpack plugin.

webnoob: For info:

https://github.com/webpack-contrib/copy-webpack-plugin/blob/21d9c7d3217c472bc195ec7461a3b26e1360c886/src/utils/createPatternGlob.js#L30-L33

it appends **/*
so it's like there are multiple inputs
so we can't just ignore folder/*
but **/folder/* works

So, if we use glob matching in from instead, we could make it relative.
But I don't think it's required at this point. The fix is inline with docs and works

Issue #7441, opened 2020-07-16
Icon Genie v2.3.0 profile generator incorrectly checking for hash in splashscreenColor hex value


taliator:
**Describe the bug**
Icon Genie profile generator incorrectly checking for hash in splashscreenColor hex value, and won't accept a hex value without a hash.

**Codepen/jsFiddle/Codesandbox (required)**
N/A

**To Reproduce**
1. Try to generate an IconGenie profile (replacing with relevant filenames):
    `icongenie profile -o foo -a spa,capacitor -i app-icon.png -b app-splashscreen.png --quality 2 --splashscreen-color 286DD7`
2. See error where matching pattern expects the hash:
    ![image](https://user-images.githubusercontent.com/1482924/87695532-9726a680-c75d-11ea-9595-4a55d0088376.png)
3. Now try to generate a profile with the hash:
    `icongenie profile -o foo -a spa,capacitor -i app-icon.png -b app-splashscreen.png --quality 2 --splashscreen-color #286DD7`
    ![image](https://user-images.githubusercontent.com/1482924/87695684-cccb8f80-c75d-11ea-9170-17de53a8fc1b.png)
5. Now escape the hash
    `icongenie profile -o foo -a spa,capacitor -i app-icon.png -b app-splashscreen.png --quality 2 --splashscreen-color \#286DD7`
6. Profile generates successfully, so try to generate icons:
    `icongenie generate -p ./icongenie-foo.json`
7. See error (obviously, since the profile should *not* include the hash)
    ![image](https://user-images.githubusercontent.com/1482924/87695922-203ddd80-c75e-11ea-9471-fe756958e17f.png)

**Expected behavior**
The profile should generate without the hash prepending the color (remove the hash from the pattern matching check).

**Screenshots**
See above.

**Platform (please complete the following information):**
OS: MacOS Catalina 0.15.5
Node: v14.2.0
NPM: 6.14.4

**Additional context**
N/A

IlCallo:
Have you tried as `--splashscreen-color=286DD7` too?

taliator:
@IlCallo I had not, but I just ran
`icongenie profile -o foo -a spa,capacitor -i app-icon.png -b app-splashscreen.png --quality 2 --splashscreen-color=286DD7`
 and it returns the same error, because the regex is still checking for a hash:

![image](https://user-images.githubusercontent.com/1482924/87784998-626e2a00-c805-11ea-8d35-f2ed649baf28.png)

[...]

PR #7463, opened 2020-07-20
fix(icongenie): Incorrect validation when generating profile #7441

webnoob:


icongenie/lib/utils/validate-profile-object.js
```diff
@@ -21,13 +21,22 @@ const paramsSchema = Joi.object({
     Joi.number().integer().min(0)
   ).min(1).max(2),
 
+  splashscreenIconRatio: Joi.number().integer().min(0).max(100)
+}
+
+const colorParamsSchemaProfileSave = {
+  themeColor: Joi.string().pattern(/^[0-9A-Fa-f]{3}([0-9A-Fa-f]{3})?$/),
```
yusufkandemir, line 28: We can create a function to createPatterns, this will simplify this a bit.
```js
const createColorPattern = requiresHash => new RegExp(`^${requiresHash ? '#' : ''}[0-9A-Fa-f]{3}([0-9A-Fa-f]{3})?$`)
```

yusufkandemir (approved):
👍

icongenie/lib/utils/validate-profile-object.js
```diff
@@ -67,18 +53,30 @@ const assetsSchema = Joi.array().items({
...
+ * When generating the profile file, we don't want to validate with # on the hex color.
+ * When generating the icon, we're expecting a hash on the color (automatically added to user input via the CLI)
+ */
+const getColorParamsSchema = (requireHash) => {
+  const colorPattern = Joi.string().pattern(new RegExp(`^${requireHash ? '#' : ''}[0-9A-Fa-f]{3}([0-9A-Fa-f]{3})?$`))
+  return {
+    themeColor: colorPattern,
+    pngColor: colorPattern,
+    splashscreenColor: colorPattern,
+    svgColor: colorPattern,
```
yusufkandemir, line 66: _style:_
I guess we don't use dangling commas 🙃

yusufkandemir (approved):
It looks even better now 🔥. I left a little comment.

icongenie/lib/utils/validate-profile-object.js
yusufkandemir: I guess it has been slipped through when copy/pasting from the middle of the old pattern object.

webnoob: Ah whoops. Will sort. Nice spot.

PR #7492, opened 2020-07-25
feat: updated fontawesome, added feather icons (svg), updated svg format for QIcon

hawkeye64:
- Updated Fontawesome to v5.14.0
- Added Feather Icons (SVG only)
- Updated Extras SVG generation and updated quasar QIcon to handle new format
Here's a few things I learned:
1. If there were any attributes on the svg element itself, these were ignored: fixed
2. The svg generation was only looking for "styles" on child elements. Any attributes were ignored: fixed
3. Svg generation was using 'z' or 'Z' to close a segment ('M' starts a segment). It took a long time with help from @jvillanueva to figure out this was the issue and caused a lot of the feather icons to draw incorrectly. After looking around, there were issues with icons in other icon sets. For instance, the "airplay" icon is a TV with a triangle at the bottom. The triangle if on top of the TV. It should cause the triangle to have an edge around it that causes this segment ot appear not to touch the TV. Other icons that were arrows, were caused to have the end of the arrow (the V) filled in. Removing the 'Z' and 'z' fixed all of these. This was pretty much a days worth of work.
4. Taking attributes from the svg element and turning them back into attributes didn't work. '.q-icon' class caused issues. overriding the attributes. Turning the attributes into a "style" solved the override.

Images of issues resolved:
image
image
image
image
image
image


hawkeye64:
This is ready for review/merge

hawkeye64:
Release notes:
```
@quasar/extras:
- update: fontawesome-v5 upgraded to v5.14.0
- feat: feather icons (svg only)
- fix: improvements to SVG parsing

quasar:
- feat: svg-feather-icons icon set
- fix: improvements to QIcon handling "SVG icon format" (https://quasar.dev/vue-components/icon#Svg-icon-format)
```

ui/src/components/icon/QIcon.js
```diff
@@ -60,8 +62,8 @@ export default Vue.extend({
         }
       }
 
-      if (icon.startsWith('M') === true) {
-        const [ def, viewBox ] = icon.split('|')
+      if (icon.startsWith('M') === true || icon.startsWith('m') === true) {
```
rstoenescu, line 65: An svg starting with "m" is kind of pointless ("m" is for "move to position relative to current one", where current one is always 0,0). Let's try to make all of them start with "M" to avoid one more "if" statement. QIcon needs to be rendered as fast as possible as its probably the most used one in the entire framework.

ui/src/components/icon/QIcon.js
```diff
@@ -172,13 +175,22 @@ export default Vue.extend({
       data.attrs.focusable = 'false' /* needed for IE11 */
       data.attrs.viewBox = this.type.viewBox
 
+      // append attrs as style
+      if (this.type.attrs !== undefined) {
+        console.log(this.type.attrs)
+        let attrs = []
+        while ((attrs = REGEX_ATTRS.exec(this.type.attrs)) !== null) {
+          data.style[attrs[1]] = attrs[2]
```
rstoenescu, line 183: It would be better if we just read the style as a String and assign it as such (data.style = styleAsString), rather than splitting them individually.

rstoenescu (changes requested):
Excellent work! Got a few comments though.

ui/src/components/icon/QIcon.js
hawkeye64: That was the initial attempt. However, when used as "styles=" the QIcon styling overwrote it. I have a video I can show you with the issue.

ui/src/components/icon/QIcon.js
hawkeye64: I was more concerned about people using custom. A couple of the test icon packages I was using had lower-case 'm' and had issues (wouldn't render). That because the path was grabbed without change. I did modify that in the build code, so I am fine reverting this back to just 'M'.

extras/build/utils/index.js
```diff
@@ -143,14 +183,14 @@ function parseSvgContent(name, content) {
   }
 
   const result = {
-    viewBox: viewBox !== '0 0 24 24' ? `|${viewBox}` : ''
+    viewBox: viewBox !== '0 0 24 24' ? `|${viewBox}` : '|',
+    svgStyle: svgStyle !== '' ? `|${svgStyle}` : '|'
```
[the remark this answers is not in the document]
hawkeye64: How to determine if it is viewBox or attributes then?

docs/src/boot/components.js
```diff
@@ -7,6 +7,18 @@ import DocInstallation from 'components/DocInstallation.vue'
...
+import SvgMaterialIconsRound from 'components/svg/SvgMaterialIconsRound'
+import SvgMaterialIconsSharp from 'components/svg/SvgMaterialIconsSharp'
+import SvgMdi from 'components/svg/SvgMdi'
+import SvgIonicons from 'components/svg/SvgIonicons'
+import SvgEvaIcons from 'components/svg/SvgEvaIcons'
+import SvgFontawesome from 'components/svg/SvgFontawesome'
+import SvgLineAwesome from 'components/svg/SvgLineAwesome'
+import SvgThemify from 'components/svg/SvgThemify'
+import SvgFeatherIcons from 'components/svg/SvgFeatherIcons'
+
```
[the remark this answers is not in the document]
hawkeye64: I will make a new PR

PR #7533, opened 2020-07-30
chore(typescript): exclude src-cordova to shorten build time

qianbin:

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

When the quasar project using typescript contains cordova mode, it takes long time in webpack build stage. and sometimes build process exits with the error like"Type checking and linting aborted - probably out of memory. Check `memoryLimit` option in ForkTsCheckerWebpackPlugin configuration."

### DIFF
--- a/app/tsconfig-preset.json
+++ b/app/tsconfig-preset.json
@@ -29,5 +29,5 @@
   },
   // Needed to avoid files copied into 'dist' folder (eg. a `.d.ts` file inside `src-ssr` folder)
   // to be evaluated by TS when their original files has been updated
-  "exclude": ["/dist", ".quasar", "node_modules"]
+  "exclude": ["/dist", ".quasar", "node_modules", "src-cordova/platforms", "src-cordova/www"]
 }


IlCallo:
The same happens with Capacitor and also other modes (eg. electron).
For example, Capacitor would need something link this:

```json
{
  "extends": "@quasar/app/tsconfig-preset",
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "src/*": ["src/*"],
      "app/*": ["*"],
      "components/*": ["src/components/*"],
      "layouts/*": ["src/layouts/*"],
      "pages/*": ["src/pages/*"],
      "assets/*": ["src/assets/*"],
      "boot/*": ["src/boot/*"],
      "*": ["src-capacitor/node_modules/*"]
    }
  },
  "exclude": [
    "/dist",
    ".quasar",
    "node_modules",
    "src-capacitor/android",
    "src-capacitor/ios"
  ]
}
```

But adding all the configurations for all modes in the preset could cause some problems in the long run.
I was thinking about providing a different preset for all modes, but the underlying issue is that Quasar "modes" haven't been properly adapted to work with TS yet.

Until we start quasar v2, which should place the foundations to a progressive migration of the codebase to TS (nothing written in stone yet), I'd prefer to keep this on hold

qianbin:
well, how about config 'include' instead of 'exclude'? in my past practice, 'include' was always preferred.

IlCallo:
@qianbin unluckily switching to `include` won't resolve the main problem: `include` and `exclude` both aren't "mergeable" when the tsconfig preset get extended, so they must be replicated in the project tsconfig.
The same problem applies to `compilerOptions.paths` for some reason (even if there the value is a nested object and could theoretically be merged).

Yet, I saw they recently added support for ["Solution style" configuration](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#support-for-solution-style-tsconfigjson-files) and this could be the holy grail we have been waiting for.
Will keep exploring in this area

rstoenescu:

Issue #7520, opened 2020-07-29
Using preFetch to mutate vuex causes errors across page components


pianopronto:
**Describe the bug**
I am using preFetch to call an API and populate vuex with the information each page will need. Each page's preFetch call to the API returns an object that overwrites the old page's object in vuex. In other words, the page object in vuex contains all the information for that page only.

When I mutate the store and replace the object, it happens before the router-view component changes, causing errors where the old page tries to access data that is no longer there. Specifically, computed properties in the old page are firing since the store is mutated before the old page is unmounted.

**Codepen/jsFiddle/Codesandbox (required)**
Basic github reproduction example, instead of using an API, I simply call the vuex mutation in prefetch directly. https://github.com/pianopronto/quasar-prefetch-test

**To Reproduce**
1. Clone repo
2. Run quasar dev
3. Open dev tools
4. Switch between the pages using the links in the drawer

**Expected behavior**
There should be someway to mutate vuex in a preFetch hook after the old component has unmounted. When using a non async preFetch, nextTick does work (see topic-next-tick branch) but when using an async prefetch that workaround doesn't work.

**Screenshots**
![image](https://user-images.githubusercontent.com/51970393/88827167-73605900-d17e-11ea-8434-7f512fc1a837.png)

The above shows the errors when running the master branch of the test repo, and switching from the Index page to the Other page. I put in logging statements in the preFetch hooks and router hooks. You can see after "finishing prefetch in other page" when the vuex mutation happens, the computed handler runs in the index page.

**Edit:** Added a few more console.log statements in each component's beforeMount, mounted, beforeDestroy, destroyed hooks just to give a better idea what's happening.

**Platform (please complete the following information):**
OS: Ubuntu 18.04
Node: v12.14.1
NPM: 6.13.4
Yarn: 1.21.1
Browsers: Chromium 83
iOS: n/a
Android: n/a
Electron: n/a

**Additional context**
Am I going about this the wrong way? To give more context, each page has vastly different info that needs to be stored in vuex.

pdanpdan:
Hello, this is expected because that's the lifecycle of vue-router. You should check in your component of you have the right data. Or just don't use preFetch and use a Vue-router hook that suits you.

pianopronto:
Thanks @pdanpdan, you're right, vue-router mounts the new component before destroying the old one. I'll simply improve the computed properties to make sure the expected data is there. Thanks for the quick answer and confirming this can't be an issue with Quasar.

[...]

PR #7547, opened 2020-08-01
feat(preFetch): add router instance to the parameters received by the function #7520

pdanpdan:


rstoenescu:
It looks like a good idea, but it really isn't. Devs can use it to navigate to a different route, essentially by-passing the redirect() param which can lead to a mess. And this is just the tip of the iceberg.

Issue #6714, opened 2020-03-31
problem with RTL code generation


ghost:
**Describe the bug**
Hello. according to the well written Quasar documents, to support RTL mode, language pack and RTL flag must be added to the config file.
as i was reading this repository issues I found out that 'rtlcss' module was taking responsibility for adding RTL converted code and i couldn't find anyway to configure it. i am developing a first RTL website (mainly rtl with no ltr support in mind) and i dont need this module to generate any code for me. but i cant set the RTL flag ( in the config file ) to false, because the components visual behavior depend on it. there is a way to disable one css rule at a time but, adding this "ignore rule" to every line of the code is frustrating.
is there anyway to disable this code generation but still keep the RTL behavior of the components ? 
Thanks in advance :)

**To Reproduce**
1. Follow [Quasar RTL support](https://quasar.dev/options/rtl-support#Enabling-RTL-support) guide
2. add margin-left to an element (developing a RTL website)
3. margin-left is converted to margin-right while it was not necessary

**Expected behavior**
There should be an option to make rtlcss module, stop generating unwanted code OR providing a new Object as the config for rtl support in Quasar.config for additional control over this convertion.

pdanpdan:
Hello, sorry to be so late with the reply.
It would help us very much if you could do a research on rtlcss configuration options and find a way to disable it for your specific case. That would allow us to implement the required changes at framework level.
Thank you.

Saeid-Za:
Hello, no problem at all, thanks for replying.
Actually I did some research on [rtlcss](https://rtlcss.com/) dependency that is being used in quasar source code. I will try to present my findings on the matter.
### Solution 1 
It appears that quasar is depending on [rtlcss postcss plugin](https://github.com/vkalinichev/postcss-rtl) rather than rtlcss itself. after reading the rtlcss and the postcss plugin documentation, I came with an idea, Ignore every lines of CSS that should not be transformed as [documented in rtlcss](https://rtlcss.com/learn/usage-guide/control-directives/#ignore) and [quasar](https://quasar.dev/options/rtl-support#Marking-CSS-rules-as-exceptions) itself.

code snipped to ignore rules as documented in RTLCSS:

```
.code {
  /*rtl:ignore*/
  direction:ltr;
  /*rtl:ignore*/
  text-align:left;
}
```

code snipped to ignore rules as documented in Quasar (SCSS version):
```
.my-class {
  margin-left: 10px #{"/* rtl:ignore */"};
}
```

[...]

PR #7592, opened 2020-08-07
feat(app): allow client CSS transform as fromRTL when using RTL #6714, #7488

pdanpdan:
Usage: set rtl to { fromRTL: true } in quasar.conf


Saeid-Za:
Hello, i was wondering when would this feaure be merged? looking forward to it.
Thanks!

PR #7671, opened 2020-08-24
feat(QTime): Add grid snapping to QTime with options

thexeos:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

The current dragging (and clicking) behavior when using `*-options` for `options-fn` leads to confusing (from user's point of view) behavior. They may be dragging the cursor over the number, or clicking on the number, yet the time would not get selected.

current-behavior

current-behavior-still

Code-wise, this of course makes sense, for example on the above screenshot, the user is clearly dragging over the 29th minute, which is not allowed by `minute-options` and so no update is performed.

This PR adds snapping to the update function. It only activates if the developer had limited user selection. For both dragging and clicking, the computed value from the cursor position snaps to the nearest allowed value. More on implementation below. This is how it behaves with this PR:

proposed-behavior

The implementation relies on mapping values after they were converted from degrees into time units (hours, minutes, seconds). The computed value contains an array with 24 elements for hours and 60 elements for minutes and seconds. Element index is the value obtained from cursor position and element value is the snapped to value.

For example, if we allow minutes to be set in increments of 5, our `minute-options` would be equal to `[0, 5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55]`. The `minuteSnappingGrid` would then be `[0, 0, 0, 5, 5, 5, 5, 5, 10, 10, 10, 10, 10, 15, 15, 15, 15, 15, 20, 20, 20, 20, 20, 25, 25, 25, 25, 25, 30, 30, 30, 30, 30, 35, 35, 35, 35, 35, 40, 40, 40, 40, 40, 45, 45, 45, 45, 45, 50, 50, 50, 50, 50, 55, 55, 55, 55, 55, 0, 0]`. Now, if the user was to click on *4* minute sector on the clock dial, we would lookup the 4th element in the array, giving us the snapped value of *5*.

Use of `lastIndexOf()` guarantees that we always have at least one value to snap to (in case developer passes invalid `minute-options`, for example `['not a number', null, -10]`, or `options-fn` that always returns `false`).

### DIFF
--- a/ui/src/components/time/QTime.js
+++ b/ui/src/components/time/QTime.js
@@ -282,14 +282,14 @@ export default Vue.extend({
 
     calculateSnappingGrid (inSel, count) {
       if (inSel === void 0) {
-        return void 0
+        return
       }
 
-      const snappingGrid = [...Array(count).keys()].map(inSel)
+      const snappingGrid = [ ...Array(count).keys() ].map(inSel)
 
       let consecutiveGaps = (count - 1) - snappingGrid.lastIndexOf(true)
       if (consecutiveGaps === -1) {
-        return void 0
+        return
       }
 
       for (let i = 0; i < count; i++) {
@@ -318,7 +318,8 @@ export default Vue.extend({
           }
 
           snappingGrid[i] = i
-        } else if (snappingGrid[i] === false) {
+        }
+        else if (snappingGrid[i] === false) {
           consecutiveGaps++
         }
       }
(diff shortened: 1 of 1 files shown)


rstoenescu:
Clear explanation and good implementation. A pleasure to review. Nice work and thanks for contributing!
This will go into Quasar v1.13.2.

PR #7820, opened 2020-09-25
feat(QTable): add labelformatter

martijn788:

- [ ] Bugfix
- [X] Feature
- [X] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

The omitting columns definitions is an amazing feature for quick table generation, but sometimes all the labels are okay but you need to change 1 label. Or the use case where the labels should be translated with i18n for example. Therefore it is very useful that you can insert a function for changing the labels. Another example is when you want a different capitalize-structure.

### DIFF
--- a/ui/src/components/table/table-column-selection.js
+++ b/ui/src/components/table/table-column-selection.js
@@ -2,7 +2,13 @@ import { isNumber } from '../../utils/is.js'
 
 export default {
   props: {
-    visibleColumns: Array
+    visibleColumns: Array,
+    labelFormat: {
+      type: Function,
+      default (name) {
+        return name.toUpperCase()
+      }
+    }
   },
 
   computed: {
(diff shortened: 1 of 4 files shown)


ui/src/components/table/table-column-selection.js
```diff
@@ -2,7 +2,13 @@ import { isNumber } from '../../utils/is.js'
 
 export default {
   props: {
-    visibleColumns: Array
+    visibleColumns: Array,
+    labelFormat: {
+      type: Function,
+      default (name) {
+        return name.toUpperCase()
```
hawkeye64, line 9: I think the default here would be to just `return name`

hawkeye64: I see, it's because of this was removed and was the default:
```js
label: name.toUpperCase(),
```

martijn788: Yeah, it used to be the default. So I used that again,

rstoenescu:

PR #7905, opened 2020-10-15
QTime: set minutes/seconds to 0 by default after switching views

janne-rantala:
I have had complaints from clients that the QTime is not working logically. The problem is that very often, when the client is selecting the time, it is sharp such as 13:00, 16:00 or 20:00 (or if seconds are used, 13:00:00, 15:30:00, etc). Currently, when the hour is clicked, the minute selection becomes visible, and the minute is null (not selected) by default. At this point the client is practically ready and he/she is willing to click OK button. However, the minute (0) is still required to be selected, which creates unnecessary step for the client. 

Solution: after switching from hour view to minute view, we set the minute to 0 by default (and seconds as well, when entering from minute view to seconds view). This modification is added to this PR.

I don't see any reason why the minute/second should be null after selecting hour/minute.

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/components/time/QTime.js
+++ b/ui/src/components/time/QTime.js
@@ -379,9 +379,11 @@ export default Vue.extend({
     __goToNextView () {
       if (this.view === 'Hour') {
         this.view = 'Minute'
+        this.__setMinute(0)
       }
       else if (this.withSeconds && this.view === 'Minute') {
         this.view = 'Second'
+        this.__setSecond(0)
       }
     },
 


ui/src/components/time/QTime.js
```diff
@@ -379,9 +379,11 @@ export default Vue.extend({
     __goToNextView () {
       if (this.view === 'Hour') {
         this.view = 'Minute'
+        this.__setMinute(0)
```
hawkeye64, line 382: Before setting minute or second to 0 (zero), I think you should verify they are null, in which case it would be fine to do those actions. 
What if a User set a time that had minutes or seconds to the model? Then these would be lost.

petronellius:
We are facing a similar problem. We would need something like "hour picker". Minutes and seconds are always 0, so we don't want to annoy a user by selecting minutes.
[This is the actual solution with annoying minutes selection.](https://codepen.io/petronellius/pen/LYzVyXw?editors=101)

rstoenescu:

PR #7981, opened 2020-11-01
feat(ui): Sync core animation classes with animatecss

yusufkandemir:

- Feature (_for `quasar` and `ui/dev` app_)
- Bugfix (_for `ui/dev` app_)
- Code style update (_for `ui/dev` app_)

- Yes, but it depends on the aim.

Our updating policy for animate.css is not 100% clear to me, so I need some feedback on this. They updated the default value of `animation-duration` to `1s` and removed `animated.hinge` class. Logically, if there are some users that are using these or relying on the current `.3s` duration, this will be a breaking change technically.
So, since these are small and additive(_in a positive way for UX_) changes, we can either prepare a small migrate guide and ship this in the next minor version, or we can change the default values to match with old values and ship it as is, and leave the other changes to the next major version.

- It's submitted to the `dev` branch and **not** the `master` branch

**Other information:**

From https://discordapp.com/channels/415874313728688138/596275881907978240/772097608365375509:
> `@b0otable#7831`:
> I can't seem to get some of the other options in animate.css to work
> ```html
> <transition
>   appear
>   enter-active-class="animated fadeInUpBig delay-4s"
>   leave-active-class="animated fadeOut"
>   >
>   <div class="text-caption"> (scroll down for more info!) </div>
> </transition>
> ````
> I can get the text to come in, but it flashes super quick and doesn't delay
> Am I misunderstanding the docs?

### DIFF
--- a/ui/dev/src/pages/components/transition-animatecss.vue
+++ b/ui/dev/src/pages/components/transition-animatecss.vue
@@ -6,9 +6,13 @@
       </q-card-section>
 
       <q-card-section>
-        <div class="caption">(only 4 anims showcased here)</div>
+        <div class="caption text-center text-italic">
+          (only 4 anims showcased here)
+        </div>
+
         <br />
-        <div class="row no-wrap">
+
+        <div class="row no-wrap q-col-gutter-sm">
           <q-select
             class="col"
             v-model="enter"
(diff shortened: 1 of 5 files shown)


rstoenescu:
Awesome work. Let's revert back to .3s duration for Quasar v1 (pushed a commit). We'll get back to this for Quasar v2. Thanks, Yusuf.

Issue #7991, opened 2020-11-04
QDate title/subtile "Day, Month DayNumber" change langage


oliviermorpa:
quasar version 1.14.3

QDate show the date in "Day, Month dayNumber".
when setting language to FR, we still have this "Day, month, dayNumber" but in France (and other country), we display:
"Day dayNumber Month"

see example:
Ven, Dec 4 should be Ven. 4 Dec
![qdate](https://user-images.githubusercontent.com/9378538/98078025-8cf5f880-1e71-11eb-8033-85f7e72a9a2f.JPG)

oliviermorpa:
be crefull to change the , to a point

Ven**,** Dec 4 should be Ven**.** 4 Dec

the best maybe should be to parameter the display format ?

webnoob:
We'll change this as we have done for other formats - i.e the chinese one. Will be done today / tomorrow.

webnoob:
@oliviermorpa Hey! Sorry for the delay on this.

Is this format correct? 
![image](https://user-images.githubusercontent.com/10404747/98574384-c2ef0e80-22af-11eb-9d80-0b6d82b5bdc2.png)

oliviermorpa:
yes that's the correct in France.
Thanks

rstoenescu:
Fix will be available in Quasar v1.14.4.

PR #8019, opened 2020-11-09
fix(lang): Correct date picker header date format for french #7991

webnoob:


webnoob:
Sorry for the messy history. My tree went crazy...

IlCallo:
Maybe better to redo the branch starting from up-to-date dev branch? So we avoid the commit frenzy :P

hawkeye64:
I always do a branch off the upstream branch to avoid the merge messages

webnoob:
Doesn't matter, they get squashed. I branched the way I normally do but something odd happened, not sure why. Either way, it won't matter in the main commit to Quasar as the squash will sort it.

PR #8024, opened 2020-11-10
fix(build): better handling of types that may or may not be an array

hawkeye64:


rstoenescu:
This is good, but all the warnings also need to be fixed before merging.

hawkeye64:
@rstoenescu Not being a TS guy I am not sure if I'll be able to do that. It was more to help show what still needs to be done. I wouldn't know how to type `'activeEl'`, but I can give it  atry on some of the others and let you know how far I get.

hawkeye64:
@rstoenescu I was able to add tsTYpe on all json api that had a "defintion" defined. For others, I am going to have to relent to someone like @IlCallo to resolve. Fixed a few other issues I found along the way as well.

ui/build/build.types.js
```diff
@@ -342,32 +358,38 @@ function writeIndexDTS (apis) {
...
       // If we find the symbol as part of the generated Quasar API,
-      //  we don't need to import it from custom TS API patches
+      // we don't need to import it from custom TS API patches
       if (apis.some(definition => definition.name === name)) {
         return
       }
 
-      writeLine(contents, `import { ${name} } from './api'`)
+      importName.push(name)
+      // writeLine(contents, `import { ${name} } from './api'`)
```
IlCallo, line 371: This can be removed, right?

ui/src/plugins/Dialog.json
```diff
@@ -13,7 +13,7 @@
       "params": {
         "opts": {
           "desc": "Dialog options",
-          "tsType": "QDialogOptions",
+          "tsType": "DialogOptions",
```
IlCallo, line 16: Care about this: it was the only one specified because it was the only one used.
Changing the `tsType` here you could have broken something (not sure tho, I only found one reference for this in the repo).

Does the generated `index.d.ts` file contain any errors?

ui/build/build.types.js
```diff
@@ -68,21 +67,29 @@ function convertTypeVal (type, def, required) {
   if (fallbackComplexTypeMap.has(t)) {
     if (def.definition) {
       const propDefinitions = getPropDefinitions(def.definition, required, true)
-      let lines = []
+      const lines = []
       propDefinitions.forEach(p => lines.push(...p.split('\n')))
       return propDefinitions && propDefinitions.length > 0 ? `{\n        ${lines.join('\n        ')} }${t === 'Array' ? '[]' : ''}` : fallbackComplexTypeMap.get(t)
     }
+    // uncomment to find additonal typings that probably should not be 'any[]' or `LooseDefinition`
+    // console.error(`(${++typeErrorCount}) warning: '${key}' missing 'tsType' for complex "type": "${t}"...`)
```
IlCallo, line 75: Do you remember how many of this kind are there right now?

ui/build/build.types.js
hawkeye64: The comment, yes. I left it there and asked you for review because I wanted to make sure that was proper ts importing

ui/build/build.types.js
hawkeye64: about 258

ui/src/plugins/Dialog.json
hawkeye64: I see no errors in the file (vetur, eslint, etc) unless I am doing it wrong.

hawkeye64: There is now 8494 lines in index.d.ts

hawkeye64: In regards to QDialogOptions => DialogOptions
Because QDialog and Dialog (global) are two separate entities, I felt it should not have the `Q`
The acess point is `this.$q.dialog` and the other is `<q-dialog ...`

ui/build/build.types.js
hawkeye64: Maybe we can do a type of
```
export type StringOrArrayOrObject = string | any[] | LooseDictionary
```

ui/build/build.types.js
IlCallo: If I understand correctly, the new code keeps the references in an array and then import them all with a single statement, instead of adding one import per interface as I did before. So I think it's a good change

ui/build/build.types.js
IlCallo: For style we could use a more strict type, but I prefer to address this in another PR, so we don't have to block this to wait me

IlCallo:
I made a diff between previous generated index file and the new one.
Seems like there is a problem where arrays of complex object are needed.

Example: `QBtnToggle.options` should be `QBtnToggleOptions[]`, but it's `QBtnToggleOptions` instead.

Previously it was 
```ts
opts?: {
  // .. all the complex typings
}[]
```

I think it's due to the updates on the arrays management you added.

Affected types which I found:
- `BottomSheetActions`
- `QBtnToggleOptions`
- `QTableColumnOptions`

---

I also found a problem with `DialogPromptOptions.isValid`, but in general with options which accept a function.
Previously it was `isValid? : Function`, now it's more strict `isValid (val? : string): boolean`.
It's actually better, but it loses the optionality of the property.
The correct type there would be `isValid?: (val? : string) => boolean`

Affected types which I found:
- `DialogPromptOptions.isValid`
- `NotifyOptions.onDismiss`
- `NotifyDefaultOptions.onDismiss`
- `QEditorDefinitions.handler`
- `QTableColumnOptions.sort`
- `QTableColumnOptions.format`
- `QTableRequestServerInteractionOptions.filter`

---

Another one!
`DialogOptions.ok` was `string | { [index: string]: any } | boolean`, but now it is `DialogOkOptions` which is `{ [index: string]: any }`

Affected types which I found:
- `DialogOptions.ok`
- `DialogOptions.cancel`
- `NotifyOptions` (should accept both `string` or the object)

---

There may be others problems I missed.
Try to do the diff too after you found the source of these problems: you just need to `yarn build` on your dev branch `ui` folder, save it somewhere outside the `dist` folder, then switch to this PR branch and do the same. Then select both the files, right click and "Compare selected"

yusufkandemir:
Quasar's codebase and TS support have changed a lot since this PR. I believe we no longer have any problems regarding `any` vs `any[]` or whatever was the main problem here. `tsType` works a bit differently too. Types are only generated and exported when `"autoDefineTsType": true` is set. We add manually add custom-named, advanced, hand-crafted types as needed, not for everything available. So, I am closing this PR.

PR #8046, opened 2020-11-17
feat(ClickOutside): do not close menus opened below non-seamless dialogs on pointer interaction in dialog

pdanpdan:


rstoenescu:
vue3 equivalent? :)

Issue #7933, opened 2020-10-21
Infinite Scroll in reverse mode always auto-fire at least once


kapchew:
**Describe the bug**
Infinite Scroll in reverse mode always auto-fire at least once. If it is in normal mode (without reverse), it works as expected

**Codepen/jsFiddle/Codesandbox (required)**
https://codepen.io/kapchew/pen/GRqjBre

**To Reproduce**
1. Please check the console that shows it will always trigger the onLoad event upon first load, regardless how may items initially loaded
2. If the "reverse" property is removed, it works properly

**Expected behavior**
Reverse mode should work as normal mode.

**Screenshots**

rstoenescu:
Fix will be available in Quasar v1.14.4.

PR #8051, opened 2020-11-17
fix(QInfiniteScroll): position scroll on bottom before first poll when using reverse mode #7933

pdanpdan:


rstoenescu:
ported to vue3

Issue #8041, opened 2020-11-14
QSelect dense with selected-item slot containing QAvatar wrong icon size


NicoP-S:
**Describe the bug**
If i use qSelect with selected item slot to make different qchips with different colors. The avatar icon has the wrong size in qselected dense mode. Maybe i am doing it the wrong way?

![image](https://user-images.githubusercontent.com/8988747/99142105-3e571400-2652-11eb-942b-beb6b823bfe8.png)

**Codepen/jsFiddle/Codesandbox (required)**
https://jsfiddle.net/szL0cmbp/2/

**To Reproduce**
Check the codepen. The icons seem  to be broken. I fixed it by setting
font-size on the avatar if it is in dense mode. But it seems to be the wrong solution?

rstoenescu:
Fix will be available in Quasar v1.14.4.

PR #8054, opened 2020-11-17
fix(QField): the avatar size should only apply for marginals #8041

pdanpdan:


rstoenescu:
Cherry-picked to vue3

Issue #7623, opened 2020-08-13
Wrong Q-select baseline align, if there is blank value


FragsterAt:
**Describe the bug**
Wrong Q-select baseline align, if there is blank value

**Codepen/jsFiddle/Codesandbox (required)**
https://jsfiddle.net/gdh6cr9e/30/

**To Reproduce**
1. Go to fiddle
2. See the bug %) (pay attention: blank q-input is fine)
3. Select value in "blank" q-select
4. See all text on one line
5. Clear any input
6. See the bug again

**Expected behavior**
all inputs and text on one line

**Screenshots**
https://imgur.com/1MKrDEm

**Platform (please complete the following information):**
Browsers: Firefox, chrome, edge

**Additional context**
blank q-input is fine.
In some cases workaround is class="self-start"

rstoenescu:
Fix will be available in Quasar v1.14.4.

PR #8055, opened 2020-11-17
fix(QSelect): force height of the span having the selected value to fix baseline alignment #7623

pdanpdan:


rstoenescu:
cherry-picked to vue3

Issue #8026, opened 2020-11-11
Dynamically changing a q-btn-dropdown split prop causes menu state inconsistency


dsl101:
**Describe the bug**
I have a `q-btn-dropdown` which I would like to use to offer a set of actions, and 'hold' the most recent action as the main button action using `split`. To begin with, the button is not split, as no previous action has been taken. Clicking shows the list of available actions, and choosing one calls the handler for that action as well as storing the most recent action id. At this point the button becomes 'split' but the dropdown indicator does not show the correct state of the menu (it is closed, but the arrow indicates it is open).

**Codepen/jsFiddle/Codesandbox (required)**
https://codepen.io/dsl101/pen/gOMQQKp?editors=1010

**To Reproduce**
1. At the beginning, the button is not split, and clicking shows the list of available actions
2. Choosing an action calls the handler, and sets `lastMultiAction`
3. The button is now `split`, but the dropdown arrow does not reset to the 'closed' orientation
4. Opening and closing the menu gets the arrow back in sync

**Expected behavior**
The dropdown arrow to stay in sync with whether the menu is open.

**Screenshots**

Start state: 
![image](https://user-images.githubusercontent.com/900542/98810968-f51b8000-2417-11eb-9777-9e54f65ce853.png)

Menu open:
![image](https://user-images.githubusercontent.com/900542/98810982-01074200-2418-11eb-9d1a-b375e14ab4c1.png)

Action chosen, but arrow indicates menu still open:
![image](https://user-images.githubusercontent.com/900542/98811069-209e6a80-2418-11eb-8f28-a801634ccb4a.png)

**Platform (please complete the following information):**
OS: Windows 10
Browsers: Firefox 82.0.3, Chrome (dev) 88.0.4315.5

hawkeye64:
After investigating this issue I found that the Qmenu `before-hide` doesn't get called:
```js
  'before-hide': e => {
    this.showing = false
    this.$emit('before-hide', e)
  },
```
So, `this.showing` remains true

pdanpdan:
I think the main cause should be fixed: calling before-hide and hide in befoieDestroy in QMenu if it's open

rstoenescu:
Fix will be available in Quasar v1.14.4.

PR #8057, opened 2020-11-17
fix(QBtnDropdown): close menu when split mode changes; make menu act as fit even if button is split #8026

pdanpdan:


rstoenescu:
ported to vue3

Issue #7783, opened 2020-09-17
QInput @input handler called twice when calling QDialog plugin inside.


metalsadman:
**Describe the bug**
QInput `@input` handler called twice when calling QDialog plugin inside.

**Codepen/jsFiddle/Codesandbox (required)**
https://codepen.io/metalsadman/pen/oNxPYpw
**To Reproduce**
1. Type something in the input
2. handler is called twice

**Expected behavior**
handler called once.

**Screenshots**

**Platform (please complete the following information):**
OS:
Node:
NPM:
Yarn:
Browsers:
iOS:
Android:
Electron:

**Additional context**
Workaround wrap the dialog call inside `Vue.nextTick`.

rstoenescu:
Fix will be available in Quasar v1.14.4.

PR #8059, opened 2020-11-18
fix(QInput): prevent multiple emitting of the same value when the emmit is in the same tick #7783

pdanpdan:


rstoenescu:
ported to vue3

Issue #7667, opened 2020-08-23
QTabs inside QToolbar


cheasea:
**Describe the bug**
When you reload the page, an arrow appears on the right inside the QTabs. When you click on the arrow, it disappears. QTab contain only icons. QTabs is inside QToolbar.

https://jsfiddle.net/cheasea/f0pyeLwm/18/

**To Reproduce**
1. Open jsfiddle above
2. Reload the page (exactly reload, not rerun code)

**Expected behavior**
Arrow should not appear.

**Screenshots**
![Screenshot_37](https://user-images.githubusercontent.com/37739893/90978793-18413c80-e559-11ea-836c-f2b49b065ba5.png)

**Platform (please complete the following information):**
OS: Windows 7
Browsers: Chrome

rstoenescu:
Fix will be available in Quasar v1.14.4.

PR #8060, opened 2020-11-18
fix(QTabs): calculate size of content as sum of children sizes because scroll is sometimes updated too late #7667

pdanpdan:


rstoenescu:
ported to vue3

PR #8068, opened 2020-11-19
feat(docs): Remove Content-Type header

andreiTn:

- [x] Documentation

- [ ] Yes
- [x] No

**Other information:**

`Content-Type` should not be set so the browser can generate the right header.
https://stackoverflow.com/a/39281156/8227702

This is working after a few hours of trying to figure out why my form isn't sending the right form data
```javascript
...
  async uploadFactory(files) {
      return new Promise((resolve) => {
        resolve({
          url: this.url,
          method: 'POST',
          headers: [
            { name: 'Authorization', value: `Bearer <token>` },
          ],
        });
      });
    }
...
```

### DIFF
--- a/docs/src/pages/vue-components/uploader.md
+++ b/docs/src/pages/vue-components/uploader.md
@@ -195,7 +195,6 @@ export default {
           url: 'http://localhost:4444/fileuploader/upload',
           method: 'POST',
           headers: [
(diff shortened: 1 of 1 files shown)


rstoenescu:
Thanks for contributing!

Issue #8071, opened 2020-11-19
Drag&Drop Outlook Mail (native app) into QUploader moves the mail in the trash section


Medhyena:
**Describe the bug**
When you drag&drop an outlook mail from the outlook native app to the quploader. It moves the mail to the trash section in the outlook native app.

**Codepen/jsFiddle/Codesandbox (required)**
https://quasar.dev/vue-components/uploader#QUploader-API

**To Reproduce**
1. Go to https://quasar.dev/vue-components/uploader#QUploader-API
2. Drag & drop any mail from Outlook (native app) to one QUploader
3. It appears in the QUploader
4. But the mail has "disappeared" from my Outlook inbox and has been moved to the trash section

**Expected behavior**
The mail shouldn't be moved by the Quasar component...

rstoenescu:
What is Quasar got to do with the behavior of Outlook?

Medhyena:
Well, if you drag&drop anywhere else, it works just fine, but on the QUploader IN PARTICULAR, it does move the mail while it shouldn't. So it's obviously something from Quasar side and not Outlook... Same issue for the QFile btw, I just checked.

Medhyena:
Nevermind this issue seems to appear only in the browser (I am using Chrome). I will check by my side on other browsers !
Sorry for that ^^

pdanpdan:
Ok, I think something can be done according to this:
- https://answers.microsoft.com/en-us/msoffice/forum/all/drag-and-drop-works-however-deletes-email-can-we/865f60c3-7980-4465-8a19-4d60b0ae5aba?rtAction=1575537469456&page=2
- https://support.google.com/chrome/thread/14094827?hl=en

rstoenescu:
Fix will be available in Quasar v1.14.6.

rstoenescu:
@Medhyena thanks for insisting on this

PR #8072, opened 2020-11-19
fix(QUploader/QFile): set dropEffect to copy for DnD to prevent moving emails to trash in outlook #8071

pdanpdan:


rstoenescu:
To be ported to vue3 branch

PR #8073, opened 2020-11-19
feat(core css): add more rtl ignore rules

pdanpdan:


rstoenescu:
To be ported to vue3 branch

PR #8084, opened 2020-11-20
perf/security(app): Optimize nested objects env variables

yusufkandemir:

- Refactor
- Performance
- Security

- No

**Other information:**

_Given_
```js
envDefinitions: {
  config: {
    app: {
      name: 'Foo'
      // ...
    }
  }
}
```

_When_ `process.env.config.app.name` is referenced.
_Then_ it should be replaced with `Foo`.

_But_ it was failing, the actual value was:
`{"app":{"name":"Foo"}}.app.name`

This can result in lots of unnecessary duplicate code in the final bundle. Another concern might be security, normally `process.env` itself will not be replaced by Webpack, but the current implementation that handles the environment variables might result in some secrets to slip into the final bundle silently.

### DIFF
--- a/app/lib/helpers/parse-build-env.js
+++ b/app/lib/helpers/parse-build-env.js
@@ -1,8 +1,12 @@
+const flat = require('flat')
+
 module.exports = function parseBuildEnv (envDefinitions, rootDefinitions) {
   const env = {}
 
-  for (const key in envDefinitions) {
-    env[`process.env.${key}`] = JSON.stringify(envDefinitions[key])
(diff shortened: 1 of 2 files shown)


hawkeye64:
Awesome PR @yusufkandemir !! :heart: 
I had to go look at https://www.npmjs.com/package/flat to verify what was going to happen.

hawkeye64:
I have requested @rstoenescu to review for merge

rstoenescu:
Good work, @yusufkandemir !
This is the kind of enhancements that I want to see.

PR #8090, opened 2020-11-22
fix(docs): update anchors in docs to use start/end

pdanpdan:


rstoenescu:
Docs don't use rtl, but this PR doesn't hurts.

Issue #7997, opened 2020-11-04
QCircularProgress's foreground and background colors are inverted in Safari when reverse prop is true


pawee:
**Describe the bug**
QCircularProgress's foreground and background colors are inverted in Safari when reverse prop is true

**To Reproduce**
1. Go to https://quasar.dev/vue-components/circular-progress with Safari and Chrome
2. Scroll down to reverse example
3. Compare the colors

hawkeye64:
![image](https://user-images.githubusercontent.com/10262924/100627158-e4a64900-32e3-11eb-8d58-9ec85820eae1.png)

webnoob:
@pdanpdan We had a similar issue with another component a few months back, can you recall which one it was and the fix required?

hawkeye64:
@webnoob Looks like it's because Safari is the only browser that doesn't allow negative numbers in the 3d-rotate. So, if the code is modified to always be positive (-90 === +270, for a circle) then it will work for all.
I believe I saw @pdanpdan working on a fix already.

rstoenescu:
Fix will be available in Quasar v1.14.6.

PR #8126, opened 2020-11-30
fix(QCircularProgress): safari does not support negative strokeDashOffset #7997

pdanpdan:


rstoenescu:
Pls port to vue3-work.

rstoenescu:
Ported it.

Issue #8044, opened 2020-11-17
Icons are activating/deactivating synchronously when having multiple QEditors with :toolbar props


rickxu0423:
**Describe the bug**
If we have multiple QEditors with :toolbar props (eg: :toolbar="[['bold', 'italic', 'underline']]"), the icons of those QEditors will activate/deactivate synchronously when we change only one of them (ie: Both the 'bold' icon of the first QEditor and the 'bold' icon of the second QEditor will become activated if we only bold the first QEditor).

**Codepen/jsFiddle/Codesandbox (required)**
https://codepen.io/Rickxu0423/pen/wvWOYNX?editable=true&editors=101%3Dhttps%3A%2F%2Fquasar.dev%2F

**To Reproduce**
1. Go to https://codepen.io/Rickxu0423/pen/wvWOYNX?editable=true&editors=101%3Dhttps%3A%2F%2Fquasar.dev%2F
2. Select the word 'Content1' inside the first QEditor
3. Click on the 'bold' icon in the toolbar of the first QEditor
4. See the 'bold icon' becoming activated in the second QEditor

**Expected behavior**
Icon not affecting each other when we have multiple QEditors with customized toolbar in one page.

**Screenshots**
![q-editor-issue](https://user-images.githubusercontent.com/45017976/99338093-699b5800-2838-11eb-8896-9d6b9bd7537d.jpg)

**Platform (please complete the following information):**
OS: Windows
Node: 12.14.1
NPM: 6.13.4
Browsers: Chrome

**Additional context**
Note: The bug does not appear with default toolbar.

rstoenescu:
Fix will be available in Quasar v1.14.6

rickxu0423:
@rstoenescu Thanks for the update!

pdanpdan:
This is not really fixing the broblem, just hiding it and breaking the indicators in all cases.
Needs more consideration.

rstoenescu:
A last focused approach would be good.

pdanpdan:
I thought of saving the last status, but it becomes strange when the model is changed from outside.
I think it's fair to consider the toolbar should show no status when the editor is not active.

PR #8127, opened 2020-11-30
fix(editor): Icons are activating/deactivating synchronously when having multiple QEditors with :toolbar props (#8044)

hawkeye64:


rstoenescu:
Good job!

Issue #8137, opened 2020-12-02
Make it possible to clear validation error message when the input field is cleared


DarkLite1:
**Is your feature request related to a problem? Please describe.**
On `q-input` there's the property `clearable`. I was hoping that when a field is cleared the validation error would be cleared too. I couldn't find a property for that.  

**Describe the solution you'd like**
A `Boolean` property that can clear the validation error message when the clear button/icon is clicked when the property `clearable` is used on `q-input` or when a the text of the ref is simply cleared.

**Describe alternatives you've considered**
Using the `resetVladition()` function on the `q-input` ref.

```html
<template>
          <q-input
            :label="$t('application.sapTruckRoster.label.driverId')"
            ref="driverIdInput"
            v-model="driverId"
            outlined
            hide-bottom-space
            lazy-rules
            :rules="[
              (val) =>
                (val && val > 9799999999 && val < 9999999999) ||
                $t('application.sapTruckRoster.error.driverId'),
            ]"
          >
            <template v-if="driverId" v-slot:append>
              <q-icon
                name="close"
                @click.stop="clearText()"
                class="cursor-pointer"
              />
            </template>
          </q-input>
</template>
```
```typescript
import { defineComponent, Ref, ref, watch } from '@vue/composition-api'

export default defineComponent({
  setup() {
    const driverId = ref()
    const driverIdInput = ref<null | { resetValidation: () => null }>(null)

    const clearText = () => {
      driverId.value = null
      driverIdInput.value?.resetValidation()
    }
  
    const onReset = () => {
      driverId.value = null
    }
    return { onReset, driverId, clearText, driverIdInput, }
  },
})
```

**Additional context**
Talked about this on Discord and @pdanpdan asked me to create an isue for this.

rstoenescu:
Enhancement will be available in Quasar v1.14.7.

arnonrdp:
I think we can improve this functionality even further.

In my opinion, resetValidation() should also be triggered if the user erases all characters from his input. What do you think about it, @DarkLite1 and @pdanpdan?

DarkLite1:
> I think we can improve this functionality even further.
> 
> In my opinion, resetValidation() should also be triggered if the user erases all characters from his input. What do you think about it, @DarkLite1 and @pdanpdan?

Yup, that seems logical to me

PR #8148, opened 2020-12-04
feat(QField): reset validation status for the field when clear button is pressed #8137

pdanpdan:


rstoenescu:
Please port to vue3-work.

PR #8154, opened 2020-12-05
fix(app): Added support for old behavior of nested env variables

yusufkandemir:

- Bugfix

- No

**Other information:**

Consider the following configuration for env.
```js
{
  foo: 1,
  bar: {
    baz: 2,
    qux: 3
  }
}
```

The following was working all the time.
```
process.env.foo => 1
process.env.bar.baz => 2
process.env.bar.qux => 2
```

But the following got broken after #8084:

```
process.env.bar => { baz: 2, qux: 3 }
```

This PR aims to preserve the old behavior to avoid breaking changes.

**Context:**
This issue was reported by multiple people that are using [@quasar/qenv](https://github.com/quasarframework/app-extension-qenv) and [@quasar/dotenv](https://github.com/quasarframework/app-extension-dotenv) app-extensions, to @hawkeye64, then we got in contact and I confirmed, investigated, and solved the issue.

**Implementation-wise details:**

'flat' dep got replaced with custom logic.
This custom logic preserves the original nested objects with their root keypath.

### DIFF
--- a/app/lib/helpers/parse-build-env.js
+++ b/app/lib/helpers/parse-build-env.js
@@ -1,9 +1,7 @@
-const flat = require('flat')
-
 module.exports = function parseBuildEnv (envDefinitions, rootDefinitions) {
   const env = {}
 
-  const flatEnv = flat(envDefinitions)
+  const flatEnv = flattenObject(envDefinitions)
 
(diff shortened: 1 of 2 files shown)


rstoenescu:
Good work!

Issue #8169, opened 2020-12-08
QSelect with use-input and append slot will not trigger dropdown if append slot content is clicked


DelaneDiamon:
**Describe the bug**
Using QSelect with use-input and append slot will not trigger dropdown popup if append slot is clicked, but will focus input without cursor.

**Codepen/jsFiddle/Codesandbox (required)**
https://jsfiddle.net/Diamon33/mxct8rpy/32/

**To Reproduce**
1. Click 'expand_more' icon
2. See no dropdown

**Expected behavior**
QSelect dropdown is expected to appear

**Additional context**
If use-input is **not** used - works by design

Possible workarounds:
1. this.$q.iconSet.arrow.dropdown = 'expand_more' in created hook or wherever. Not optimal, also introduces additional CSS styling and conditional class binding and non-material icons (ex. FontAwesome) are seem to be non-selectable by this.
2. append slot with @click with $refs.{refname}.showPopup - works, but you have to adjust icon element size to occupy fill input height, otherwise there will still be a space with non-dropdown click behavior around the icon.

pdanpdan:
Why don't you use `dropdownIcon` prop?
I'll check the behavior.

DelaneDiamon:
I (and our team) do use dropdown-icon, but product requirements are to change this arrow's color (and sometimes shape) whenever it has validation error or this dropdown is a required form field. 
I know it's CSS matter, but we went with "easier" path of icon color class bindings and icon names for append slot, and found out this behavior. I wanted to clarify if it's intended or not.

pdanpdan:
this is available in v1.15.0

Issue #7898, opened 2020-10-14
QSelect - Improve keyboard support (Home/End - PgUp/PgDown)


labs20:
I need to deliver navigational control over the options list using also the keys:
  **Home**           => Go to first option
  **End**              => Go to last option
  **Page Up**      => Scroll up one page* of options  
  **Page Down** => Scroll down one page of options
\* a page being the amount of options displayed on the popup viewport.

Also, the wrapping **last <=> first <=> last** option using keyboard should be optional, controlled by a property.

**Describe the solution you'd like**
Implement natively the support for mentioned keys and create a "no-wrap" property to enable/disable such behaviour.

**Describe alternatives you've considered**
I've tried to intercept key events to implement it by myself, but looks like those events are not fired from the options popup.

**Additional context**
Firing keyboard and mouse events from popup interaction would be a nice thing.

Thanks

pdanpdan:
Hello.
About disable navigation wrapping - it won't be done - it's pointless as if someone overshoots by mistake he can get back with one keypress.
About Home/End - they will only work if QSelect does not have `use-input`.
About Page Up/Down - I'll see what can be done :)

pdanpdan:
this is available in v1.15.0

PR #8176, opened 2020-12-09
feat(QSelect): improvements for QSelect #8169, #7898

pdanpdan:
- open options list on autofocus when using QMenu also (make it consistent with the case of QDialog)
- improve/fix autocomplete
- fix virtualScrollItemSize when not dense
- add PG_UP, PG_DOWN, HOME and END kbd navigation in list (HOME and END if there is no text in input)
- keep input in the base control when using QDialog to prevent visual changes
- prevent duplicate clicks due to propagation from label to input

#8169, #7898


rstoenescu:
Pls port to vue3

Issue #7929, opened 2020-10-20
QPagination should have props for page button styles


mockingjet:
QPagination should provide some button props like flat, round, outline

Maybe an option; for example, 

`
button-style-props="{ flat: true, round: true, size: '20px', padding: '10px' }" 
`

can be provided.

hawkeye64:
@mockingjet The issue here is you would need separation between props and attributes and for a lot of users this could get into confusing territory.

rstoenescu:
Enhancement will be available on Quasar v1.14.8.

mockingjet:
@rstoenescu @pdanpdan  button props lack "flat"

pdanpdan:
@mockingjet What would be the use case?

mockingjet:
@pdanpdan what do you mean by "use case"? I happen to need this implmentation 
![image](https://user-images.githubusercontent.com/32839503/111491256-ac355880-8776-11eb-92d3-1a4bb777f004.png)

mockingjet:
[line180](https://github.com/quasarframework/quasar/blob/dev/ui/src/components/pagination/QPagination.js#L180) have default flat: true
But it seems not working correctly as [this](https://quasar.dev/vue-components/button#Design)
I guess it is because props are mixed [here](https://github.com/quasarframework/quasar/blob/dev/ui/src/components/pagination/QPagination.js#L210)

mockingjet:
From [here](https://github.com/quasarframework/quasar/blob/e32a8dcfc8771806519402dec65d7f2aa29634e6/ui/src/components/pagination/QPagination.js#L380) can see only "active button" is not flat. Hope to add one more prop to determine if it needs flat style

suggestions: 
\<q-pagination flat /> where flat only works on active button

rstoenescu:
@mockingjet Will be available in Quasar v1.15.6 (and v2.0.0-beta.10):

![qpagination](https://user-images.githubusercontent.com/4665641/111618187-bbef8400-87ec-11eb-81f0-144e22bfdb5a.png)

PR #8178, opened 2020-12-09
feat(QPagination): add more button props #7929

pdanpdan:


rstoenescu:
Please port to vue3-work

PR #8185, opened 2020-12-11
perf(QSelect): re-render only the options that have changed active or focused state

pdanpdan:


rstoenescu:
Pls port to vue3-work branch.

Issue #8184, opened 2020-12-11
Tree node can not be shown correctly


yshanli:
I try to reload tree node with new tree data, but I find that some lazy node can not be built again.

Play with codepen.io:

https://codepen.io/yyl0107/pen/QWKGqYK?editors=1010

Before click root node btn:
![image](https://user-images.githubusercontent.com/41136184/101874687-545de300-3bc4-11eb-8c61-c435aedc209e.png)
After click root btn:  Node 1.1 is not a lazy node any more
![image](https://user-images.githubusercontent.com/41136184/101874776-75becf00-3bc4-11eb-8171-44ff1d74de7f.png)

pdanpdan:
Please make a codepen that shows your case and remove that code from the body of the issue.
Thank you

yshanli:
try https://codepen.io/yyl0107/pen/QWKGqYK?editors=1010

rstoenescu:
Fix will be available in Quasar v1.14.8.

PR #8189, opened 2020-12-11
fix(QTree): only consider a lazy node loaded if it has an array in the childrenKey #8184

pdanpdan:
#8184


rstoenescu:
Pls port to vue3

PR #8215, opened 2020-12-16
docs: fix parsing release notes; improve scrolling in docs

pdanpdan:


rstoenescu:
Port to vue3 pls. Note that there is no page transition effect in the new docs.

PR #8228, opened 2020-12-17
QDrawer: replace mouseout by mouseleave on examples

mdiniz:
On these QDrawer examples, where the mouse controls the "mini" state, the `mouseout` event is used
https://github.com/quasarframework/quasar/blob/dev/docs/src/examples/QDrawer/MiniToOverlay.vue#L17
https://github.com/quasarframework/quasar/blob/dev/docs/src/examples/QDrawer/MiniMouseEvents.vue#L17

This makes an event trigger whenever a person hovers different elements of the menu (that means, it gets triggered a lot). It'd be better to use the `mouseleave` event, which would trigger only once, when the person moves moves the pointer away from the entire QDrawer.

Difference between the two events from MDN (https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseleave_event):

> mouseleave and mouseout are similar but differ in that mouseleave does not bubble and mouseout does. This means that mouseleave is fired when the pointer has exited the element and all of its descendants, whereas mouseout is fired when the pointer leaves the element or leaves one of the element's descendants (even if the pointer is still within the element).

- [ ] Bugfix
- [ ] Feature
- [X] Documentation
- [ ] Code style update
- [X] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

### DIFF
--- a/docs/src/examples/QDrawer/MiniMouseEvents.vue
+++ b/docs/src/examples/QDrawer/MiniMouseEvents.vue
@@ -14,7 +14,7 @@
 
         :mini="miniState"
         @mouseover="miniState = false"
-        @mouseout="miniState = true"
+        @mouseleave="miniState = true"
 
         :width="200"
         :breakpoint="500"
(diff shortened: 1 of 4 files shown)


rstoenescu:

Issue #8188, opened 2020-12-11
"Extension server error: Object not found: <top>"


faxinwang:
**Describe the bug**
when runing the `quasar dev -m electron`,  it always reports:
```
[14564:1212/004403.341:ERROR:CONSOLE(1)] "Extension server error: Object not found: <top>", source: devtools://devtools/bundled/extensions/extensions.js (1)
[14564:1212/004404.073:ERROR:CONSOLE(1)] "Extension server error: Object not found: <top>", source: devtools://devtools/bundled/extensions/extensions.js (1)
[14564:1212/004405.072:ERROR:CONSOLE(1)] "Extension server error: Object not found: <top>", source: devtools://devtools/bundled/extensions/extensions.js (1)
[14564:1212/004406.423:ERROR:CONSOLE(1)] "Extension server error: Object not found: <top>", source: devtools://devtools/bundled/extensions/extensions.js (1)
...
```
it sames that this https://github.com/electron-userland/electron-webpack-quick-start/issues/56 can solve it.

just a line of code.

pdanpdan:
Please make a PR for this. Thank you.

hawkeye64:
@faxinwang The fix will be in the next @quasar/app release, but that won't help existing Quasar/Electron apps.

Please update your `electron-main.dev.js with the following code:
```js
/**
 * This file is used specifically and only for development. It installs
 * `electron-debug` & `vue-devtools`. There shouldn't be any need to
 *  modify this file, but it can be used to extend your development
 *  environment.
 */

import electronDebug from 'electron-debug'
import installExtension, { VUEJS_DEVTOOLS } from 'electron-devtools-installer'
import { app, BrowserWindow } from 'electron'

app.whenReady().then(() => {
  // allow for a small delay for mainWindow to be created
  setTimeout(() => {
    // Install `electron-debug` with `devtron`
    electronDebug({ showDevTools: false })

    // Install vuejs devtools
    installExtension(VUEJS_DEVTOOLS)
      .then(name => {
        console.log(`Added Extension: ${name}`)
        // get main window
        const win = BrowserWindow.getFocusedWindow()
        if (win) {
          win.webContents.on('did-frame-finish-load', () => {
            win.webContents.once('devtools-opened', () => {
              win.webContents.focus()
            })
            // open electron debug
            console.log('Opening dev tools')
            win.webContents.openDevTools()
          })
        }
      })
      .catch(err => {
        console.log('An error occurred: ', err)
      })
[...]

PR #8240, opened 2020-12-18
fix: "Extension server error: Object not found: <top>" (#8188)

hawkeye64:
I have tested this with both QCalendar and QMarkdown electron builds and it resolves the issue.


hawkeye64:
I was getting same:
image

And now I am not and things work as before.

PR #8241, opened 2020-12-19
Fix: safe check for `$q` on component instance

mesqueeb:
If I try to bundle some Quasar components in standalone UI components, so that devs can use the component regardless of their Vue environment (ideally we'd want them to just be able to import and use the component), this mixin is causing some issues because `$q` is undefined in that case.

It's fixed with a simple `?` safe check and I see no harm in adding this safety to line 12.

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/mixins/dark.js
+++ b/ui/src/mixins/dark.js
@@ -9,7 +9,7 @@ export default {
   computed: {
     isDark () {
       return this.dark === null
-        ? this.$q.dark.isActive
+        ? this.$q?.dark.isActive
         : this.dark
     }
   }


ui/src/mixins/dark.js
```diff
@@ -9,7 +9,7 @@ export default {
   computed: {
     isDark () {
       return this.dark === null
```
Milancos, line 11: Can't you just do here `return !!this.dark || !!(this.$q  && this.$q.dark.isActive)`
Not sure if `?` (optional chanining) can be used here, is it supported in quasar?

IlCallo:
AFAIK Quasar components aren't meant/optimized to be used as standalone components.
This is not the only place where a `this.$q` reference is used into a mixin, as Quasar components expect it to exist, and we probably don't have bandwidth to find an elegant and general fallback solution when `this.$q` isn't present.
For these reasons, I don't think we can accept this PR, sorry.

I'd suggest you to just add a `$q.dark` object on your Vue instance to patch the problem

mesqueeb:
Makes sense!

Issue #8260, opened 2020-12-24
QRouteTab is not active when query string is present


Demivan:
**Describe the bug**
When there is query string present q-route-tab no longer has active class.

**Codepen/jsFiddle/Codesandbox (required)**
https://codesandbox.io/s/qroutetab-issue-1147-drq84

**To Reproduce**
1. Go to codesandbox link
2. Click on tab `One` - it gains active indicator
2. Click on `Route with query` button
3. Route-tab loses active indicator

**Expected behavior**
Route tab should still be active even in url has query string.

**Screenshots**
1.14.3:
![screenshot_2020-12-24_11-22-38_465063399](https://user-images.githubusercontent.com/2339406/103078402-76029580-45da-11eb-9a46-ac21aa7c8b05.png)
1.14.4(5):
![screenshot_2020-12-24_11-22-48_966292390](https://user-images.githubusercontent.com/2339406/103078419-8155c100-45da-11eb-9c0c-68af9412d393.png)

**Platform (please complete the following information):**
OS: All
Node: 14
Yarn: berry
Browsers: All

**Additional context**
It broke in 1.14.4. Here is codesanbox link with 1.14.3 quasar version that works: https://codesandbox.io/s/qroutetab-issue-1143-nob2u
I suspect `fix(QRouteTab): fix click event handling #7686, #7143 (#8020)` broke it.

pdanpdan:
Sorry, I must have been tired.
Thank you for finding it.

rstoenescu:
Fix will be available in Quasar v1.15.0.

PR #8261, opened 2020-12-24
fix(QRouteTab): fix include path check and some cleanup #8260

pdanpdan:
The '/' added on the end is for makeing sure we are not matching partial path parts

#8260


rstoenescu:
Pls port to vue3 branch.

Issue #8288, opened 2021-01-04
Dialog plugin got undefined refs when using async component


wowissu:
**Describe the bug**
Using async component in dialog custom component
```js
this.$q.dialog({
  component: () => import('./async-comp.vue'),
  parent: this
}); 
```
Cuz QGlobalDialog did the mounted before custom component loaded so got error 
![Screen Shot 2021-01-04 at 4 45 52 PM](https://user-images.githubusercontent.com/12434914/103516953-50b03a00-4eac-11eb-82c1-aa43923f23ff.png)

**Codepen/jsFiddle/Codesandbox (required)**
(https://codesandbox.io/s/pedantic-microservice-v5tn3)

**To Reproduce**
1. Open (https://codesandbox.io/s/pedantic-microservice-v5tn3)
2. Click button
4. See error in console

**Expected behavior**
should work all kind of component

**Additional context**
Here is my solution. maybe is will help
add hook on custom component and do show() after mounted.

// global-dialog.js
![Screen Shot 2021-01-04 at 4 59 00 PM](https://user-images.githubusercontent.com/12434914/103518117-25c6e580-4eae-11eb-9f26-1d0a749c3778.png)

pianopronto:
@wowissu Maybe the dialog plugin should check for a function, but you can do it something like this:
```
  async showModal() {

    const component = (await import('./async-comp.vue')).default

    this.$q.dialog({
      component,
      parent: this,
    })
  }

```

pdanpdan:
It would make sense to use a hook on the component.

wowissu:
> @wowissu Maybe the dialog plugin should check for a function, but you can do it something like this:
> 
> ```
>   async showModal() {
> 
>     const component = (await import('./async-comp.vue')).default
> 
>     this.$q.dialog({
>       component,
>       parent: this,
>     })
>   }
> ```

This is what I am doing for now, but async component is basic support of Vue.
according to the life cycle of Vue component, access $refs in mounted is always be able to be "undefined"

rstoenescu:
Fix will be available in Quasar v1.15.1

PR #8296, opened 2021-01-06
feat(DialogPlugin): allow using async components as custom component #8288

pdanpdan:
#8288


rstoenescu:
Pls test if needed for vue3-work and port it.

PR #8298, opened 2021-01-06
feat(Notify): allow passing class and style for action buttons

pdanpdan:


rstoenescu:
No need to port to vue3-work. It's already there.

Issue #8307, opened 2021-01-08
Vertical QSeparator is misplaced when inset="true"


vandelpavel:
I'v found this old issue  #5397 and i'm having the same problem.
I'm on 1.15.0 and it's happening on chrome.

You can see the exact behaviour in the other issue.

**Additional context**
I'v also found that the problem comes when i'm using a custom height over the toolbar.
Removing the height works just fine.

Consider change height: inherit to auto.

pdanpdan:
Hello. What browser and what version, because you specify chrome and I cannot reproduce it.
Please also add a codepen.

vandelpavel:
Here is a code pen https://codepen.io/vandelpavel/pen/gOwjgJN.
Try to uncomment the height.

By the way my browser is on 87.0.4280.88.

rstoenescu:
Fix will be available in Quasar v1.15.1

PR #8308, opened 2021-01-08
fix(QSeparator): do not force width or height to avoid overflowing the parent #8307

pdanpdan:
Also exclude a few styles from RTL

#8307

### DIFF
--- a/ui/src/css/core/flex.sass
+++ b/ui/src/css/core/flex.sass
@@ -135,9 +135,9 @@
       > *
         margin-left: $size
     &-y-#{$name}
-      margin-top: -$size
+      margin-top: -$size #{"/* rtl:ignore */"}
       > *
-        margin-top: $size
+        margin-top: $size #{"/* rtl:ignore */"}
     &-#{$name}
       @extend .q-gutter-x-#{$name}
       @extend .q-gutter-y-#{$name}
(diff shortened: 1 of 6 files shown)


ui/src/css/core/flex.sass
```diff
@@ -135,9 +135,9 @@
       > *
         margin-left: $size
     &-y-#{$name}
-      margin-top: -$size
+      margin-top: -$size #{"/* rtl:ignore */"}
```
rstoenescu, line 138: any non horizontal related CSS does not gets translated for RTL purposes, so this addition should in theory not matter

pdanpdan: I know it should not have been processed, but unfortunately it is.

rstoenescu: Just checked. It doesn't process it. Those lines are not getting doubled.

pdanpdan: This is an extract from `quasar.rtl.css`:

```css
[dir] .q-gutter-y-none, [dir] .q-gutter-none {
  margin-top: 0;
}
[dir] .q-gutter-y-none > *, [dir] .q-gutter-none > * {
  margin-top: 0;
}
```

rstoenescu: The "dir" attribute is assigned to the `<html>` tag by Quasar. So it's fine.

pdanpdan: I know, but a rule with dir gets higher priority than one without. So a child of .q-gutter-y-md cannot have it's vertical padding changed by a .q-my-lg

PR #8326, opened 2021-01-12
fix #8325: Docs example axios token visibility

### LINKED ISSUE #8325: [docs] Potential security risk using an example code in practice
The [Ajax Requests](https://quasar.dev/quasar-cli/ajax-requests#Introduction) documentation includes a helpful example to automatically populate the `Authorization` header for axios, based on returned token data from a previous authentication:

```js
function setAxiosHeaders (token) {
  axios.defaults.headers.common['Authorization'] = 'Bearer ' + token
}
```

This is fine when only a single API endpoint is used, but in the general case it can lead to accidentally sending the `token` to other 3rd party services when invoking them using axios.

This example could be made safer by either calling out the hidden dangers explicitly in a comment (this is what `axios` now does in its README), or modifying the example to use a custom instance for accessing specific APIs. I think the latter would be helpful, as that's probably what the user would want in all cases.

aliclark:

- [ ] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

The [Ajax Requests](https://quasar.dev/quasar-cli/ajax-requests#Introduction) documentation includes a helpful example to automatically populate the `Authorization` header for axios, based on returned token data from a previous authentication:

```js
function setAxiosHeaders (token) {
  axios.defaults.headers.common['Authorization'] = 'Bearer ' + token
}
```

This is fine when only a single API endpoint is used, but in the general case it can lead to accidentally sending the `token` to other 3rd party services when invoking them using axios.

This example could be made safer by either calling out the hidden dangers explicitly in a comment (this is what `axios` now does in its README), or modifying the example to use a custom instance for accessing specific APIs. I think the latter would be helpful, as that's probably what the user would want in all cases.

This PR adds an "api" custom instance to the axios boot file examples, that can be used to securely set access tokens specific to that API and be confident they won't leak out to other services when using the default axios instance.

Fixes #8325


rstoenescu:
Thanks for contributing!

PR #8335, opened 2021-01-16
Update meta.md to note it also makes sense for SPA

glorat:

- [x] Documentation

- [x] No

**Other information:**
It is just a doc update - correcting a misunderstanding on meta tags that are dynamically generated for SPAs as part of SEO. Original docs say it doesn't make much sense when in fact it is a very useful SEO technique for SPAs. References available upon request (or just google the topic of googlebot and SPA)

### DIFF
--- a/docs/src/pages/quasar-plugins/meta.md
+++ b/docs/src/pages/quasar-plugins/meta.md
@@ -5,7 +5,7 @@ desc: A Quasar plugin to easily handle the meta tags of an app, helping you to a
 **Better SEO for your website!** The Meta plugin can dynamically change page title, manage `<meta>` tags, manage `<html>` and `<body>` DOM element attributes, add/remove/change `<style>` and `<script>` tags in the head of your document (useful for CDN stylesheets or for json-ld markup, for example), or manage `<noscript>` tags.
 
 ::: tip
-Take full advantage of this feature by using it with **Quasar CLI**, especially **for the SSR (Server-Side Rendering) builds**. It doesn't quite make sense to use it for SPA (Single Page Applications) too since the meta information in this case will be added at run-time and not supplied directly by the webserver (as on SSR builds).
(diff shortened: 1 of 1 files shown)


docs/src/pages/quasar-plugins/meta.md
```diff
@@ -5,7 +5,7 @@ desc: A Quasar plugin to easily handle the meta tags of an app, helping you to a
 **Better SEO for your website!** The Meta plugin can dynamically change page title, manage `<meta>` tags, manage `<html>` and `<body>` DOM element attributes, add/remove/change `<style>` and `<script>` tags in the head of your document (useful for CDN stylesheets or for json-ld markup, for example), or manage `<noscript>` tags.
 
 ::: tip
-Take full advantage of this feature by using it with **Quasar CLI**, especially **for the SSR (Server-Side Rendering) builds**. It doesn't quite make sense to use it for SPA (Single Page Applications) too since the meta information in this case will be added at run-time and not supplied directly by the webserver (as on SSR builds).
+Take full advantage of this feature by using it with **Quasar CLI**, especially **for the SSR (Server-Side Rendering) builds**. It also makes sense to use it for SPA (Single Page Applications). Although the meta information in this case will be added at run-time and not supplied directly by the webserver (as on SSR builds), modern web-crawlers like the Googlebot will render even dynamic pages and extract out the dynamically set meta information.
```
hawkeye64, line 8: Remove "even" (as in "render even dynamic") from this line as it is not needed.
I think this tip should have a reference added to it. We can use: https://developers.google.com/search/docs/guides/javascript-seo-basics

hawkeye64 (changes requested):

docs/src/pages/quasar-plugins/meta.md
glorat: Both good suggestions. I've pushed an update to address

rstoenescu:
Thanks for contributing!

PR #8348, opened 2021-01-20
feat: new private composable useFocusHelper

hawkeye64:
I was able to convert everything, except for the QTree.js:528 where it has this `ref: el => { blurTargets[ m.key ] = el }`
I tested all changes in dev testing platform.
Using useFocusHelper will keep everything unified.


rstoenescu:
There are a few things:

1. You could supply the function as param. Doesn't breaks anything.
2. It should return the h('div') directly. Why put a strain on Vue to flatten the children parameter array.
3. This is more of an util than a composable. It always returns the same result regardless of the setup() in which it is called. Not everything needs to be a composable.

Regardless, there are subtle differences between the focus helper divs/span. Let's leave them as they are.

Issue #6859, opened 2020-04-21
QSlider - max = 1 doesn't update slider length 


m0jimo:
**Describe the bug**
If you change a `:max` slider value to `1` slider length remains on its previous length.
Any other value > 0 updates the length of the slider to the max value.

**Codepen/jsFiddle/Codesandbox (required)**
[Codepan example](https://codepen.io/m0jimo/pen/XWmjvzY)

**To Reproduce**
1. Go to Codepan
2. Click on button SET MAX STEP TO = 1
3. See the slider length
You can repeat the action with following steps:
4. Click on the badge to set value = 2. Length is changed.
5. Click to the bage to set value = 10
6. Click to SET MAX STEP TO = 1 again

**Expected behavior**
If the `:max` value = 1 it should indicate value 1 but without moving to the end of slider.

**Screenshots**
Example of a slider if the max value is changed to `:max="1"` from `10`.
![Snímek obrazovky 2020-04-21 v 12 55 48](https://user-images.githubusercontent.com/37825539/79858928-7655eb80-83d0-11ea-8f61-dc5af78645bb.png)
vs
Example of s slider if the max value is change to `2`
![Snímek obrazovky 2020-04-21 v 12 55 56](https://user-images.githubusercontent.com/37825539/79859178-e7959e80-83d0-11ea-9d76-dc30e6208071.png)

**Platform (please complete the following information):**
Chrome, OS X, Codepan

pdanpdan:
Hello, can you please specify how do you expect a zero-length interval to be used as a slider?
I agree an error in the console would be nice in this case, but other than that it's only incorrect usage.

m0jimo:
I think the line can be grey without steps to the end and the only marker would be on the first position.

I agree that the slider should not be used for such an interval but if it happens it is confusing to see non existing steps.

m0jimo:
As I think more about interval, You are right. 
It looks like slider with two positions on my previous image. It can't have a grey line because it looks exactly as for two steps. It could have "white color" just to keep the size of the slider and `1` marker only on the first position.

rstoenescu:
Enhancement will be available in Quasar v1.15.2

PR #8367, opened 2021-01-23
fix(QSlider/QRange): make control not editable is min >= max and guard all divisions by 0 #6859

pdanpdan:
#6859


rstoenescu:
Pls make a minMaxDiff (= this.max - this.min) computed prop to be used everywhere.

Issue #7282, opened 2020-06-23
QSlideItem - Slide Action Active color


pietrorobbi:
**Is your feature request related to a problem? Please describe.**
I've have a UX problem with slide item: it's difficult to undersand for the user when sliding traslation fire action.

**Describe the solution you'd like**
My idea is add a "ActiveColor[left/right/...]" props. 
That color is applied on background when "this.__scale === 1"
#7288

pietrorobbi:
Hi, i'm new on Quasar community, no interest on this feature?

pdanpdan:
I would suggest another approach: a @slide event that is emitted while sliding and that allows changing colors/texts in the slots as required

rstoenescu:
Enhancement will be available in Quasar v1.15.2.

pietrorobbi:
Good suggest! Thank you

PR #8369, opened 2021-01-23
feat(QSlideItem): add @slide event to allow customizing slots while sliding #7282

pdanpdan:
#7282


rstoenescu:
Please port to vue3-work.

Issue #8446, opened 2021-02-09
Autofocus of <q-input/> is not working in v2.0.0-beta.3


xiazhichen1:
**Describe the bug**
Simply put, \<q-input autofocus /> will not automatically get focused after mounted. I also tried using template refs but still failed.

**Codepen/jsFiddle/Codesandbox (required)**
[autofocus](https://codepen.io/zxia31/pen/jOVroxe?editors=101)
[template refs](https://codepen.io/zxia31/pen/YzpWbjZ?editors=1010)

**To Reproduce**
For the `autofocus` Codepen, the input field does not get focused automatically.
For the `template refs` Codepen, the ref of the input field is retrieved but its `focus()` does not work.

**Platform (please complete the following information):**
Quasar Version: v2.0.0-beta.3

rstoenescu:
Thanks for reporting!
Fix will be available in Quasar v2.0.0-beta.5

PR #8470, opened 2021-02-13
[Q2] fix(useField): don't override public focus method if it's already defined #8446

pdanpdan:
#8446

### DIFF
--- a/ui/src/composables/private/use-field.js
+++ b/ui/src/composables/private/use-field.js
@@ -529,14 +529,17 @@ export default function (state) {
   }
 
   // expose public methods
-  Object.assign(proxy, { focus, blur })
+  Object.assign(proxy, {
+    focus: proxy.focus || focus,
+    blur
+  })
 
   onMounted(() => {
     if (isRuntimeSsrPreHydration === true && props.for === void 0) {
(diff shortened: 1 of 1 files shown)


ui/src/composables/private/use-field.js
```diff
@@ -529,14 +529,17 @@ export default function (state) {
...
+    blur
+  })
 
   onMounted(() => {
     if (isRuntimeSsrPreHydration === true && props.for === void 0) {
       state.targetUid.value = getTargetUid()
     }
 
-    props.autofocus === true && focus()
+    props.autofocus === true && proxy.focus()
```
rstoenescu, line 542: In the previous change you're looking if proxy.focus() exists, else use the local focus(). Here you assume proxy.focus() always exists. Which behavior is right?

pdanpdan: Exactly. If it didn't exist before it is the local one for sure.

The problem is that I'm not sure it's the food solution and not only a patch.
It would mean we have to add checks to all the use files for all the things added in proxy and  in state.
Maybe it would be better to move the invocation of the use files above the assign to state and proxy?

PR #8475, opened 2021-02-14
Update use-router-link.js

onurkose:
We have this nice prop named "active" in q-item component but it became ineffective with v.2
It is really helpful to handle active state manually when vue-router's exact param is just not enough in certain situations (eg. parent > child links: navigating within child links while keeping the parent active)

- [ x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/composables/private/use-router-link.js
+++ b/ui/src/composables/private/use-router-link.js
@@ -148,7 +148,7 @@ export default function () {
   const linkClass = computed(() => (
     hasLink.value === true
       ? (
-          linkIsExactActive.value === true
+          linkIsExactActive.value === true || props.active
             ? ` ${ props.exactActiveClass } ${ props.activeClass }`
             : (
                 props.exact === true


suyundukov (commented):
Hey, @onurkose!
I'm not sure if we are talking about the same thing, but `active` prop works perfectly for me (see [this](https://codepen.io/pen/?&editors=101) Codepen from docs).
Can you provide reproduction of your issue?

onurkose:
> Hey, @onurkose!
> I'm not sure if we are talking about the same thing, but `active` prop works perfectly for me (see [this](https://codepen.io/pen/?&editors=101) Codepen from docs).
> Can you provide reproduction of your issue?

The Codepen link is not working. But if you relate to [this](https://next.quasar.dev/vue-components/list-and-list-items#Example--Menu), just want you to know that that example does not use vue-router. I'll try to create an codepen example.

onurkose:
Here you go: https://codepen.io/onurkose/pen/PobWgKX?editors=1011

Please take a look at <q-item> lines especially. I made them in a way that some of them are active=true by default and with some other variations. In this example, the only way to make an item active is to click the link, all active props are ignored.

rstoenescu:
Hi,

You either let the routing system handle this or you manually handle it. But not both cause it doesn't makes any sense.

onurkose:
Hi, I've updated the [codepen](https://codepen.io/onurkose/pen/PobWgKX?editors=1010) with another example.
vue-router's handling is just not enough to handle all situations. To keep the active state as true when the user navigates deep into components like list > view > edit situations. Quasar's active prop is just there and so nice to have, and I thought it'd be a good option to override vue-router when needed, and not using it just makes vue-router handling available, nothing more. I believe using it gives more control, and totally makes sense. Because there's no option like "I don't want vue-router to handle my link states", it's a must and there's only two options for that.

onurkose:
Yes I think you are right. I've tried to improve the codepen example and discovered that it's possible to use as it is. Thanks.

rstoenescu:
Yes, if you want to handle it yourself then `@click="$router.push(...)" :active="..."`

Issue #8481, opened 2021-02-15
Form required disabled input gets triggered when sumbit


FlorinBadea:
Open https://quasar.dev/vue-components/form#Usage in Codepen and add disable attr to one of the two inputs and press "submit" btn.
Result: **disabled input** gets validated (see pic attached).

![image](https://user-images.githubusercontent.com/3803202/107954370-0a6af200-6fa5-11eb-8bcf-c7dcd44a2dc1.png)

rstoenescu:
Fix will be available in Quasar v1.15.4.

yangjiafu:
v1.15.10 disabled components can't use resetValidation void, because disabled components no class "q-validation-component"

PR #8485, opened 2021-02-16
fix(field validation): don't validate disabled components; guard calls to validate and resetValidation #8481

pdanpdan:
#8481

### DIFF
--- a/ui/src/components/form/QForm.js
+++ b/ui/src/components/form/QForm.js
@@ -113,7 +113,7 @@ export default Vue.extend({
       this.validateIndex++
 
       this.getValidationComponents().forEach(comp => {
-        comp.resetValidation()
+        typeof comp.resetValidation === 'function' && comp.resetValidation()
       })
     },
 
--- a/ui/src/mixins/validate.js
+++ b/ui/src/mixins/validate.js
@@ -34,6 +34,10 @@ export default {
       this.__validateIfNeeded()
     },
 
+    disable (val) {
+      val === true && this.resetValidation()
+    },
+
     reactiveRules: {
       handler (val) {
         if (val === true) {
--- a/ui/dev/src/pages/form/form.vue
+++ b/ui/dev/src/pages/form/form.vue
@@ -11,6 +11,7 @@
     <q-toggle v-model="greedy" label="Greedy" />
     <q-toggle v-model="loading" label="Loading" />
     <q-toggle v-model="customInput" label="Custom Input" />
+    <q-toggle v-model="titleIsDisabled" label="Disable Title QSelect" />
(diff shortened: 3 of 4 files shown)


ui/src/components/form/QForm.js
```diff
@@ -49,7 +49,7 @@ export default Vue.extend({
 
       for (let i = 0; i < components.length; i++) {
         const comp = components[i]
-        const valid = comp.validate()
+        const valid = typeof comp.validate === 'function' ? comp.validate() : true
```
rstoenescu, line 52: comp is ensured to have validate() from this.getValidationComponents()

ui/src/components/form/QForm.js
```diff
@@ -113,7 +113,7 @@ export default Vue.extend({
       this.validateIndex++
 
       this.getValidationComponents().forEach(comp => {
-        comp.resetValidation()
+        typeof comp.resetValidation === 'function' && comp.resetValidation()
```
rstoenescu, line 116: not having resetValidation() would be a user-error. This is required.

ui/src/mixins/validate.js
```diff
@@ -34,6 +34,15 @@ export default {
...
 
+    disable (val) {
+      if (val === true) {
+        this.resetValidation()
+      }
+      else {
+        this.__validateIfNeeded(true)
+      }
+    },
+
```
rstoenescu, line 45: When passing from "enable" to "disable" it makes sense to resetValidation().
But not the other way around. Let the submit trigger this, or whatever the developer has configured. So please remove the `this.__validateIfNeeded(true)`

rstoenescu (changes requested):

ui/src/components/form/QForm.js
pdanpdan: My bad, it's checked.

ui/src/components/form/QForm.js
pdanpdan: There is no documentation I can find about it being required. Even we have in dev at least one case like this.
And I can think of some reasons not to implement it :)
The problem is that if it's not present it all crashes and it's not easy to see why.

ui/src/mixins/validate.js
pdanpdan: ok

ui/src/components/form/QForm.js
pdanpdan: As this is something seldom called , I think the benefits of this being checked are worth it.

rstoenescu:
Please check vue3-work for porting the same idea. There's two methods that I've designed there (a mixin for Options API and a composable for Composition API). Both work differently than Quasar v1. @pdanpdan

PR #8496, opened 2021-02-18
fix(field validation): don't validate disabled components; guard calls to resetValidation; fix autofocus

pdanpdan:
In the new QForm validation procedure the focus can no longer go to the first not valid component in document, but goes to the first registered one


rstoenescu:
Not quite ok, but merging it and making changes.

rstoenescu:
Mini-issues were:
* No need to do a breaking change with the public useFormChild() composable
* Not ideal to require props in public composables since we can access them anyway
* Missing TS typedef change

PR #8551, opened 2021-02-28
Fixing (QUploader) uploadSizeLabel

ghost:
This small patch helps with the accumulation of previously uploaded filesizes in the size label as described [here](https://forum.quasar-framework.org/topic/7898/q-uploader-with-composition-api/14)
It would be great if uploadSize ref was in state because that would further allow the users manipulate the label but keeping it this way makes no breaking changes.

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/components/uploader/uploader-core.js
+++ b/ui/src/components/uploader/uploader-core.js
@@ -81,7 +81,7 @@ export function getRenderer (getPlugin) {
     queuedFiles: ref([]),
     uploadedFiles: ref([]),
     uploadedSize: ref(0),
-
+ 
     updateFileStatus,
     isAlive () {
       return vm.isDeactivated !== true && vm.isUnmounted !== true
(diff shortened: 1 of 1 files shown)


rstoenescu:
Hi,

Thanks for contributing. However:

That’s the whole point with the label. It shows how much you’ve uploaded. There’s also the max-files and max-total-size props. If we remove from the total size (as you proposed) then those props are essentially useless since they can be bypassed easily.

But the larger picture is that you also have access to the “reset()” method, which resets the components to its initial state. You could do that after the successful upload.

Also `uploadSize -= file.__uploaded` should have been `uploadSize.value ....` since uploadSize is a ref.

Issue #8572, opened 2021-03-04
Support Typescript for Electron, PWA, SSR (?) and quasar.conf


IlCallo:
**Is your feature request related to a problem? Please describe.**
Currently Quasar Modes only support JS files

**Describe the solution you'd like**
We should support TS in there too

**Additional context**
As a start, we need to create the TS version of all current files in these folders:
- https://github.com/quasarframework/quasar/tree/dev/app/templates/bex/js
- https://github.com/quasarframework/quasar/tree/dev/app/templates/electron/main-process
- https://github.com/quasarframework/quasar/tree/dev/app/templates/pwa
- https://github.com/quasarframework/quasar/tree/dev/app/templates/ssr
- https://github.com/quasarframework/quasar-starter-kit/blob/master/template/quasar.conf.js

This should be a pretty easy task, if anyone wants to jump in, please do!
We'll then manage the integration ourselves

Then I think a `--format=ts` option already exists for store creation, we could use that when applying the templates, or auto-detecting the presence of `tsconfig.json` file and act accordingly

faxinwang:
So we need to extend or chain the webpack configuration in the quasar.conf.js by ourself in order to supported typescript in electron, right?

IlCallo:
@faxinwang
If we are talking about Quasar v2, AFAIK you just need to change the files extension
https://next.quasar.dev/quasar-cli/developing-electron-apps/electron-with-typescript

Thanks for the contribution! I left some comments on it

yndue736:
> This is very similar issue that occurred here on the [portable table saw](https://allabouttool.com/best-portable-table-saw/) website but the suggestion given below is looking suitable for it.

yusufkandemir:
Every mode for both Quasar CLI with Vite and Quasar CLI with Webpack will now be scaffolded in TS by default if your project is a TS project 🎉

(_Except for BEX TS templates for app-webpack, it might need breaking changes. We might find another way or do it in Quasar 3, while also adding new BEX features just like been done in Vite_)

PR #8575, opened 2021-03-04
add app/templates/electron/main-process/electron-main.ts file

faxinwang:

- [x] Other, please describe:

contribute for [Support Typescript for Electron, PWA, SSR](https://github.com/quasarframework/quasar/issues/8572)
provide a typescript version of electron main entry file.

- [x] No

### DIFF
--- a/app/templates/electron/main-process/electron-main.ts
+++ b/app/templates/electron/main-process/electron-main.ts
@@ -1,0 +1,58 @@
+import { app, BrowserWindow, nativeTheme } from 'electron'
+
+try {
+  if (process.platform === 'win32' && nativeTheme.shouldUseDarkColors === true) {
+    require('fs').unlinkSync(require('path').join(app.getPath('userData'), 'DevTools Extensions'))
+  }
+} catch (_) { }
+
+/**
+ * Set `__statics` path to static files in production;
+ * The reason we are setting it here is that the path needs to be evaluated at runtime
+ */
+if (process.env.PROD) {
+  //@ts-ignore
+  global.__statics = __dirname
+}
+
+let mainWindow:BrowserWindow|null;
+
+function createWindow () {
+  /**
+   * Initial window options
+   */
+  mainWindow = new BrowserWindow({
+    width: 1000,
+    height: 600,
+    useContentSize: true,
+    webPreferences: {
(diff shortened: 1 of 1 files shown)


app/templates/electron/main-process/electron-main.ts
```diff
@@ -0,0 +1,58 @@
...
+    require('fs').unlinkSync(require('path').join(app.getPath('userData'), 'DevTools Extensions'))
+  }
+} catch (_) { }
+
+/**
+ * Set `__statics` path to static files in production;
+ * The reason we are setting it here is that the path needs to be evaluated at runtime
+ */
+if (process.env.PROD) {
+  //@ts-ignore
```
IlCallo, line 14: `@ts-ignore` isn't really an acceptable way to fix TS issues :P
Which error was it throwing previously? Maybe TS miss the context to understand this file is run into Node environment and must be made aware of

IlCallo (changes requested):

IlCallo:
Do you, by chance, have time to convert `electron-main.dev.js` file too?

faxinwang:
The `electron-main.dev.js` can be just renamed to `electron-main.dev.ts` file, because there is nothing defined in it.

app/templates/electron/main-process/electron-main.ts
faxinwang: The error is that `__statics` is not defined on `Global`. This means the environment(vscode/tsc) know that this file is to run into Node environment. I think this can be handled using .d.ts declaration file. I still new to typescript. I am learning on it.

IlCallo:
> The `electron-main.dev.js` can be just renamed to `electron-main.dev.ts` file, because there is nothing defined in it.

Yeah, can you do so? :D
Also, check arrow functions params are automatically infered by TS, it's not that automatic they are, some tweaks could be needed even if no explicit variable declaration is performed

app/templates/electron/main-process/electron-main.ts
IlCallo: No problem, contributing is a good way to learn :P
I think a .d.ts declaration which augment `Global` interface should do the trick

yusufkandemir:
This PR is still not complete but will be handled in #13758. Thanks for the contribution.

Issue #8380, opened 2021-01-26
Bootstrap-like Scrollable Dialog (the page scrolls instead of the dialog itself)


thomas-topway-it:
Hello, we are developing a quite large project heavily based on Quasar framework (for the front-end), and the only big limitation we have found with Quasar is the adopted solution for scrollable dialogs.

Because the Bootstrap's modal (see here for a demo
https://getbootstrap.com/docs/4.0/components/modal/

offers a greater accessibility on small screen devices, or a better user experience in general, we are using  (in certain cases) Bootstrap's modals within Quasar, but that's almost an hack and not the optimal solution.

You can check our solution here

https://culturaitaliana.org/books/Maria_Antonietta_Torriani/Serate_d'inverno/Prefazione

typing in the multiple choice questions whatever value (then a bootstrap modal will be shown, instead then a Quasar dialog).

We would like an option in the existing component to allow scrollable dialogs to behave as Bootstrap's modals.

Also note that we could provide a better example (in our application) which is only accessible upon logging-in, where the limitations of quasar dialogs vs bootstrap-like dialogs are more evident and where on the other side the adoption of the Bootstrap dialog interferes with standard Quasar dialogs, or creates other issues which could be solved through the proposed feature.

Thank you, and looking forward to receiving your feedback
(Thomas)

thomas-topway-it:
an update:

setting the following css

``

.q-dialog__inner {
    width: 600px;
    margin: auto;
}


.fixed-full {
 position: relative;
 display: block;
}

.fullscreen {
 position: relative;
 
}

.q-body--prevent-scroll {
    position: relative !important;

}

.q-dialog__inner--maximized > div, .q-dialog__inner--minimized > div {
 max-width: none !important;
 max-height: none !important;
 height: auto !important;
 overflow: hidden;
}

.q-body--dialog {
 overflow: scroll;
}


.q-dialog__backdrop.fixed-full {
position: fixed
}
``

and then toggling the position of q-app between relative and fixed, 
it works, however nested dialogs (dialogs in dialogs) do not seem to work,
and  transition-hide does not work as well.

Can you provide a way fix the issues above, independently by the feature request ?

(Thomas)

pdanpdan:
You can see an example here: https://codepen.io/pdanpdan/pen/gOLyOgW?editors=1010

thomas-topway-it:
@pdanpdan
[...]

PR #8628, opened 2021-03-15
docs(QDialog): add example of full page scroll dialog #8380

pdanpdan:
#8380


rstoenescu:
Example is good, but main page content bounces to the right.

Issue #8623, opened 2021-03-15
$q.loading update message


NicoP-S:
**Describe the solution you'd like**
I would like to use the loading plugin message to show some progress.
The problem is: the message prop is not reactive.
So i tried to call the show() method multiple times. This works but there is always this annoying fade effect.
https://quasar.dev/quasar-plugins/loading#Example--Show-and-Change

Would it be possible to make the prop reactive or add a updateMessage method?
https://jsfiddle.net/g1u2ezfj/24/

rstoenescu:
This will be the default behavior (not transitioning if Loading is already active) in Quasar v1.15.6 (and v2.0.0-beta.10).

PR #8630, opened 2021-03-15
feat(Loading): add option to update message without transition #8623

pdanpdan:
#8623


rstoenescu:
Make this the default behavior. Backdrop color should still animate.

PR #8647, opened 2021-03-19
Electron mode - add ts version

DarkLite1:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

**Other information:**
Related to #8575 , #8572


app/templates/electron/ts/main-process/electron-main.ts
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Set `__statics` path to static files in production;
+ * The reason we are setting it here is that the path needs to be evaluated at runtime
+ */
+
+if (process.env.PROD) {
+  global.__statics = __dirname
+}
+
+let mainWindow: null | BrowserWindow = null
```
IlCallo, line 23: Usually `null` is placed after the real type. In TS is also better to avoid `null` altogether, using `undefined` instead if you're not forced to do so.

So just adding `let mainWindow: BrowserWindow | undefined` should be fine

app/templates/electron/ts/main-process/electron-main.ts
```diff
@@ -0,0 +1,73 @@
...
+      nodeIntegrationInWorker: !!process.env.QUASAR_NODE_INTEGRATION
+
+      // More info: /quasar-cli/developing-electron-apps/electron-preload-script
+      // preload: path.resolve(__dirname, 'electron-preload.js')
+    }
+  })
+
+  if (process.env.APP_URL) {
+    try {
+      await mainWindow.loadURL(process.env.APP_URL)
```
IlCallo, line 46: I think `process.env.APP_URL` will actually be always defined in this context. You can probably replace this with a not-null assertion like `mainWindow.loadURL(process.env.APP_URL!)`.
Looking at the original implementation, you can probably also avoid the `await` (and consequentially the `async` on the main function), as it's probably not important to wait for it.

In that case, remember to disable the related linting rule

app/templates/electron/ts/main-process/electron-main.ts
```diff
@@ -0,0 +1,73 @@
...
+    } catch (error) {
+      throw new Error(`Failed loading main window: ${error as string}`)
+    }
+  }
+
+  mainWindow.on('closed', () => {
+    mainWindow = null
```
IlCallo, line 53: Following previous update, it becomes `mainWindow = undefined`

app/templates/electron/ts/main-process/electron-main.ts
```diff
@@ -0,0 +1,73 @@
...
+  })
+}
+
+app.on('ready', () => {
```
IlCallo, line 57: Can be simplified to `app.on('ready', createWindow)` again when removing the `async/await`

app/templates/electron/ts/main-process/electron-main.ts
```diff
@@ -0,0 +1,73 @@
...
+  })()
+})
+app.on('window-all-closed', () => {
+  if (process.platform !== 'darwin') {
+    app.quit()
+  }
+})
+app.on('activate', () => {
+  void (async () => {
+    if (mainWindow === null) {
```
IlCallo, line 69: Following previous update, it becomes `mainWindow === undefined`

app/templates/electron/ts/main-process/electron-main.ts
```diff
@@ -0,0 +1,73 @@
...
+    await createWindow()
+  })()
+})
+app.on('window-all-closed', () => {
+  if (process.platform !== 'darwin') {
+    app.quit()
+  }
+})
+app.on('activate', () => {
+  void (async () => {
```
IlCallo, line 68: Can be simplified again when removing the async/await

app/templates/electron/ts/typings.d.ts
```diff
@@ -0,0 +1,9 @@
+export {}
```
IlCallo, line 1: Won't this work if added directly on the top of `app/templates/electron/ts/main-process/electron-main.ts`?
I think it should

IlCallo (changes requested):

app/templates/electron/ts/main-process/electron-main.ts
DarkLite1: I used `null1 because of the code used in the end:
```ts
mainWindow === null
```

app/templates/electron/ts/main-process/electron-main.ts
DarkLite1: Using just `mainWindow.loadURL(process.env.APP_URL!)` throws:
> Forbidden non-null assertion
> Promises must be handled appropriately or explicitly marked as ignored with the `void` operator.

The second error can be resolved:
```ts
void mainWindow.loadURL(process.env.APP_URL!)
```

app/templates/electron/ts/typings.d.ts
DarkLite1: Tried adding `export {}` to the top of `app/templates/electron/ts/main-process/electron-main.ts` but that failed:
> TS2339: Property '__statics' does not exist on type 'Global'.
> Augmentations for the global scope can only be directly nested in external modules or ambient module declarations.

app/templates/electron/ts/main-process/electron-main.ts
IlCallo: Put this before the line
`// eslint-disable-next-line @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-floating-promises`

app/templates/electron/ts/typings.d.ts
IlCallo: Ok, I'll check this myself :)

app/templates/electron/ts/main-process/electron-main.ts
DarkLite1: Done :)

yusufkandemir:
Thanks for the contribution. This is no longer applicable for the `dev` branch, which is for Quasar v2. #13758 will handle it for Quasar v2, both for app-webpack and app-vite.

PR #8665, opened 2021-03-22
[extras/ui/docs/dev] Bootstrap Icons integration

hawkeye64:


rstoenescu (approved):
Looks good!
Removed the minimum Quasar version since in v2 docs it's implied it's gonna be 2.0.0+ (but will need the min version when PRing to v1).

So: release q/extras then merge this PR; then PR to v1 too.

Good work!

PR #8800, opened 2021-04-06
feat(cli): git auto initialization on creation

IlCallo:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Needs https://github.com/quasarframework/quasar-starter-kit/pull/150, https://github.com/quasarframework/quasar-starter-kit-app-extension/pull/18 and https://github.com/quasarframework/quasar-starter-kit-ui/pull/53 to work.
As discussed into https://github.com/quasarframework/quasar-starter-kit/pull/148

This will work only if `complete` function from the starter kit is a Promise to preserve backwards compatibility with custom starter kits

### DIFF
--- a/cli/bin/quasar-create
+++ b/cli/bin/quasar-create
@@ -9,9 +9,10 @@ const argv = parseArgs(process.argv.slice(2), {
     k: 'kit',
     c: 'clone',
     o: 'offline',
-    h: 'help'
+    h: 'help',
+    n: 'nogit'
   },
-  boolean: ['c', 'o', 'h'],
+  boolean: ['c', 'o', 'h', 'n'],
   string: ['k', 'b']
 })
 
@@ -21,7 +22,7 @@ if (argv.help) {
     Creates a Quasar App or App Extension project folder
 
   Usage
-    $ quasar create <project-name> [--kit <kit-name>] [--branch <version-name>]
+    $ quasar create <project-name> [--kit <kit-name>] [--branch <version-name>] [--nogit]
 
   App Examples
     $ quasar create my-project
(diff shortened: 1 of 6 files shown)


bazuka5801:
@IlCallo 
Hmm... in some scenarios we create monorepo, and often we initializing repo after creating app, then with this feature we need remember new cli arg + add this on every project that need.
In my opinion, Quasar is framework, non team-work project like github desktop or other git clients.
If we want add this, this should be optional feature with default value `false`

IlCallo:
Sorry, I cannot understand what is the problem, can you provide us an example?
You mean it's a problem for monorepos as you'll get a git initialization for every subproject you create?

Consider that Quasar CLI is pretty much the only one which still doesn't auto initialise a git repo upon project creation.
Eg. Vue CLI and Angular CLI do and let you opt out, this behaviour is the de-facto standard

bazuka5801:
@IlCallo not for every project
ex.
```
myproject
-> admin
-> app
-> backend
.....
```
I create project with quasar in app folder, then create repo in root directory (myproject) and push to github

hawkeye64:
I tend to agree that this should be a question asked to the user, just like the "use yarn or npm". At work, this would have caused issues, because the client is deep inside a mono-repo.

IlCallo:
> I create project with quasar in app folder, then create repo in root directory (myproject) and push to github

I see... But the vast majority of the community wouldn't follow your workflow, and standards should support the majority.
Again: all major CLI already do this and I'm not aware of complaints, while we had some for not being aligned with that de-facto standard.

> I tend to agree that this should be a question asked to the user, just like the "use yarn or npm". At work, this would have caused issues, because the client is deep inside a mono-repo.

Putting it on CLI has technical motivations: I initially implemented it as an option on starter kit, but since we have multiple starter kits, we would need to replicate the same code on all of them, so Razvan asked me to move it in the CLI.

Nested monorepo should surely be supported, what do you think of preventing the git init if we find a `.git` folder into the current folder ancestors?

hawkeye64:
I think it is always best to ask the User what they want. Put a question into the project setup questions as only the User will know, at that time, what they need.

PR #8879, opened 2021-04-13
feat(ui): Add min-height so badge can be used without label (indicator)

charliegilmanuk:
- [ ] Bugfix
- [x] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

This PR allows badges to be used without a label, basically just by adding a single CSS attribute (`min-height`).  Previously there was no minimum height so when the badge is used without a label it's basically a flat line.   

This allows users to for example:
- Add an indeterminate "new" state indicator to avatars, buttons and any other components that badges are usually attached to.
- Add "dots" of any color

Docs have been updated with some basic usage
image

### DIFF
--- a/docs/src/examples/QBadge/Indicators.vue
+++ b/docs/src/examples/QBadge/Indicators.vue
@@ -1,0 +1,19 @@
+<template>
+  <div class="q-pa-md q-gutter-md">
+    <q-badge rounded color="yellow" />
+    <q-badge rounded color="green" />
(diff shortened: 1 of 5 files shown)


rstoenescu:
Thanks for contributing!
This will go into Quasar v1.15.12.

PR #8908, opened 2021-04-16
Fix(QDialog): maximized dialog position if draggable

snakems:


rstoenescu:
Hi,

This will go into Quasar v1.15.11 and Quasar v2.0.0-beta.13.

pdanpdan:
@snakems hello, can you please explain why you need this change? We would need a reproduction case of your problem (in a codepen) because at a first view i cannot understand what it would fix.

snakems:
@pdanpdan   [Example bug](https://codesandbox.io/s/elated-fire-74jk4)  Drag dialog, then maximize it.

Issue #8915, opened 2021-04-16
PreFetch not working with class components


Evertvdw:
**Describe the bug**
When using class components the `preFetch` hook is not called. This is because the preFetch function is located not under `c` but under `c.__c` when using class components. See:
![image](https://user-images.githubusercontent.com/5757159/114994902-91512380-9e9d-11eb-8634-bb4d0784f402.png)

In the `client-prefetch.js` this line checks if the preFetch function exists: `.filter(m => m.c && typeof m.c.preFetch === "function")` which is of course never true when the preFetch function is not located there. 

This also needs to be adjusted for SSR as the same preFetch checking logic is applied there.

**To Reproduce**
Create a preFetch hook in the `@Options` decorator when creating a class component. For example:
```ts
import { Vue, Options } from "vue-class-component";

@Options({
	preFetch() {
		return new Promise<void>(resolve => {
			console.log("preFetch hook called!");
			resolve();
		}) as any;
	},
})
export default class Test extends Vue {}
```

**Expected behavior**
PreFetch should work for class components, so I guess the logic of checking the preFetch function should be extended to also account for class components.

Can this get some priority? Because we cannot really upgrade if this is not working properly because we do depend on the preFetch hook quite a bit.

hawkeye64:
@Evertvdw I believe preFetch can only be called if it belongs to a route.

PR #8922, opened 2021-04-16
fix(preFetch): Call preFetch with class components #8915

Evertvdw:
preFetch is now called when used inside @Options() in a class component. This fixes issue #8915.

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/app/templates/entry/client-prefetch.js
+++ b/app/templates/entry/client-prefetch.js
@@ -17,9 +17,12 @@ import { LoadingBar } from 'quasar'
 import { isRunningOnPWA } from './ssr-pwa'
 <% } %>
 
+// Class components return the component options (and the preFetch hook) inside __c property
 <% if (!ctx.mode.ssr || ctx.mode.pwa) { %>
 import App from 'app/<%= sourceFiles.rootComponent %>'
-let appPrefetch = typeof App.preFetch === 'function'
+let appPrefetch =
+  typeof App.preFetch === 'function' ||
+  (App.__c && typeof App.__c.preFetch === 'function');
 <% } %>
 
 function getMatchedComponents (to, router) {
@@ -61,6 +64,7 @@ export function addPreFetchHooks (router<%= store ? ', store' : '' %>, publicPat
         ))
       })
       .filter(m => {
+        // Class components return the component options (and the preFetch hook) inside __c property
         if (m.c && typeof m.c.preFetch === 'function') return true
         if (m.c && m.c.__c && typeof m.c.__c.preFetch === 'function') return true
         return false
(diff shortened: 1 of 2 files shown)


outofmemoryagain:
@Evertvdw appPrefetch will also need to be handled.  Please make the changes in both the client and server files.

IlCallo:
@Evertvdw thanks for the contribution, may you add a small comment on both file explaining why that's needed?

outofmemoryagain:
@Evertvdw just to clarify you can add prefetch hook to the main App component
```ts
@Options({
  preFetch: () => {
    console.log('app prefetching...');
  }
})
export default class App extends Vue {}
```

I will a comment inline

app/templates/entry/server-entry.js
```diff
@@ -137,16 +137,23 @@ export default ssrContext => {
...
+      if (typeof App.preFetch === 'function' || App.__c && typeof App.__c.preFetch === 'function') {
+        matchedComponents.unshift(App)
+      }
       // Call preFetch hooks on components matched by the route.
       // A preFetch hook dispatches a store action and returns a Promise,
       // which is resolved when the action is complete and store state has been
       // updated.
       matchedComponents
-      .filter(c => c && c.preFetch)
+      .filter(c => {
```
outofmemoryagain, line 149: Why not simplify this to be 
```ts
.filter(c => {
        // Class components return the component options (and the preFetch hook) inside __c property
        return (typeof c.preFetch === 'function') || (c.__c && typeof c.__c.preFetch === 'function')         
       #})
```

app/templates/entry/client-prefetch.js
```diff
@@ -60,18 +63,25 @@ export function addPreFetchHooks (router<%= store ? ', store' : '' %>, publicPat
           m.path.indexOf('/:') > -1 // does it has params?
         ))
       })
-      .filter(m => m.c && typeof m.c.preFetch === 'function')
-      .map(m => m.c.preFetch)
+      .filter(m => {
```
outofmemoryagain, line 66: Why not simplify this to be
```ts
      .filter(m => {
        // Class components return the component options (and the preFetch hook) inside __c property
        return (m.c && typeof m.c.preFetch === 'function') || (m.c && m.c.__c && typeof m.c.__c.preFetch === 'function')
      })
```

app/templates/entry/server-entry.js
Evertvdw: I fond my version a bit more readable, less thinking about what it does. But if you insist I can change it.

app/templates/entry/client-prefetch.js
IlCallo: Probably readability, as it's easier for the eye to understand those are branches for different use-case.
But the comment I asked to add solves the same problem, we can further simplify it as

```ts
// Class components return the component options (and the preFetch hook) inside __c property
.filter(m => (m.c && typeof m.c.preFetch === 'function') || (m.c && m.c.__c && typeof m.c.__c.preFetch === 'function'))
.map(m => m.c.__c ? m.c.__c.preFetch : m.c.preFetch)
``` 

Since if aren't required anymore

app/templates/entry/server-entry.js
IlCallo: I'd use the short version to keep the codebase style coherent, removing braces too.
Even if I agree that splitting it into multiple lines is more readable

Evertvdw: I used the short versions of the filter now.

app/templates/entry/client-prefetch.js
Evertvdw: Done!

IlCallo:
Thanks @Evertvdw!
Will be released into next beta version :)

PR #9404, opened 2021-05-24
docs(electron): Replace deprecated 'remote' module with @electron/remote (#8806)

### LINKED ISSUE #8806: @electron/remote support
"@quasar/app": "^3.0.0-beta.1"
  "electron": "^12.0.2"

image

https://github.com/electron/remote

```js
// Before
const { BrowserWindow } = require('electron').remote

// After
const { BrowserWindow } = require('@electron/remote')
```

Support new version of remote module

yusufkandemir:

- Documentation

- No

- It's submitted to the `vue3-work` branch and _not_ the `master` branch
- The related issue is referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)
- Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs).

**Other information:**
Closes #8806.

Deprecation Notice: https://github.com/electron/electron/issues/21408

### DIFF
--- a/docs/src/pages/quasar-cli/developing-electron-apps/electron-accessing-files.md
+++ b/docs/src/pages/quasar-cli/developing-electron-apps/electron-accessing-files.md
@@ -28,9 +28,9 @@ We can use the userData directory, which is reserved specifically for our applic
 // electron-main or electron-preload
 
 import path from 'path'
-import { remote } from 'electron'
+import { app } from '@electron/remote'
 
-const filePath = path.join(remote.app.getPath('userData'), '/some.file')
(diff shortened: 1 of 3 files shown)


TobyMosque:
you can access the remote property if set the `BrowserWindow > webPreferences > enableRemoteModule` to true, but that would be avoided, i think the same would be applied to a 3th module like `@electron/remote`.

IMHO, we would update that example to make use of the preload script, or even drop it.:
https://next.quasar.dev/quasar-cli/developing-electron-apps/electron-preload-script

PR #9416, opened 2021-05-25
chore: update readme files to reflect correct version of released packages

hawkeye64:
Previously, the release versions were showing Quasar v1 values...
This is for the vue3-work branch


hawkeye64:
The readme files now look like this:

image

image

image

PR #9709, opened 2021-06-18
Update export-file.js

scg32:
Proposal
In my project where DOM is complex and quasar is as additional library UMD ,function not working correctly ,above solution should fix issue with created link to download ,optional additional parameter is  'id' parent element

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/utils/export-file.js
+++ b/ui/src/utils/export-file.js
@@ -6,7 +6,7 @@ function clean (link) {
   link.remove()
 }
 
-export default function (fileName, rawData, mimeType) {
+export default function (fileName, rawData, mimeType, parentElementId) {
   const blob = new Blob([ rawData ], { type: mimeType || 'text/plain' })
 
   // IE11 has its own stuff...
@@ -21,7 +21,12 @@ export default function (fileName, rawData, mimeType) {
 
   link.classList.add('hidden')
   link.style.position = 'fixed' // avoid scrolling to bottom
-  document.body.appendChild(link)
+  if (typeof parentElementId !== 'undefined') {
+    const parentElement = document.getElementById(parentElementId)
+    parentElement.appendChild(link);
+  } else {
+    document.body.appendChild(link)
+  }
 
   try {
     link.click()


rstoenescu:
Hi,

Thanks for contributing.
Can you detail more of the problem, please? I fail to understand why this would be needed. A codepen would be wonderful.

Thanks.

scg32:
Hi,
Thank for your answer.
Prepare something on codepen will be painful:) for example in that project what I mentioned we are using html with php jquery etc... and additional we sometimes build component with quasar umd ,but always this  component is wrapped  in  parent or parents so is very nested .I am try using it like Quasar.exportFile but link which is created is not accessible from this context,I want use it to export data from q-table component ,so this is a reason my proposal
I hope so now is more clarify

rstoenescu:
I still don't understand why you need these changes.
It doesn't matter from where this method is called. When called (not a second before or after) it temporarily adds a node in document's body tag, simulates a click on it, then removes it immediately (with no delay between these actions).

scg32:
> I still don't understand why you need these changes.
> It doesn't matter from where this method is called. When called (not a second before or after) it temporarily adds a node in document's body tag, simulates a click on it, then removes it immediately (with no delay between these actions).

yes I agreed with that,the normal project will be working fine , but this is specific environment and doesn't works,so those changes, where you can send as param (tag id) will be helpful in my opinion

rstoenescu:
Well, without actually **explaining the specific environment** I cannot agree with this PR. Without the necessary explanations it's just bloatware from our perspective. Everyone enjoys a non-bloated framework, but if we keep on extending it without solid reasons then it's gonna end up as a trash product.

Please provide the logical explanation why this should be in the framework for everyone to use or feel free to write the adjusted method in your own app then use it.

PR #8852, opened 2021-04-11
feat(app v3): Add typescript support for app extensions

stefanvanherwijnen:
**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

https://github.com/quasarframework/quasar/issues/7329

PR is submitted to the vue3-work branch and it could be merged into dev, but AE TS support will require Quasar v2 (that is, the changes I made to the starter-kits are based on Quasar v2). The only internal change is that it looks up the source folder from `main` in package.json instead of assuming that it is `src`. Thus, a TS AE can have `"main": "dist/index.js"` in order for Quasar to use it correctly.

It also provides types for the API.

PR #9758, opened 2021-06-22
feat(app/ts): add types for App Extension APIs

stefanvanherwijnen:
https://github.com/quasarframework/quasar/pull/8852

This time on the 'dev' branch.

### DIFF
--- a/app/lib/app-extension/Extension.js
+++ b/app/lib/app-extension/Extension.js
@@ -125,7 +125,8 @@ module.exports = class Extension {
 
   isInstalled () {
     try {
-      require.resolve(this.packageName  + '/src/index', {
+      const pkgSrc = this.__getPkgSrc()
+      require.resolve(`${this.packageName}/${pkgSrc}/index`, {
         paths: [ appPaths.appDir ]
       })
     }
@@ -301,11 +302,18 @@ module.exports = class Extension {
     )
   }
 
+  __getPkgSrc () {
+    const pkg = require.resolve(this.packageName + `/package.json`, {
+      paths: [ appPaths.appDir ]
+    })
+    return require(pkg).main.split('/')[0]
+  }
+
   __getScript (scriptName, fatalError) {
     let script
-
+    const pkgSrc = this.__getPkgSrc()
     try {
-      script = require.resolve(this.packageName + '/src/' + scriptName, {
+      script = require.resolve(`${this.packageName}/${pkgSrc}/${scriptName}`, {
         paths: [ appPaths.appDir ]
       })
     }
--- a/app/lib/app-extension/Extension.js
+++ b/app/lib/app-extension/Extension.js
@@ -302,6 +302,10 @@ module.exports = class Extension {
     )
   }
 
+  /**
+   * Get the app extension's source folder from the package.json entrypoint "main"
+   * TODO: Should be replaced or extended with "exports" for ESM support
+   */
   __getPkgSrc () {
     const pkg = require.resolve(this.packageName + `/package.json`, {
       paths: [ appPaths.appDir ]
--- a/app/types/app-extension.d.ts
+++ b/app/types/app-extension.d.ts
@@ -1,0 +1,61 @@
+import { IResolve } from './app-paths'
+
+export interface IndexAPI {
+  ctx: Record<string, unknown>
+  extId: string
+  prompts: Record<string, unknown>
+  resolve: IResolve,
+  appDir: string,
+  // __hooks: IndexAPIHooks,
+  getPersistentConf: () => Record<string, unknown>,
+  setPersistentConf: (cfg: Record<string, unknown>) => void,
+  mergePersistentConf: (cfg: Record<string, unknown>) => void,
+  compatibleWith: (packageName: string, semverCondition?: string) => void,
+  hasPackage: (packageName: string, semverCondition?: string) => boolean,
+  hasExtension: (extId: string) => boolean,
+  getPackageVersion: (packageName: string) => string|undefined
+  extendQuasarConf: (fn: Function) => void,
+  chainWebpack: (fn: Function) => void,
+  extendWebpack: (fn: Function) => void,
+  chainWebpackMainElectronProcess: (fn: Function) => void,
+  extendWebpackMainElectronProcess: (fn: Function) => void,
+  chainWebpackWebserver: (fn: Function) => void,
+  extendWebpackWebserver: (fn: Function) => void,
+  registerCommand: (commandName: string, fn: Function) => void,
+  registerDescribeApi: (name: string, relativePath: string) => void,
+  beforeDev: (fn: Function) => void,
+  afterDev: (fn: Function) => void,
+  beforeBuild: (fn: Function) => void,
+  afterBuild: (fn: Function) => void,
+  onPublish: (fn: Function) => void,
+}
+
+export interface InstallAPI {
+  extId: string
+  prompts: Record<string, unknown>
+  resolve: IResolve,
+  appDir: string,
+  getPersistentConf: () => Record<string, unknown>,
+  setPersistentConf: (cfg: Record<string, unknown>) => void,
+  mergePersistentConf: (cfg: Record<string, unknown>) => void,
+  compatibleWith: (packageName: string, semverCondition?: string) => void,
+  hasPackage: (packageName: string, semverCondition?: string) => boolean,
+  hasExtension: (extId: string) => boolean,
+  getPackageVersion: (packageName: string) => string|undefined,
+  extendPackageJson: (extPkg: object | string) => void,
+  extendJsonFile: (file: string, newData: object) => void,
+  render: (templatePath: string, scope?: object) => void,
+  renderFile: (relativeSourcePath: string, relativeTargetPath: string, scope?: object) => void,
+  onExitLog: (msg: string) => void
+}
+
+export interface UninstallAPI {
+  extId: string
+  prompts: Record<string, unknown>
+  resolve: IResolve,
+  appDir: string,
+  getPersistentConf: () => Record<string, unknown>,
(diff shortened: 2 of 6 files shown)


smadep:
What's the status of this? Would really like to use "best-practice" http-auth with @stefanvanherwijnen 's extension

app/lib/app-extension/Extension.js
```diff
@@ -301,11 +302,18 @@ module.exports = class Extension {
     )
   }
 
+  __getPkgSrc () {
```
IlCallo, line 309: Seems like this is about having the ability to define into which folder the index file should be searched, which I guess it's related to the index.js file being build in some way and placed into a `dist` folder
Can you confirm this and possibly add a comment explaining it?

app/types/app-extension.d.ts
```diff
@@ -0,0 +1,61 @@
...
+export interface InstallAPI {
+  extId: string
+  prompts: Record<string, unknown>
+  resolve: IResolve,
+  appDir: string,
+  getPersistentConf: () => Record<string, unknown>,
+  setPersistentConf: (cfg: Record<string, unknown>) => void,
+  mergePersistentConf: (cfg: Record<string, unknown>) => void,
+  compatibleWith: (packageName: string, semverCondition?: string) => void,
+  hasPackage: (packageName: string, semverCondition?: string) => boolean,
```
IlCallo: Many of these properties are shared between the different APIs interfaces, can you group them into multiple smaller interfaces to increase maintainability?

app/types/app-extension.d.ts
```diff
@@ -0,0 +1,61 @@
+import { IResolve } from './app-paths'
+
+export interface IndexAPI {
+  ctx: Record<string, unknown>
```
IlCallo: You should be able to use `QuasarContext` here

app/types/app-extension.d.ts
```diff
@@ -0,0 +1,61 @@
...
+  appDir: string,
+  // __hooks: IndexAPIHooks,
+  getPersistentConf: () => Record<string, unknown>,
+  setPersistentConf: (cfg: Record<string, unknown>) => void,
+  mergePersistentConf: (cfg: Record<string, unknown>) => void,
+  compatibleWith: (packageName: string, semverCondition?: string) => void,
+  hasPackage: (packageName: string, semverCondition?: string) => boolean,
+  hasExtension: (extId: string) => boolean,
+  getPackageVersion: (packageName: string) => string|undefined
+  extendQuasarConf: (fn: Function) => void,
```
IlCallo: This could be better typed using `QuasarConf` type, eg `(cfg: QuasarConf, api: IndexAPI) => void`
The docs and code comment about this seems to be wrong, as it specifies the second param as the context, while it's an api proxy instead

app/types/app-extension.d.ts
```diff
@@ -0,0 +1,61 @@
...
+  chainWebpack: (fn: Function) => void,
```
IlCallo: You should be able to type these using `QuasarConf`, as these are proxies to those methods
Here's an helper you could use to extract parameters of those functions and how to use it

Eg.
```ts
// `DeepRequired` and `DeepNonNullable` are taken from `ts-essentials`, which you can add as a dev dependency to q/app

type QuasarConfProxy = DeepRequired<DeepNonNullable<QuasarConf>>;
type ExtractQuasarConfParameters<
  FirstLevelKey extends keyof QuasarConfProxy,
  SecondLevelKey extends keyof QuasarConfProxy[FirstLevelKey],
  MaybeFunction = QuasarConfProxy[FirstLevelKey][SecondLevelKey]
> = MaybeFunction extends (...args: any) => any
  ? Parameters<MaybeFunction>
  : never;

chainWebpack: (fn: (...args: [...ExtractQuasarConfParameters<"build", "chainWebpack">, IndexAPI]) => void) => void,
extendWebpack: (fn: (...args: [...ExtractQuasarConfParameters<"build", "extendWebpack">, IndexAPI]) => void) => void,
```

It seems like the second parameter is always the `isClient/isServer` object, even for webpack methods which shouldn't care about it. I guess this a technical limitation. It can be added manually or we can expose `InvokeParams` from q/app directly

app/types/app-extension.d.ts
```diff
@@ -0,0 +1,61 @@
...
+  extendWebpack: (fn: Function) => void,
+  chainWebpackMainElectronProcess: (fn: Function) => void,
+  extendWebpackMainElectronProcess: (fn: Function) => void,
+  chainWebpackWebserver: (fn: Function) => void,
+  extendWebpackWebserver: (fn: Function) => void,
+  registerCommand: (commandName: string, fn: Function) => void,
```
IlCallo: It's always better to avoid `Function`, probably a better type would be `(payload: {args: unknown[], params: Record<string, string>}): Promise<void> | void` or something similar

app/types/app-extension.d.ts
```diff
@@ -0,0 +1,61 @@
...
+  registerDescribeApi: (name: string, relativePath: string) => void,
+  beforeDev: (fn: Function) => void,
```
IlCallo: The function callback of these 4 could be more specific like `(api: IndexAPI, payload: {quasarConf: QuasarConf}): Promise<void> | void`

app/types/app-extension.d.ts
```diff
@@ -0,0 +1,61 @@
...
+  afterDev: (fn: Function) => void,
+  beforeBuild: (fn: Function) => void,
+  afterBuild: (fn: Function) => void,
+  onPublish: (fn: Function) => void,
```
IlCallo: As others, this can easy be made more specific

app/types/app-extension.d.ts
```diff
@@ -0,0 +1,61 @@
+import { IResolve } from './app-paths'
+
+export interface IndexAPI {
+  ctx: Record<string, unknown>
+  extId: string
+  prompts: Record<string, unknown>
+  resolve: IResolve,
+  appDir: string,
+  // __hooks: IndexAPIHooks,
```
IlCallo: This is an implementation detail which shouldn't be present here

IlCallo (changes requested):

IlCallo:
@stefanvanherwijnen sorry for the delay, I took a long trip to "AutomaticTestingLand" these months 😁 
Many types can be greatly improved, but it's a really nice work even with what you did just now
Let me know if you plan to improve the types with my suggestions, or I can merge the PR and try to find some time to apply them myself

stefanvanherwijnen:
The actual types are indeed not very well defined 😁. I'll try to improve them next week now that I have some more experience with typescript and the AE API.

app/lib/app-extension/Extension.js
stefanvanherwijnen: The main goal is to respect the entrypoint in `package.json`. Right now it is hardcoded to `src`. Now I think of it this solution also isn't ideal, as it won't be possible to use a nested folder for `dist`. Assuming `dist/index.js` is the compiled version of `src/index.ts`, it will work fine.

Ideally it would look up the entrypoints from the `exports` field in `package.json`, but this isn't supported in the starter kits at the moment.

stefanvanherwijnen:
I updated the types. Let me know if there is anything else I need to change.

app/lib/app-extension/Extension.js
IlCallo: I think this is good enough for now, can you leave a comment there explaining how it works now and the current limitations?

app/types/app-extension.d.ts
IlCallo: Could you check if the second expected parameter for all `chain/extendWebpack` methods is actually the `{ isClient: boolean, isServer: boolean}`?
I'm pretty sure into `QuasarConf` only some of those methods actually have the client/server object, so we may need to split the tuple type returned from `ExtractQuasarConfParameters` in some cases, manually add the `{ isClient: boolean, isServer: boolean}` param, then add other params (if there's any), otherwise people could end up trying to use `api` as second param and will get the client/server object instead

IlCallo:
Really cool, if you have any time to check the `{ isClient: boolean, isServer: boolean}` that would be really helpful :)

app/types/app-extension.d.ts
stefanvanherwijnen: According to the [docs](https://quasar.dev/app-extensions/development-guide/index-api) it is the same for all methods, however for the SSR methods the added comment is ` // isClient is always "false" and isServer is always "true"`

IlCallo: It is the same into AE Index API, but it isn't into the `QuasarConf` type, that's where things get tricky and we may need to tamper with the returned type extracted from `QuasarConf`, I don't think all have the same parameters (but I checked this nearly 3 weeks ago, I may be wrong)

stefanvanherwijnen: The extracted types are
```
type extendWebpack = (fn: (args_0: WebpackConfiguration, args_1: InvokeParams, args_2: IndexAPI) => void) => void
type chainWebpack = (fn: (args_0: Config, args_1: InvokeParams, args_2: IndexAPI) => void) => void
```
which seems correct? QuasarConf['build'] extendWebpack and chainWebpack are also statically typed: https://github.com/quasarframework/quasar/blob/87f7cd257e361e9bb79d189cc862018585361bd4/app/types/configuration/build.d.ts#L38-L46

IlCallo: You're right, I got confused by the fact that Mode specific `extend/chainWebpack` methods don't have `InvokeParams`, but since you're using the default ones for all signatures, it should be good, thanks for following up

IlCallo (approved):

stefanvanherwijnen:
Thanks :smile:

IlCallo:
Thanks to you for the contribution :)

PR #9895, opened 2021-07-01
fix(QSelect): prevent click propagation when showing as dialog #9887

pdanpdan:
#9887


rstoenescu:
Please add back the az-latn file :)

pdanpdan:
One of the .latn or .Latn should go.

Issue #9915, opened 2021-07-02
QInput textarea with trim


GitNomster:
**Describe the bug**
When i use trim in textarea, adding a space at the end of the text, then change position of cursor to any part of text, then typing - move cursor to the end. This problem is missing in text input type and textarea without trim.

**Codepen**
[https://codepen.io/gitnomster/pen/BaRNOLP](https://codepen.io/gitnomster/pen/BaRNOLP)

**To Reproduce**
Type something in input. Add space, then click on random position of text and then type any letter.

**Expected behavior**
Clear typing without jumping.

**Example**
![bug](https://user-images.githubusercontent.com/29253035/124245840-d0585200-db39-11eb-8b58-1443c1c4b617.gif)

**Platform (please complete the following information):**
Quasar Version: 1.15.21
@quasar/app Version: 2.2.10

Quasar mode:
  - [x] SPA
  - [x] PWA

Tested on:
  - [x] SPA
  - [x] PWA

OS: Win 10 x64
Browsers: Chrome 91.0.4472.124

rstoenescu:
Fix will be available in Quasar v1.15.22.

PR #9939, opened 2021-07-04
fix(QInput): restore caret position in textarea #9915

pdanpdan:
#9915


rstoenescu:
Please port to v2 too.

Issue #9980, opened 2021-07-06
QCard issue in samsung browser


ibrahim-developper:
**Describe the bug**
When open the web using my phone on google chrome the card is showing normal but when open the web on samsung chrome the card view have some issue.

**Codepen/jsFiddle/Codesandbox (required)**

`<q-card class="q-ma-lg" flat bordered>
      <q-item>
        <q-item-section avatar>
          <q-avatar>
            the image
          </q-avatar>
        </q-item-section>

        <q-item-section>
          <q-item-label>Title</q-item-label>
          <q-item-label caption> Subhead </q-item-label>
        </q-item-section>
      </q-item>

      <q-separator />
      <q-card-section horizontal>
        <q-card-section> text </q-card-section>
        <q-separator vertical />

        <q-card-section class="col-4">
          Lorem ipsum dolor sit amet, consectetur adipiscing elit.
        </q-card-section>
      </q-card-section>
    </q-card>`
this the same code from quasar.dev only i'm adding some margin 

**Screenshots**
<img src="https://user-images.githubusercontent.com/52491316/124565736-06077e80-de4b-11eb-8ce7-0148675dbadc.png" width=50% height=50%>

<img src="https://user-images.githubusercontent.com/52491316/124565914-36e7b380-de4b-11eb-8c71-53e2e2700e4d.png" width=50% height=50%>

**Platform (please complete the following information):**
Quasar Version:
@quasar/app Version:

Quasar mode:
  - [•] SPA
  - [ ] SSR
  - [ ] PWA
  - [ ] Electron
  - [ ] Cordova
  - [ ] Capacitor
  - [ ] BEX

Tested on:
  - [•] SPA
  - [ ] SSR
  - [ ] PWA
  - [ ] Electron
  - [ ] Cordova
  - [ ] Capacitor
  - [ ] BEX

OS:
Node: v14.15.4.
NPM: npm@6.14.10
Yarn:
Browsers: samsung internet Version 14.2.1.69
iOS:
Android: 9
Electron:

hawkeye64:
![1625591309234798014805874339914](https://user-images.githubusercontent.com/10262924/124640412-9beed980-de4a-11eb-87e5-0cdfef2ed2d9.jpg)
![16255913630742913573946949768053](https://user-images.githubusercontent.com/10262924/124640463-ad37e600-de4a-11eb-8382-630af5533a0f.jpg)

hawkeye64:
[...]

PR #9998, opened 2021-07-07
fix(ui): QCard issue in samsung browser (#9980)

hawkeye64:
This also applies to Opera Mobile


hawkeye64:
This is the only instance I found of `:not` using Selectors Level 4. I changed it to use Selectors Level 3.
https://www.caniuse.com/?search=%3Anot

PR #10022, opened 2021-07-09
fix(mask): don't use computed to detect if mask should be used because they are updated too late #10031

pdanpdan:
#10031


codenamezjames:
Any update on this one? I had to down grade my quasar version :(

tequila99:
> Any update on this one? I had to down grade my quasar version :(

I have the same question :(

Issue #10127, opened 2021-07-20
QSelect input now fails WCAG 2.0 A standards


wreality:
**Describe the bug**
The changes introduced by #9983 cause QSelects to actively fail WCAG 2.0 A checks in axe-core now.  Admittedly, the change is more accessible, but now fails standards checks whereas before it was less accessible but passed the check. :-)  

The specific violations:
 - Ensures ARIA attributes are allowed for an element's role: ARIA attribute is not allowed: aria-multiselectable="false"
   - Likely Remedy: `aria-multiselectable` belongs on the element with `role="listbox"`
 - aria-labelledby or aria-label attribute does not exist, references elements that do not exist or references elements that are empty
   - Likely Remedy: Should add `aria-label` similar to how QInput behaves.
 - Invalid ARIA attribute value: aria-activedescendant="f_a1004b26-2180-4fad-8186-a3b48c387fe1_-1"
   - Likely Remedy: `aria-activedescendant` must be a descendant of the element **or in the `aria-owns` comma-separated list**.  The current implementation puts the element with `role="listbox"` in the `aria-owns` attr rather than the individual elements that have `role="option"`.  Use comma-separated list of `role="option"` items in the `aria-owns` attribute. (This one is silly, the implementation here is cleaner, but [this is what the spec](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role) says)

**Codepen/jsFiddle/Codesandbox (required)**
Codepen doesn't really describe the problem.  The described violations occur when running [Axe devtools](https://www.deque.com/axe/) on [quasar.dev](https://v1.quasar.dev/vue-components/select#overview). 

**Expected behavior**
No accessibility violations.

**Platform (please complete the following information):**
Quasar Version: 1.15.23

pdanpdan:
`aria-multiselectable`: agree
`aria-label`: agree, forgot about it

`aria-activedescendant`: I can't make heads or tails of it because of these examples:
- https://www.digitala11y.com/aria-activedescendant-properties/
- https://www.digitala11y.com/aria-owns-properties/
- https://www.digitala11y.com/aria-expanded-state/
- https://www.digitala11y.com/aria-autocomplete-properties/

Also the examples do not seem to agree on the position of the `aria-expanded` attr.
But I'm not going to add a list of all option ids to `aria-owns` no matter what that checker sais :)

I'll make a test build this evening and ask you to test.

pdanpdan:
[...]

PR #10157, opened 2021-07-21
feat(QSelect): more improvements for a11y #10127

pdanpdan:
#10127


bou2you:
Looking forward to this fix!

PR #10140, opened 2021-07-20
feat(docs): new SVG Icon Explorer

hawkeye64:

PR #10173, opened 2021-07-23
Revert "feat(docs): new SVG Icon Explorer"

hawkeye64:
Reverts quasarframework/quasar#10140
- Adds too much MB to vendor file


yusufkandemir:
@hawkeye64 Instead of:
```js
importedIcons.value = Object.freeze(require('@quasar/extras/' + val.value + '/index.js'))
```
we could do
```js
importedIcons.value = await import(
  /* webpackPreload: true */
  `@quasar/extras/${val.value}/index.js`
)
```
which imports the module asynchronously thus creating different chunk(s) which means no impact on the vendor chunk, because of `import()`. It also adds a [preload](https://web.dev/preload-critical-assets/) hint because of [`webpackPreload: true`](https://webpack.js.org/guides/code-splitting/#prefetchingpreloading-modules), which means that these modules will be preloaded since they can be needed at any time when the user is on the SVG icon explorer page(_since the user can trigger a module import when interacting with the dropdown_), which will increase the performance.

Also, AFAIK we should use [`shallowRef`](https://v3.vuejs.org/api/refs-api.html#shallowref) instead of using a `ref` and dealing with `Object.freeze()`, and if I am not mistaken `shallowRef` will be even better compared to `Object.freeze()`, both in terms of code simplicity and performance.

So, if you ever have time to play around with the proposed solution and verify it works and performs better, I guess we can add this great feature back, and avoid wasting your previous efforts 🙃

hawkeye64:
@yusufkandemir Even after those changes, the vendor file is still near 12mb (11972kb). The original is 717 kb

image

hawkeye64:
Even tried this, but no chunk created.

```js
      import(
        /* webpackChunkName: "svg-explorer" */
        /* webpackMode: "lazy" */
        '@quasar/extras/' + val.value + '/index.js'
      ).then(svgFile => {
        importedIcons.value = markRaw(svgFile)
      })
```

PR #10200, opened 2021-07-26
fix(docs): 'fontawesome' should be 'fontawesome-v5'

hawkeye64:


yusufkandemir (approved):
Nice 👍 I couldn't find other old references, so that should be it.

PR #10240, opened 2021-07-29
feat(VirtualScroll,QInfiniteScroll): restore scroll position when using keep-alive

pdanpdan:


francoism90:
@pdanpdan Thanks for this commit. :)

How exactly does it work, it doesn't seem to be mentioned in the API?

Is it also useful when you use a store (Pinia/Vuex) as it does seem to restore the previous position when switching between pages/tabs?

Thanks!

pdanpdan:
It's internal, automatic. There is no need to expose it in API

Issue #10083, opened 2021-07-15
Hot reload in Quasar Dev is broken


athamour1:
**Describe the bug**
Hot reload in Quasar Dev is broken

**Codepen/jsFiddle/Codesandbox (required)**
Fork a Codepen (https://codepen.quasar.dev) or a jsFiddle (https://jsfiddle.quasar.dev) or a Codesandbox (https://codesandbox.quasar.dev) and hit save then copy-paste link here.

**To Reproduce**
just open a quasar project in dev and go in the console inside the browser

**Screenshots**
![image](https://user-images.githubusercontent.com/39333862/125748190-6e6f5124-5b36-4860-b5b9-79d4b5d18015.png)

**Platform (please complete the following information):**
Quasar Version : 3.0.2
@quasar/app Version : "^3.0.0"

OS : 5.12.16-1-MANJARO
Node : v16.4.1 
NPM : 7.18.1
Yarn : 1.22.10
Browsers : [ "Firefox Dev 90.0b12 (64-bit)", "Google Chrome 91.0.4472.114 (Official Build) (64-bit)" ] 

**Additional context**
I have heard from several friends, have the same problems in windows.

hawkeye64:
I have lot (LOTS!) of projects for Quasar v2 and non-of them have broken HMR. I would check your environment. Maybe a firewall or some other watcher of your ports.


https://user-images.githubusercontent.com/10262924/125863430-ca86f12e-6a45-4fc5-9fa0-9b135bbff485.mp4

hawkeye64:
Maybe try a different port, too

athamour1:
thats a good idea i will try to replicate the problem in a clean installation and in a new port.

oefeningen:
I have the same problem.  I've made a video to show you what happens.
I used a brand new quasar project.  (but it occured also in all my other v2 projects,  v1 projects work fine)
When i change something in the data, the changes are not visible, allthough you see the reloading in console.  When i change something in the template, you can see the changes directly, but only those in the template, not the data.
I have to manually refresh to see the changes.  

https://user-images.githubusercontent.com/29915296/126867937-f5fe7a00-cb28-4586-9a2e-12955438f321.mp4

therealcoder1337:
i have a similar error: hmr doesn't work then changing a file in linked dependencies, e. g.

- `cd my-quasar-project`
- `npm link my-local-dependency`
- change something in "my-local-dependency"
- change gets noticed, it looks like it's rebuilding
[...]

PR #10289, opened 2021-08-04
feat(app): update webpack-dev-server to v4.0.0-rc.0

hawkeye64:
The only issue I saw is handling for this which may need extra tweaking:
- host and port options can't be null or empty string
https://github.com/webpack/webpack-dev-server/releases/tag/v4.0.0-rc.0

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -110,7 +110,7 @@
     "webpack": "^5.35.0",
     "webpack-bundle-analyzer": "4.4.2",
     "webpack-chain": "6.5.1",
-    "webpack-dev-server": "4.0.0-beta.3",
+    "webpack-dev-server": "4.0.0-rc.0",
     "webpack-merge": "5.8.0",
     "webpack-node-externals": "3.0.0"
   },


LeoniePhiline:
Yes! This update is dearly needed.

Right now, neither the old nor the new option validate successfully:

Old:
```js
  devServer: {
    sockPort: 443,
  },
```
(This fails, because `sockPort` no longer validates.)

New:
```js
  devServer: {
    webSocketURL: {
      port: 443,
    },
  },
```
(This fails, because `webSocketURL` does not validate in webpack-dev-server@4-beta, but that is fixed in rc.0.)

See also: https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md

LeoniePhiline:
@hawkeye64 Why did you close the PR?

therealcoder1337:
@LeoniePhiline i guess because more work is needed, as i outlined in https://github.com/quasarframework/quasar/issues/10083#issuecomment-893016702 and https://github.com/quasarframework/quasar/issues/10083#issuecomment-893640785
simply bumping the dependency will not help, because of needed migration steps.

LeoniePhiline:
Ah I see. Thank you!

PR #8553, opened 2021-02-28
feat(QDate): support swipe gesture for switching between months

thexeos:
**What kind of change does this PR introduce?**

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

**The PR fulfills these requirements:**

- [x] It's submitted to the `dev` branch and _not_ the `master` branch
- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)
- [ ] It's been tested on a Cordova (iOS, Android) app
- [ ] It's been tested on a Electron app
- [ ] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs) (for faster update click on "Suggest an edit on GitHub" at bottom of page) or explained in the PR's description.

If adding a **new feature**, the PR's description includes:
- [x] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

When using QDate on mobile, hitting the navigation arrow to switch between months may be difficult (in fact, you can see on the screencap below that we have hidden navigation arrows for year, as they are rarely used and would sometimes result in mis-taps on mobile). It is common for native calendar controls/inputs to allow swiping to switch between months.

In addition, there already is a sliding animation/transition between months and this feature complements it nicely.

You can see it in action here:

ZMYwudRj1z

If you believe this should be behind a `swipeable` property on QDate, I can add that and the corresponding documentation to this PR. I personally believe that this should be a default behavior on mobile.

PR #10368, opened 2021-08-12
feat(QDate): support swipe gesture for switching between months

thexeos:

- [X] Feature

- [X] No

**Other information:**

This PR is same as #8553, but compatible with Quasar v2.

### DIFF
--- a/ui/src/components/date/QDate.js
+++ b/ui/src/components/date/QDate.js
@@ -1,6 +1,8 @@
-import { h, defineComponent, ref, computed, watch, Transition, nextTick, getCurrentInstance } from 'vue'
+import { h, withDirectives, defineComponent, ref, computed, watch, Transition, nextTick, getCurrentInstance } from 'vue'
 
 import QBtn from '../btn/QBtn.js'
+
+import TouchSwipe from '../../directives/TouchSwipe.js'
 
 import useDark, { useDarkProps } from '../../composables/private/use-dark.js'
 import useCache from '../../composables/private/use-cache.js'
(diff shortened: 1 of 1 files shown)


cyfung1031 (commented):
Looks Good!

hawkeye64:
@thexeos Why wouldn't you just use the swipe directive in devland?

thexeos:
> @thexeos Why wouldn't you just use the swipe directive in devland?

You could, but you would have to do some intricate math and rely on layout of QDate never changing in future releases or risk it all breaking.

Dev would have to account for these cases:
- User tapped on month and now the screen *should not be* swipeable (swiping horizontally there should have no effect)
- User tapped on year and now the screen *should be* swipeable (arrows on either side suggest this)
- User has swiped over the year field in default view (swiping horizontally there should switch the year)
- User has swiped over the month field in default view (swiping horizontally there should switch the month)
- QDate is *not* in minimal mode, and user has swiped over the "current selection" area (swiping there could, in theory, be interpreted as switching to the next/previous date)

And of course, to account for there, dev would have to either rely on Quasar-defined class names to inspect which elements swipe overlapped, or to calculate pixel offsets, thus relying on stable Quasar-defined QDate's layout.

Now, having listed these, I think it would also make sense to handle swipes over month and year in default view, and also over year selection view 🙂

hawkeye64:
> Now, having listed these, I think it would also make sense to handle swipes over month and year in default view, and also over year selection view :slightly_smiling_face:

That would make sense for a complete picture

rstoenescu:

PR #10443, opened 2021-08-20
fix(ui): Correct some wrongly exposed methods (fixes #10433)

### LINKED ISSUE #10433: QSelect is wrongly exposing some methods
**Describe the bug**
QSelect is exposing some methods wrongly, making it impossible to correctly use `getOptionLabel`, `getOptionValue`, and `isOptionDisabled` methods from outside(_e.g. through template refs_).

https://codepen.io/sborick/pen/abWemqd

**To Reproduce**
1. Open the provided Codepen link
2. See that wrong values are displayed
	- For example, in the first one, it's displaying all props of the component instance get displayed instead of `Google`)
	- See the code comments and values of others to compare.

**Expected behavior**
The expected behavior is to get the correct values, which are explained in the code comments on the repro.

**Platform**
Quasar Version: 2.0.3
@quasar/app Version: -

All modes are affected since it's a core UI bug.

**Additional context**
samborick#5569 from the Discord server noticed there is something wrong with `getOptionLabel` and `getOptionValue`:
> Hi everyone, I'm having some issues with QSelect getOptionValue and getOptionLabel.  They don't seem to be returning what I expect.  Here's a codepen example of my issue: https://codepen.io/sborick/pen/abWemqd
> am I using these methods incorrectly?

After some investigation, I found that the methods are working correctly, but they were getting exposed in the wrong way.

It's happening because arguments binding is not available in arrow functions, so it uses the arguments from the closest scope, which means the setup function, that's the reason it's returning the list of props: https://github.com/quasarframework/quasar/blob/bdcafc0486f804bc59298d79e045ddefe9e5bbae/ui/src/components/select/QSelect.js#L1429-L1431

yusufkandemir:

- Bugfix

- No (_unless you count unexpected weird behaviors as a feature 🙃_)

- The related issue is referenced in the PR's title

**Other information:**
We can also explicitly accept arguments, but that would need a little extra effort to keep the actual method and the exposed one in sync. It was previously done by passing all the parameters, so I am keeping it that way.

Fixes #10433.

### DIFF
--- a/ui/src/components/infinite-scroll/QInfiniteScroll.js
+++ b/ui/src/components/infinite-scroll/QInfiniteScroll.js
@@ -141,7 +141,7 @@ export default defineComponent({
     // expose public methods
     const vm = getCurrentInstance()
     Object.assign(vm.proxy, {
-      poll: () => poll.apply(null, arguments),
+      poll: (...args) => poll.apply(null, ...args),
       trigger, stop, reset, resume, setIndex
     })
 
--- a/ui/src/components/infinite-scroll/QInfiniteScroll.js
+++ b/ui/src/components/infinite-scroll/QInfiniteScroll.js
@@ -141,7 +141,7 @@ export default defineComponent({
     // expose public methods
     const vm = getCurrentInstance()
     Object.assign(vm.proxy, {
-      poll: (...args) => poll.apply(null, args),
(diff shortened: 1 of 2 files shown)


rstoenescu:
Nice work!

PR #10460, opened 2021-08-24
feat(QTable): Adding middle click event. Requesting feedback on implementation.

MatthewSH:
Based on the discussion #10447, this will add middle click event on QTable so that you can mimic similar behavior like opening a URL in a new tab.

Here's where I need feedback on. There is not really a "middle click" event alone that's like `onclick` or `oncontextmenu`. 
There's two ways of implementing this:
1) 
```javascript
data.onClick = evt => {
  if (evt.button === 1) {
    //
  }
}
```
2) (the way it's currently implemented)
```javascript
data.onAuxclick= evt => {
  if (evt.button === 1) {
    //
  }
}
```

I used `auxclick` mainly because of what I was reading [here](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onauxclick):
> Note: Browser vendors are implementing this property as part of a plan to improve compatibility with regards to button behaviors. Specifically, event behavior is being updated so that click only fires for primary button clicks (e.g., left mouse button), while auxclick fires for non-primary button clicks. Historically, click has generally fired for the click of any device input button, although with browser behavior being somewhat inconsistent.

Auxclick in the GlobalEventHandler is marked as experimental, even though it's supported, but the [element auxclick](https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event) should be fine to use. Would like some feedback on this implementation of "middle clicking". Specifically with auxclick and the `evt.button` vs `evt.which`, which I'm having trouble finding the difference. 

This also makes me wonder, would it be worth adding just an `auxclick` event in general that ignores if it's a middle click? This would add support for a mouse with multiple keys, not just left and right.

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/ui/src/components/table/QTable.js
+++ b/ui/src/components/table/QTable.js
@@ -100,6 +100,7 @@ export default defineComponent({
     onRowClick: Function,
     onRowDblclick: Function,
     onRowContextmenu: Function,
+    onRowMiddleclick: Function,
 
     ...useDarkProps,
     ...useFullscreenProps,
@@ -444,6 +445,15 @@ export default defineComponent({
         data.class[ 'cursor-pointer' ] = true
         data.onContextmenu = evt => {
           emit('RowContextmenu', evt, row, pageIndex)
+        }
+      }
+
+      if (props.onRowMiddleclick !== void 0) {
+        data.class[ 'cursor-pointer' ] = true;
+        data.onAuxclick = evt => {
+          if (evt.button === 1) {
+            emit('RowMiddleclick', evt, row, pageIndex)
+          }
         }
       }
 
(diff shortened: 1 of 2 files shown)


MatthewSH:
This PR has been open for 2 years, feel free to re-open if the functionality is still desired.

albertodiazdorado:
I find it a pity that this PR wasn't merged. It would be very useful:
* Middle-click on a row to open in a new tab
* Right-click on a row to select "Open in new tab"

Any chance this functionality lands in Quasar?

Issue #10373, opened 2021-08-13
virtual-scroll header sticker is not displayed


Claus1:
Code example from quasar.dev does not work.
see Sticky headers table, Playing with sticky headers sections
https://quasar.dev/vue-components/virtual-scroll#example--sticky-headers-table

Original codepen code from that sections demonstrate the bug.

![image](https://user-images.githubusercontent.com/1247062/129314316-9333a785-012f-4ce0-85c8-bae741c991da.png)

Quasar Version: 2.03
@quasar/app Version: newest

Quasar mode:
  - [ x] SPA
Tested on:
  - [x ] SPA
  
OS: Linux mint
Node: newest
NPM: newest
Yarn: newest
Browsers: Chrome

pdanpdan:
thead|tbody are wrapped in a `<template>` tag on conversion, even if they are already in a template

to make the example work you can for the moment remove the inside template tag

PR #10477, opened 2021-08-26
fix(docs): prevent wrapping thead|tbody|tfoot in template when they are already in a template #10373

pdanpdan:
#10373


rstoenescu:
Do we need the same for Qv2 docs?

pdanpdan:
Yes, it should be exactly the same changes in another file. I'll do the pr in a few hours.

PR #10594, opened 2021-09-07
feat(ts): update electron and webpack-dev-server configuration types, update ts deps

IlCallo:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Configuration types for webpack dev server and electron were a bit outdated, as TS ones
I tested this on the Quasar dev project and on 2 enterprise grade projects without issues, should be safe to merge

`devDependencies` on `@quasar/app` are used to get their typings when working with Quasar configuration object description


app/types/ts-proxies.d.ts
```diff
@@ -0,0 +1,14 @@
+// Import all optional peerDependencies types here to centralize all ts-ignore commands
+// These types will have `any` type at runtime until the appropriate package is installed
+// @ts-ignore
+import type * as ElectronBuilderProxy from "electron-builder";
+// @ts-ignore
+import type * as ElectronBuilderUtilProxy from "builder-util";
```
yusufkandemir: While the other two are included in `devDependencies`, `builder-util` is not included, but still getting imported here. If we go with the approach I explained in the other comment, we must definitely add it to `devDependencies`.

app/package.json
```diff
@@ -110,10 +108,16 @@
...
     "webpack-bundle-analyzer": "4.4.2",
     "webpack-chain": "6.5.1",
-    "webpack-dev-server": "4.0.0",
+    "webpack-dev-server": "4.1.1",
     "webpack-merge": "5.8.0",
     "webpack-node-externals": "3.0.0"
   },
+  "devDependencies": {
+    "electron-builder": "22.4.0",
+    "electron-packager": "^15.3.0",
```
yusufkandemir, line 117: Is there a specific reason why `electron-builder` is pinned to an exact version, but `electron-packager` is not?

app/types/tsconfig.json
```diff
@@ -1,7 +1,9 @@
 {
   "compilerOptions": {
     "strict": true,
-    "lib": ["es2015", "dom"]
+    "lib": ["es2015", "dom"],
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true
```
yusufkandemir:
```suggestion
```
This option is enabled but I don't see any change that utilizes it. Also, it is true by default if `esModuleInterop` is true anyways. So, I think we can get rid of this line to not pollute the config.

app/types/ts-proxies.d.ts
```diff
@@ -0,0 +1,14 @@
...
+// @ts-ignore
+import type * as ElectronPackagerProxy from "electron-packager";
+
+export {
+  ElectronBuilderProxy,
+  ElectronBuilderUtilProxy,
+  ElectronPackagerProxy,
+};
```
yusufkandemir:
```suggestion
```
I guess you previously were using `peerDependencies` when playing around, then decided to use `devDependencies` to avoid headaches. So, I guess we can get rid of this file completely, and use `import type` to import the types from the packages that are defined in `devDependencies` directly. That way we will not get errors since we will have those types at the build time, and the dependencies would not be bundled because they are `devDependencies` and we are not importing anything concrete because we are using `import type`. Please let me know if it's possible.

yusufkandemir (approved):
Nice 👏 I think we can use a similar approach on other modes that depends on an external package to be installed on the userland. For example, we import `workbox-webpack-plugin` for PWA mode, although we didn't specify it anywhere.
https://github.com/quasarframework/quasar/blob/78df1f4939936fdbf591d90addf950a8f42e665b/app/types/configuration/pwa-conf.d.ts#L85

I left some questions/comments.

app/types/ts-proxies.d.ts
IlCallo: "builder-util" is an internal dep of "electron-builder", that's why we get typings even if it's not listed in devDependencies

app/package.json
IlCallo: Apparently `electron-builder` after a given version have a dependency which requires Node 14 
I didn't want to force people contributing to quasar need to switch to Node 14, but into devland this is a problem of the end user since we install the package into dev project

app/types/ts-proxies.d.ts
IlCallo: Nope, this is intended
AFAIK this would result into TS errors in devland
If you try to import from a missing package, you'll get a TS error (hence the `@ts-ignore`s) unless you use `skipLibCheck`
If you import not-existing stuff and silence the error, you'll get an `any` for what you imported, which is what we want here

Into q/app everything works fine because `devDependencies` covers what we need, so this is a safeguard for devland which don't use these packages

app/types/tsconfig.json
IlCallo: You're right, I was using this before adding `import type * as` into proxy files and I was trying TS to stop complaining, gonna remove this change

app/package.json
```diff
@@ -110,10 +108,28 @@
...
+    "quasar": "^2.0.4",
+    "vuex": "^4.0.0"
+  },
+  "peerDependencies": {
+    "electron-builder": "^22.0.0",
+    "electron-packager": "^15.0.0"
```
yusufkandemir, line 123: I think we are definitely supporting these versions, but we need to take a look to see if we are already supporting or want to support lower versions.

IlCallo: Minimum version we support are the ones written into `app/lib/electron/bundler.js`, as we upgraded Electron deps while migrating to Qv2

PR #10671, opened 2021-09-14
Feature: Qstepper add properties to hide active and error icons (fixes #8640)

### LINKED ISSUE #8640: Ability to simply number QSteps
It seems really difficult to simply _number_ the steps in QStepper. Changing the icon to show active / inactive / error are all catered for, but it seems you can only _number_ the steps using `prefix`, and that only applies when not active.

The ability to specify something like `prefix-always`. This would always display the prefix string for every state of the QStep.

Perhaps `icon`, (and maybe `active-icon`, `error-icon`, and `done-icon`) could detect being passed a number, to avoid creating an additional prop? Then it could be possible to use:
```
:icon="1"
```
But that would not account for the case where the steps should be labelled 'A', 'B', etc.

**Additional context**
This would be really useful in Quasar v1 as well, but for this specific project I'm using v2

Evertvdw:
I was implementing a custom design for the QStepper in which I ran into a problem which I could not fix by overriding some CSS. When using a step prefix I wanted to keep that prefix visible, when editing a step but also when an error occured at that step. Currently the prefix label is removed when a step is either active, done or has an error. 

This PR introduces two properties that can be set to not show either the `active-icon` or the `error-icon`. That way the prefix remains rendered and can be styled accordingly by overriding css. Also this might be useful for people that do not want to redeclare the errorIcon/activeIcon for a step when they want this icon to be the same as the steps normal icon.

- [ ] Bugfix
- [x] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
This is my first PR doing something with UI, so if I did something incorrectly, please advise :)

Edit: for a complete picture here is the example I was trying to replicate:
image

### DIFF
--- a/ui/dev/src/pages/components/stepper.vue
+++ b/ui/dev/src/pages/components/stepper.vue
@@ -15,6 +15,8 @@
       <q-toggle label="Global Navigation" v-model="globalNav" />
       <q-toggle label="Caption" v-model="caption" />
       <q-toggle label="Prefix" v-model="prefix" />
+      <q-toggle label="Hide active icon" v-model="hideActiveIcon" />
+      <q-toggle label="Hide error icon" v-model="hideErrorIcon" />
       <q-toggle label="Use 'done' prop" v-model="useDone" />
 
       <q-toggle label="Step 4 disable" v-model="stepDisable" />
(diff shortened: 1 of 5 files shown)


Evertvdw:
@rstoenescu Can you take a look at this? It's been quite a while, looks like this one slipped past.

rstoenescu:
Hi,

Sorry, it indeed slipped.
Merged it, however will change the implementation.

PR #10735, opened 2021-09-18
QDrawer - Backdrop - Color Opacity via GPU

cyfung1031:
In the old days, "rgba" was believed as a better implementation over "opacity"
(only the background color instead of everything in the DOM)
However, background / background-color property is using CPU instead of GPU.
https://www.smashingmagazine.com/2016/12/gpu-animation-doing-it-right/

For a element without children, rgba and opacity shall have the same effect.
But in terms of changing property, opacity is definitely the better option.
It can even request a GPU thread for its changing - ["will-change"](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change)

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: GPU Acceleration on Backdrop Element for QDrawer

- [ ] Yes
- [x] No

**Other information:**
https://www.smashingmagazine.com/2016/12/gpu-animation-doing-it-right/
https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

### DIFF
--- a/ui/src/components/drawer/QDrawer.js
+++ b/ui/src/components/drawer/QDrawer.js
@@ -208,7 +208,8 @@ export default defineComponent({
     )
 
     const backdropStyle = computed(() => ({
-      backgroundColor: `rgba(0,0,0,${ flagBackdropBg.value * 0.4 })`
+      backgroundColor: "#000",
+      opacity: flagBackdropBg.value * 0.4
     }))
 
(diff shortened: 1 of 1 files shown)


cyfung1031:
Merged into #10738

PR #10738, opened 2021-09-19
GPU Accelerated QDrawer Rendering

cyfung1031:
- changed from background-color rgba to static backgroundcolor with varying opacity
- changed from transform:translateX to left/right property

Currently the css for QDrawer's panel and backdrop is not fully utilized the GPU rendering

varying the alpha channel in `backgroundColor`'s `rgba` and `transform:translateX(xxxx)` were believed faster but in fact the GPU acceleration would perform much better in `opacity` and `left/right` for modern web browser engines.

* background / background-color property is using CPU instead of GPU.
   https://www.smashingmagazine.com/2016/12/gpu-animation-doing-it-right/

* being able for using `will-change` optimization
 https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [x] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Faster Rendering with GPU Acceleration

- [x] Yes
- [ ] No

For those existing plugin modifying the QDrawer's behavior should be reviewed due to the property name changed.

**Other information:**
https://www.smashingmagazine.com/2016/12/gpu-animation-doing-it-right/
https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

in long term, `will-change` might be required immediately before the start of grabbing to let the GPU prepare the changes much effectively.

### DIFF
--- a/ui/src/components/drawer/QDrawer.js
+++ b/ui/src/components/drawer/QDrawer.js
@@ -208,7 +208,8 @@ export default defineComponent({
     )
 
     const backdropStyle = computed(() => ({
-      backgroundColor: `rgba(0,0,0,${ flagBackdropBg.value * 0.4 })`
+      backgroundColor: "#000",
+      opacity: flagBackdropBg.value * 0.4
     }))
 
     const headerSlot = computed(() => (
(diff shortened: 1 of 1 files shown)


cyfung1031:
There is new founding for the bad performance and therefore recall this pull request

Issue #10821, opened 2021-09-27
Calling `hide()` on $q.dialog with custom component fails


accron-1:
**Describe the bug**
Calling `hide()` on the built-in `$q.dialog` works as expected, but the same call on dialog that has a custom component fails with the error:

```javascript
(index):464 TypeError: dialogRef.value.hide is not a function
    at Proxy.hide (global-dialog.js?0bab:73)
    at Object.eval [as callback] (Todos.vue?c70e:630)
    at Object.cap.fromNative ((index):436)
    at <anonymous>:1:18
```

-----

Based on the following code:
https://quasar.dev/quasar-plugins/dialog#composition-api-variant
I have created the following custom component for a `$q.dialog`:

```javascript
//*** AddItemDialog.vue ***
<template>
  <q-dialog ref="dialogRef" @hide="onDialogHide">
    <q-card class="q-dialog-plugin">
      <q-card-section>
        <div class="text-h6">Add item</div>
      </q-card-section>
      <q-card-section class="q-py-none">
        <div>Type item name</div>
      </q-card-section>

      <q-card-section class="q-pt-none">
        <q-input ref="input" :color="color" dense v-model="todo" autofocus></q-input>
      </q-card-section>

      <q-card-actions align="right">
        <q-btn flat label="Cancel" :color="color" @click="onCancel" v-close-popup></q-btn>
        <q-btn flat label="Add another" :color="color" @click="onAddAnother"></q-btn>
        <q-btn flat label="Add" :color="color" @click="onAdd" v-close-popup></q-btn>
      </q-card-actions>
    </q-card>
  </q-dialog>
</template>

<script setup>
  import { useDialogPluginComponent, useQuasar } from 'quasar';
  import { ref, computed } from 'vue';

  const props = defineProps({
    callback: {
      type: Function,
      required: true
    }
  });
  const emits = defineEmits( [
    // REQUIRED; need to specify some events that your
    // component will emit through useDialogPluginComponent()
    ...useDialogPluginComponent.emits,
  ]);
  const todo = ref(''),
    input = ref(null),
    $q = useQuasar();
  const { dialogRef, onDialogHide, onDialogOK, onDialogCancel } = useDialogPluginComponent();

  function onCancel() {
    todo.value = '';
    props.callback('cancel', todo.value);
  }

  function onAddAnother() {
    input.value.focus();
    props.callback('addanother', todo.value);
    todo.value = '';
  }

  function onAdd() {
[...]

PR #10850, opened 2021-09-30
account script setup when invoking hide()

metalsadman:
https://github.com/quasarframework/quasar/issues/10821

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

**Other information:**

### DIFF
--- a/ui/src/utils/private/global-dialog.js
+++ b/ui/src/utils/private/global-dialog.js
@@ -69,8 +69,18 @@ export default function (DefaultComponent, supportsCustomComponent, parentApp) {
           return API
         },
         hide () {
-          if (dialogRef.value !== null) {
-            dialogRef.value.hide()
+          if (dialogRef.value !== null && dialogRef.value.hide !== void 0) {
+            dialogRef.value.hide();
+          } else if (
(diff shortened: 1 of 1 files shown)


yusufkandemir (approved):
This solution is hacky, the real problem behind is that components using `script setup` are [closed by default](https://v3.vuejs.org/api/sfc-script-setup.html#defineexpose), and this doesn't work for this case:
https://github.com/quasarframework/quasar/blob/f138b43de8a74bbd514bd2fecbc28ec89eb5f2c2/ui/src/composables/use-dialog-plugin-component.js#L26-L27
But we can't also simply expect users to call `defineExpose`. Firstly, we don't return `show` and `hide` from the composable. Secondly, even if we return them it makes the usage more complicated than it should be, IMO. I am playing around with the Expose API these days, and will probably have a PR after this gets merged.

In short, since this is basically the same solution that is applied to the `show` method, which is working for a long time, it's fine, we can directly merge it.

PR #10890, opened 2021-10-04
fix(build): "version" issue when directly connecting to 'src' folder

hawkeye64:
When `yarn build` is run, a `version.js` file is created/updated with current version number. This is ideally a better solution than using the `@rollup/plugin-replace`.

Why is this better?

There is always a `version.js` file available always in the `src` folder.  For developers doing cutting edge development and connecting directly to the `src` folder they won't run into the `__QUASAR_VERSION__ undefined` issue.

Secondly, it's not a JSON file so won't pollute the dist stack.

From `quasar.umd.js`:
```js
  const version = '2.1.0';

  const quasarKey = '_q_';
  const timelineKey = '_q_t_';
  const stepperKey = '_q_s_';
  const layoutKey = '_q_l_';
// ...
```

The common and esm distributables are also similar.

I've submitted this PR so you have a choice of keeping the way it is or picking this PR as an alternative solution (and frankly, I think it's better because a dev can now import `vue-plugin` if needed).

If this PR is accepted, I can make similar changes to the UI kit.

### DIFF
--- a/ui/build/script.build.js
+++ b/ui/build/script.build.js
@@ -21,6 +21,8 @@ console.log(` 📦 Building Quasar ${ green('v' + require('../package.json').ver
 
 createFolder('dist')
 
+require('./script.version.js')
+
 if (!type || type === 'js') {
   createFolder('dist/vetur')
   createFolder('dist/api')
--- a/ui/build/script.version.js
+++ b/ui/build/script.version.js
@@ -1,0 +1,14 @@
+const fs = require('fs')
+const path = require('path')
+
+//  get version
+const version = require('../package.json').version
+
+// read in the template as text
+let template = fs.readFileSync(path.resolve(__dirname, './version/version.template'), 'utf-8')
+
+// do the replacement
+template = template.replace('__QUASAR_VERSION__', version)
+
+// write the file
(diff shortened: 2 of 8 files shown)


hawkeye64:
Reference: https://github.com/quasarframework/quasar-ui-qcalendar/issues/345

rstoenescu:
Hi,

The __QUASAR_VERSION__ is not the only one used in UI src. So this does not actually make it work fully.
But the usages of Quasar UI are carefully taken into account and handled properly (Quasar CLI does the auto-replacement, UMD is used through the already compiled dist, Vue CLI also does the auto-replacement). If Quasar UI is being used as indicated through Quasar CLI then this is not needed.

Furthermore, one disadvantage of this approach is that it will keep on modifying the version.js file for each new version (so adds the necessity of additional GIT commits that have no benefit to the actual code), and it requires more thought when releasing a new version to NPM. It requires you to commit the version.js file before actually building the UI and doing the release on NPM. Bottom line, why add one more possible point of failure to an already complex release system.

hawkeye64:
Thanks for the comment. I especially appreciate the thoroughness of the explanation. :heart:

PR #11026, opened 2021-10-18
fix(Meta): add a space before concatening html tag attributes to ssr context

freddy38510:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

`ctx.htmlAttrs` already contains attributes from Quasar lang plugin.

Without this PR when adding html tag attributes with Meta Plugin, the html output is: 
```javascript
<html lang=en-US dir=ltrxmlns:cc=http://creativecommons.org/ns#>
```

With this PR the output is:
```javascript
<html lang=en-US dir=ltr xmlns:cc=http://creativecommons.org/ns#>
```

### DIFF
--- a/ui/src/plugins/Meta.js
+++ b/ui/src/plugins/Meta.js
@@ -190,7 +190,11 @@ function injectServerMeta (ssrContext) {
     : ''
 
   const ctx = ssrContext._meta
-
+  
+  if(Object.keys(data.htmlAttr).length > 0) {
+    ctx.htmlAttrs += ' '
+  }
+  
   ctx.htmlAttrs += Object.keys(data.htmlAttr)
     .filter(htmlFilter)
     .map(getAttr(data.htmlAttr))
--- a/ui/src/plugins/Meta.js
+++ b/ui/src/plugins/Meta.js
@@ -194,21 +194,22 @@ function injectServerMeta (ssrContext) {
   const htmlAttr = Object.keys(data.htmlAttr).filter(htmlFilter)
 
   if (htmlAttr.length > 0) {
-    if(ctx.htmlAttrs.length > 0) {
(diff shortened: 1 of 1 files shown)


pdanpdan:
Hello. Please move `Object.keys(data.htmlAttr).filter(htmlFilter)` into a variable, and then check its length to add the space

freddy38510:
Hello. Good catch @pdanpdan, i writed this PR too fast.

Does this sound good to you now?

pdanpdan:
Now move the adding of attrs also in the if, because else it is a boop.
And I would add the space only if ctx.htmlAttr has something inside already (length > 0)

freddy38510:
Done what you suggested. I guess this PR requires the same logic applied for others tag attributes to be consistent.

pdanpdan:
If it is so and it is required for all attributes then it will be better to make a function that takes the attribute and the thing to append and make the concatenation. But I think it would be better to consult with Razvan on this.

freddy38510:
I just take a look and it missing a space in body tag attributes too when using SSR mode. The body tag already contains the attribute `data-server-rendered`.
```javascript
<body class="desktop no-touch body--light" data-server-renderedaction-scope=xyz>
```

As far as I can tell, only the `<html>` and `<body>` tags are affected by this missing space issue.

For information this does not happen when attributes for these two tags are added from the `index.template.html` file.

I'm just waiting for Ravzan feedbacks before pushing new commits to the PR.

Thanks @pdanpdan !

rstoenescu:
Good catch and good work.
Fix will be available in Quasar v2.1.6.

PR #11050, opened 2021-10-20
Added q-table body slot cols object definition

engineertdog:

- [ ] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
When using `q-table` and overriding the `body` slot, the `props.cols` should be defined as having a `name` and `value` property since all columns should have this basic common information.

Use case:
```js
    <template #body="props">
      <q-tr :props="props">
        <q-td
          v-for="col in props.cols"
          :key="col.name"
          :props="props">
            <span>{{ col.value }}</span>
        </q-td>
      </q-tr>
    </template>
```

### DIFF
--- a/ui/src/components/table/QTable.json
+++ b/ui/src/components/table/QTable.json
@@ -788,7 +788,17 @@
         "cols": {
           "type": "Object",
           "desc": "Column definitions",
-          "__exemption": [ "examples" ]
+          "__exemption": [ "examples" ],
+          "definition": {
+            "name": {
+              "type": "String",
+              "desc": "Column name"
+            },
+            "value": {
+              "type": "String",
+              "desc": "Column value for the given row"
+            }
+          }
         },
         "colsMap": {
           "type": "Object",


metalsadman:
i believe those props are already available from `colsMap`.
image

ui/src/components/table/QTable.json
```diff
@@ -788,7 +788,17 @@
...
+          "__exemption": [ "examples" ],
+          "definition": {
+            "name": {
+              "type": "String",
+              "desc": "Column name"
+            },
+            "value": {
+              "type": "String",
+              "desc": "Column value for the given row"
+            }
```
yusufkandemir, line 800: This would limit the object to just `name` and `value`, let's add a "rest" property with an explanation.
```suggestion
            },
            "...props": {
              "type": "Any",
              "desc": "Any other props you defined in 'columns'",
              "__exemption": [ "examples" ]
            }
```

ui/src/components/table/QTable.json
```diff
@@ -788,7 +788,17 @@
...
-          "__exemption": [ "examples" ]
+          "__exemption": [ "examples" ],
+          "definition": {
+            "name": {
+              "type": "String",
+              "desc": "Column name"
+            },
+            "value": {
+              "type": "String",
+              "desc": "Column value for the given row"
```
yusufkandemir, line 799: The value is not limited to `String`, it can be anything. So, let's match the type and description with `body-cell`.
```suggestion
              "type": "Any",
              "desc": "Parsed/formatted value for the given row",
              "__exemption": [ "examples" ]
```

ui/src/components/table/QTable.json
```diff
@@ -788,7 +788,17 @@
         "cols": {
           "type": "Object",
           "desc": "Column definitions",
-          "__exemption": [ "examples" ]
+          "__exemption": [ "examples" ],
+          "definition": {
+            "name": {
+              "type": "String",
+              "desc": "Column name"
+            },
```
yusufkandemir, line 796: Let's match the 'desc' with 'columns'.
```suggestion
            "name": {
              "type": "String",
              "desc": "Unique id, identifies the column",
              "examples": [ "desc", "userId" ]
            },
```
Also, `String` wouldn't work without `examples`, it's a required property in the schema.

yusufkandemir (changes requested):
Good work 👏
Don't forget to run `yarn build js api` to build/process the API. It will throw errors if the schema isn't correct. In this case, it would throw errors for missing `examples` prop. Running `yarn build js types` might be an even better idea because it lets you also see the impact on the generated types(`dist/types/index.d.ts`).

yusufkandemir:
Also, there are other props that will always exist, such as `label` and `field`. `cols` is also used in other scopes such as `header`, `body-cell`, etc. So, please also add them.

Another point, you will need to use `"required": "true"` because these will always exist. This also means that you can just add the whole definition from `columns` because it offers you all the types, but also lets you know which ones are always defined, and which ones may or may not be defined.

ui/src/components/table/QTable.json
```diff
@@ -788,7 +788,17 @@
         "cols": {
           "type": "Object",
```
yusufkandemir, line 789: I believe this should be an Array. This would mean the end type should be `{ name: string, value: any, ... }[]`.

rstoenescu:

PR #11117, opened 2021-10-26
chore(vite-plugin): Add type for transformAssetUrls

Shinigami92:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

The example https://quasar.dev/start/vite-plugin#using-quasar shows that we should use `vue({ template: { transformAssetUrls } })`, but when using `vite.config.ts` we cannot access `transformAssetUrls` without getting an error.

### DIFF
--- a/vite-plugin/index.d.ts
+++ b/vite-plugin/index.d.ts
@@ -32,10 +32,24 @@ export interface QuasarPluginOpts {
 
 export function quasar(opts?: QuasarPluginOpts): Plugin;
 
-export interface TransformAssetUrls {
-  base: unknown;
-  includeAbsolute: boolean;
-  tags: Record<string, string | string[]>;
+// `TransformAssetUrls` copy of https://github.com/vuejs/vue-next/blob/2c221fcd497d1541e667892ed908631df7e4a745/packages/compiler-sfc/src/compileTemplate.ts#L67
+export interface AssetURLTagConfig {
+  [name: string]: string[];
 }
 
-export const transformAssetUrls: TransformAssetUrls;
+export interface AssetURLOptions {
+  /**
+   * If base is provided, instead of transforming relative asset urls into
+   * imports, they will be directly rewritten to absolute urls.
+   */
+  base?: string | null;
+  /**
+   * If true, also processes absolute urls.
+   */
+  includeAbsolute?: boolean;
+  tags?: AssetURLTagConfig;
+}
+
+export type TransformAssetUrls = AssetURLOptions | AssetURLTagConfig | boolean;
+
+export const transformAssetUrls: AssetURLOptions;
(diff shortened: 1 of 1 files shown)


vite-plugin/index.d.ts
```diff
@@ -1,9 +1,9 @@
 // @ts-ignore
 /// <reference types="@quasar/vite-plugin" />
 
-import type { Plugin } from 'vite'
+import type { Plugin } from "vite";
 
-interface QuasarPluginOpts {
+export interface QuasarPluginOpts {
```
Shinigami92, line 6: It's safe to export these types/interfaces, devs can then use these

vite-plugin/index.d.ts
```diff
@@ -30,6 +30,12 @@ interface QuasarPluginOpts {
   // runMode?: "web-client" | "ssr-client" | "ssr-server";
 }
 
-export function quasar(
-  opts?: QuasarPluginOpts
-): Plugin;
+export function quasar(opts?: QuasarPluginOpts): Plugin;
+
+export interface TransformAssetUrls {
```
rstoenescu: I think it would be better if this was imported directly from vue (since a vue package defines it)

Shinigami92: Hey, thanks for the info! Didn't know that.
"Sadly" the definitions are in `compiler-sfc` and I think it would be overkill to add this as a peer dependency just for this "simple" type. So I copied it over and added a link. I think this is a good compromise.

rstoenescu:
Thanks for contributing!
This will go into "@quasar/vite-plugin" v1.0.2

vite-plugin/index.d.ts
IlCallo: If I recall correctly, compiler-sfc is now available as a nested import via `vue/compiler-sfc` or similar, since keeping those in sync for libraries was a pain. Dunno if they also export types from there, but worth taking a look

Shinigami92: I tried to import the type without adding a dependency, but it did not work / got a TS (or lint?) error

IlCallo: That's strange, vue now imports `@vue/compiler-sfc` dependency, so you should have it no matter what

image

Have you added `vue` and `vite` as devDependencies?
If you run `yarn -D vite vue@next`, you should be able to import the type as

```ts
import type { SFCTemplateCompileOptions } from "vue/compiler-sfc";

export type TransformAssetUrls = Required<SFCTemplateCompileOptions["transformAssetUrls"]>;
```

Not sure we support the whole type or only `AssetURLOptions`, haven't checked it

PR #11260, opened 2021-11-09
build(types): Improve type generation build script

yusufkandemir:

- Build-related changes

- No

- It's submitted to the `dev` branch

**Other information:**

When working with only the `build.types.js` file or JSON API files, it's extremely annoying to wait, even with a JS-only build. So, I added another parameter to the build script that allows building only the type-related files. This change reduced the build time from ~20s to ~3s on my machine.

Also, I added a diff-checking logic to output the diff of specific files. I used that for the `dist/types/index.d.ts` file on the types-only build. When working with types, it's extremely useful to see the impact on the generated types when updating JSON API files or `build.types.js`. 

**Before:**
image

**After(with no changes):**
image

**After(with changes):**
image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,6 +69,8 @@ $ yarn dev [theme] # or: npm run dev [theme]
 $ yarn build      # or: npm run build
 # build only js dist files
 $ yarn build js   # or: npm run build js
+# build only type related files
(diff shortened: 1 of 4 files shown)


rstoenescu:
Excellent idea. I'm gonna build upon it and add more subtypes.

PR #11278, opened 2021-11-10
docs(Electron): Added example code for disabling developer tools in production

MilosPaunovic:


rstoenescu (changes requested):
Good addition! Tweaked it a bit (the `if ()` part).
Can you also add it to https://github.com/quasarframework/quasar/blob/dev/app/templates/electron/electron-main.js pls?

rstoenescu (changes requested):
Everything looks good, except that the app has a blank screen??

rstoenescu (commented):
Also, isn't this clashing with:

```
mainWindow.webContents.on('devtools-opened', () => {
      mainWindow.webContents.closeDevTools()
    })
```

MilosPaunovic:
I've been looking at (and adjusting) the code on the `dev` branch as it is the same as the one on `v1` branch, which I'm using currently. It is not the same:

[https://github.com/quasarframework/quasar/blob/v1/app/templates/electron/main-process/electron-main.js](https://github.com/quasarframework/quasar/blob/v1/app/templates/electron/main-process/electron-main.js)

[https://github.com/quasarframework/quasar/blob/dce59c713ed4d0e9159d38351541204b93db7f2f/app/templates/electron/electron-main.js](https://github.com/quasarframework/quasar/blob/dce59c713ed4d0e9159d38351541204b93db7f2f/app/templates/electron/electron-main.js)

About the clash, `mainWindow.webContents.closeDevTools()` closes the opened DevTools, and the shortcut prevents 'em from opening it in the first place.

hawkeye64:
I agree this needs to be a bit better.

This code:
```js
mainWindow.webContents.on('devtools-opened', () => {
      mainWindow.webContents.closeDevTools()
    })
 ```
Causes a flash of dev-tools opening/closing on my MSI laptop screen (at 144hz), but doesn't on my Asus laptop screen (60hz).

yusufkandemir:
From https://stackoverflow.com/a/69957617
> Every answer mentions that the keyboard shortcut CTRL + SHIFT + I still works even after disabling devtools using devTools: false.
> This is because of the registered accelerator in the Electron BrowserWindow's default menu. All you have to do is remove the menu using `mainWindow.removeMenu()` and none of the development-related shortcut keys will work again. Even the ones like CTRL + R which reloads the page.

This sounds like a better and more stable solution to the problem.

hawkeye64:
> This sounds like a better and more stable solution to the problem.

I like it. I will have to test it out to make sure it is still something newer Electron versions can do.

PR #11284, opened 2021-11-11
Removed unnecessary console statements

MilosPaunovic:
Left here as part of the `QInfinteScroll` commit
[fix(QInfiniteScroll): loading slot after "done(true)" doesn't disappear](https://github.com/quasarframework/quasar/commit/ea0cb5434bdd92f9a11e9ef574464ab2d757f063)

### DIFF
--- a/ui/dev/src/pages/components/infinite-scroll.vue
+++ b/ui/dev/src/pages/components/infinite-scroll.vue
@@ -93,7 +93,6 @@ export default {
   },
   methods: {
     loadReverse (index, done) {
-      console.log('load reverse called', index)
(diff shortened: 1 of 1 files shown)


pdanpdan:
The logs are in dev :)

MilosPaunovic:
True, but it just keeps bugging me, as it looks like an unwanted leftover in the code, so the PR had to be created 😃

rstoenescu:
The point is to actually get those logs to ensure functionality.

MilosPaunovic:
Understood, I'll hanlde my mind differently 🙂

PR #11290, opened 2021-11-11
feat: new svg-ionicons-v6 svg icon set and relevant updates

hawkeye64:
- I am doing this as a PR for review.
- I fixed a number of places that referred to ionicons-v5 as a web font (which there is none)
- Added the svg-ionicons-v6 svg icon set
- Updated dev and docs


rstoenescu:
Awesome work!

Issue #11297, opened 2021-11-12
select clearable is not working (on a dialog)


ozelotdev:
**Describe the bug**
`q-select`'s clearable is not working. (on a dialog)

**Codepen/jsFiddle/Codesandbox (required)**
Editor: https://codesandbox.io/s/inspiring-stallman-ji8nt?file=/src/pages/Index.vue
Viewer: https://ji8nt.sse.codesandbox.io

**To Reproduce**
1. Open the page in mobile view.
2. Select "Example Selector" (a dialog should appear)
3. Select some items.
4. Select the clear button in the upper right corner.
5. Not working.

**Expected behavior**
The Clear button shown on the page will work.
The Clear button displayed on a dialog should work the same way.

**Screenshots**
<img width="395" alt="スクリーンショット 2021-11-12 11 09 16" src="https://user-images.githubusercontent.com/32789856/141396776-c13c0764-63f1-44f3-b0d7-37da0eefe870.png">

**Platform (please complete the following information):**
Quasar Version: v2.3.1
@quasar/app Version: 3.1.0

Quasar mode:
  - [x] SPA
  - [ ] SSR
  - [ ] PWA
  - [ ] Electron
  - [ ] Cordova
  - [ ] Capacitor
  - [ ] BEX

Tested on:
  - [x] SPA
  - [ ] SSR
  - [ ] PWA
  - [ ] Electron
  - [ ] Cordova
  - [ ] Capacitor
  - [ ] BEX

OS: Windows 10
Node: v14.18.1
NPM:6.14.15
Yarn:
Browsers: Firefox 94.0.1
iOS:
Android:
Electron:

**Additional context**
N/A

hawkeye64:
@ozelotjp Hi, thanks for finding this! I have been working on it most of the day and have a fix. I just need more testing to make sure there are no side effects with such a large component.

https://user-images.githubusercontent.com/10262924/141554173-dddc8581-f75d-4faa-9729-dc42c191b0a7.mp4

rstoenescu:
This is due to an underlying Vue breaking change.
Updated our code to Vue's latest specs.

Will be available in Quasar v2.3.2.

ozelotdev:
Thanks for this and other contributions!

PR #11317, opened 2021-11-12
fix(ui): select clearable is not working (on a dialog) (#11297)

hawkeye64:
The "state.splitAttrs.listeners.value" does not contain the listeners (like it used too. Did Vue make some undocumented changes?) as expected. I added a composable, "use-listeners.js" (quasar had something similar at start of v2). Anyway, it all works now - in a dialog/menu, or stand-alone desktop web page. I create a "dev/select-mobile" for testing this functionality. You can test before the merge to verify (it's already in dev)

### DIFF
--- a/ui/src/components/input/QInput.js
+++ b/ui/src/components/input/QInput.js
@@ -75,7 +75,7 @@ export default createComponent({
 
     const onEvents = computed(() => {
       const evt = {
-        ...state.splitAttrs.listeners.value,
+        ...state.listeners.value,
         onInput,
         onPaste,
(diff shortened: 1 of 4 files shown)


rstoenescu:
Seems like this was related to a breaking change in Vue itself.
In any case, since there is a useSplitAttrs() then that should be updated to new specs rather than partially duplicating its purpose.

hawkeye64:
> Seems like this was related to a breaking change in Vue itself. In any case, since there is a useSplitAttrs() then that should be updated to new specs rather than partially duplicating its purpose.

I didn't even think of updating that class. You humble me always with your knowledge and experience. Thanks.

PR #11352, opened 2021-11-15
fix(api/Notify): Remove 'type' from registerType's parameter

yusufkandemir:

- Bugfix

- No

**Other information:**
'type' isn't recursively handled, so it's misleading.

### DIFF
--- a/ui/src/plugins/Notify.json
+++ b/ui/src/plugins/Notify.json
@@ -667,12 +667,6 @@
           "type": "Object",
           "desc": "Notification options",
           "definition": {
-            "type": {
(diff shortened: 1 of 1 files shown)


yusufkandemir:
After a bit of thinking, I think it should be possible to actually handle the 'type' recursively. So, if we choose to do it instead, we can close this PR.
An example use case might be extending the default notification types such as `positive`, `warning`, etc.

rstoenescu:
The "type" should still be available, as you can define your custom "negative"/etc notification(s).
So let's go with the recursive stuff as that would be the right approach.

Issue #11475, opened 2021-11-25
QSelect does not emit 'virtual-scroll'


phgrund:
**Describe the bug**
The QSelect component does not emit the 'virtual-scroll' event in the new Composition API, thus triggering a warning.

**Codepen/jsFiddle/Codesandbox (required)**
Codesandbox: https://codesandbox.io/s/lucid-bartik-22k3k?file=/src/pages/Index.vue

**Steps to reproduce the behavior:**
1. Click on a QSelect with the 'virtual-scroll' event listener
2. Scroll the options
3. See warning

**Expected behavior**
No warnings triggered.

**Screenshots**
![image](https://user-images.githubusercontent.com/25393880/143487615-1f03243c-4a90-400f-be71-aec45642c6e3.png)

**Platform (please complete the following information):**
Quasar Version:
@quasar/app Version: 3.2.3

Quasar mode:
  - [X] SPA
  - [ ] SSR
  - [ ] PWA
  - [ ] Electron
  - [ ] Cordova
  - [ ] Capacitor
  - [ ] BEX

Tested on:
  - [X] SPA
  - [ ] SSR
  - [ ] PWA
  - [ ] Electron
  - [ ] Cordova
  - [ ] Capacitor
  - [ ] BEX

OS: Linux(5.10.16.3-microsoft-standard-WSL2) - linux/x64
Node: 14.17.5
NPM: 8.0.0
Yarn: 1.22.15
Browsers: Chrome v96.0.4664.45
iOS:
Android:
Electron:

yusufkandemir:
First of all, thanks for the bug report and the PR.

The same can also happen for `virtual-scroll` event for `QVirtualScroll` too and for other events such as `new-value` event for `QSelect`.

This case is normally handled by `onVirtualScroll` prop defined here: https://github.com/quasarframework/quasar/blob/77c9238ce7c074202173183e2e02d6bad87868fd/ui/src/components/virtual-scroll/use-virtual-scroll.js#L213
However, due to a bug in Vue, you are still getting the warning, see: https://github.com/vuejs/vue-next/issues/4803 it even has a PR https://github.com/vuejs/vue-next/pull/4804

So, since it's just a warning, we wouldn't want to add some workaround that can get messy over time. We would need to remember to apply the same workaround to every component that uses `useVirtualScroll`, and need to remember to remove the workaround when the issue gets fixed, etc.

[...]

PR #11477, opened 2021-11-25
fix(ui): QSelect does not emit 'virtual-scroll'

hawkeye64:
This is an alternative solution to #11475
This PR follows the current direction used by other composables (ie: useFieldEmits, etc)


yusufkandemir:
Closing, see the explanation in https://github.com/quasarframework/quasar/issues/11475#issuecomment-980692233

hawkeye64:
@yusufkandemir I did, and thanks for clarifying. Glad you're on the team. ❤️

Issue #11485, opened 2021-11-26
Type 'modelValue' of QSelectProps incomplete


Evertvdw:
**Describe the bug**
Currently with typescript it is not possible to have an object as modelValue as it is typed as:
`modelValue: number | string | any[] | null | undefined;`

**Codepen/jsFiddle/Codesandbox (required)**
This causes issues when using the QSelect like this:
https://codepen.io/Mett-codepens/pen/BadXZgj?editors=1010

**Expected behavior**
ModelValue should be typed to include Objects, like:
`modelValue: number | string | any[] | null | LooseDictionary |  undefined;`

**Platform (please complete the following information):**
Quasar Version: 2.3.3
@quasar/app Version: 3.2.3

yusufkandemir:
`type` here also needs `"Object"`:
https://github.com/quasarframework/quasar/blob/174f15fd4d29788da9b31f2f9beb79c3baa97af9/ui/src/components/select/QSelect.json#L11

Can you PR it, please? Don't forget to check if the final generated types are correct after running `yarn build js types`, it will output the diff to console.

Evertvdw:
Ok, will do!

PR #11492, opened 2021-11-26
fix(QSelect): accept object type as modelValue (fix: #11485)

Evertvdw:
@yusufkandemir 

- [x] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/ui/src/components/select/QSelect.json
+++ b/ui/src/components/select/QSelect.json
@@ -8,7 +8,7 @@
   "props": {
     "model-value": {
       "extends": "model-value",
-      "type": [ "Number", "String", "Array", "null", "undefined" ],
+      "type": [ "Number", "String", "Array", "Object", "null", "undefined" ],
       "desc": "Model of the component; Must be Array if using 'multiple' prop; Either use this property (along with a listener for 'update:modelValue' event) OR use v-model directive",
       "examples": [ "v-model=\"myModel\"" ]
     },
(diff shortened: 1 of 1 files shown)


ui/src/components/select/QSelect.json
```diff
@@ -8,7 +8,7 @@
   "props": {
     "model-value": {
       "extends": "model-value",
-      "type": [ "Number", "String", "Array", "null", "undefined" ],
+      "type": [ "Number", "String", "Array", "Object", "null", "undefined" ],
```
yusufkandemir: I just figured we are using `Any` in all the other places such as return type of `option-value` prop, `opt` references in scopes of slots, etc. even `@update:model-value` has `Any` implicitly. Also, logically there is nothing holding the users back from using bigint, symbol, etc. too. So, I think we can just simply use the type `Any` here.
```suggestion
      "type": "Any",
```

yusufkandemir (changes requested):
Thanks 👏 I suggested a change that is a fundamentally different solution. So, please consider it instead, if you don't have anything else to add.

Evertvdw:
Fine with me! I guess any will do in this case 👍

yusufkandemir (approved):
Looks better now 👏

rstoenescu:
Thanks for contributing!

PR #11506, opened 2021-11-27
fix(QIcon): Add check for lowercase 'm' for valid start of svg path

hawkeye64:
Some icon sets work perfectly well with lowercase 'm' and when adding `M0 0z` causes a black dot in upper-left of the icon. In this case, the regex needs to be `/^M|m/`

### DIFF
--- a/ui/dev/src/pages/components/icon.vue
+++ b/ui/dev/src/pages/components/icon.vue
@@ -19,6 +19,20 @@
       </q-icon>
 
       <q-icon name="android" size="5rem" />
+
+    </div>
+      <!-- This is the raw SVG -->
+      <q-icon color="secondary" size="5rem">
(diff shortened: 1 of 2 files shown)


hawkeye64:
I have tested this with the change in dev and it works perfectly fine. None of the other icon sets were disturbed.

rstoenescu:
`/^M|m/` fails when the string contains "m" anywhere inside of it.
Correct is `/^[Mm]/` --> pushed commit to fix it.

rstoenescu:
Also, this broke all mdi icons...
Fixed to `/^[Mm] ?\d/`

hawkeye64:
I forgot, this also needs to be done for v1

hawkeye64:
In v1, this seems to cause an issue for `mdi...` icons. Working on a fix.

hawkeye64:
https://github.com/quasarframework/quasar/pull/11583

I am going to re-review this update (into v2) to make sure there is no conflict with `mdi`

rstoenescu:
Fixed it already. And shipped it in latest Quasar

onurkose:
Hi, I have problems rendering simAmazon on Quasar v.2.3.4

`const mRE = /^[Mm] ?\d/` is not detecting simAmazon's beginning `M.045`

I believe it should be like this to support dots right after first character: `/^[Mm][ ?\d|.]/`

pdanpdan:
Of course it was a catch somewhere.
I suppose it could also be +- - I just hope `e` is not

hawkeye64:
Yes, this didn't work out the way intended. Before Quasar update:

image

Quasar v2.3.4 update:

image

The icons are being interpreted as `material-icons`

image

I'll dig into this.

hawkeye64:
https://github.com/quasarframework/quasar/pull/11584

hawkeye64:
Fix is also added into v1: https://github.com/quasarframework/quasar/pull/11583

haasz:
Hi,
When will there be a new release with the fix, that is, when will version 2.3.5 (or 2.4.0) be released?

rstoenescu:
@haasz It's coming this week. Next Quasar version is a big one, which is why it's taking a bit more time to be released.

haasz:
@rstoenescu Thanks for the reply.

PR #11584, opened 2021-12-07
fix(ui): QIcon interpretation of 'M.' as 'material-icons' is incorrect

hawkeye64:


hawkeye64:
Fix is also added into v1: https://github.com/quasarframework/quasar/pull/11583

rstoenescu:
Rewritten to more appropriate: `/^[Mm]\s?[-+]?\.?\d/`

hawkeye64:
I was adding that at the same time :)

rstoenescu:
On the Qv1 PR. There's one from Dan as well, backporting more stuff from Qv2 along the way.
One comment on the Qv1 PR -> rather than the greedy `\s*` I'd rather prefer the more performant match of "one space or nothing" `\s?`.

Is it ok for you that I merge Dan's PR? (cause it has more stuff to it) Otherwise please talk to each other and tell me how you want it, because I can't push both PRs without merge errors :)

hawkeye64:
Yes, merge Dan's PR, np

hawkeye64:
https://github.com/hawkeye64/quasar-extras-svg-icons/issues/18

PR #11586, opened 2021-12-08
feat(QIcon/lang): sync QIcon implementation and lang with Q2

pdanpdan:


ui/lang/ar.js
```diff
@@ -30,20 +30,18 @@ export default {
     noData: 'لا توجد بيانات',
     noResults: 'لا توجد نتائج',
     loading: 'جارٍ التحميل...',
-    selectedRecords: function (rows) {
-      return rows === 0
+    selectedRecords: rows => (
```
rstoenescu: Arrow functions will break IE11. The regular function syntax is used on purpose for Qv1.

ui/src/plugins/Notify.js
```diff
@@ -64,6 +64,14 @@ function logError (error, config) {
 const Notifications = {
   name: 'QNotifications',
 
+  // hide App from Vue devtools
+  devtools: { hide: true },
```
rstoenescu, line 68: Isn't this a Vue 3 feature of devtools?
Why is this needed?

ui/build/script.build.javascript.js
```diff
@@ -51,6 +51,7 @@ const builds = [
       },
       output: {
         file: resolve(`dist/quasar.common.js`),
+        exports: 'auto',
```
rstoenescu: I'd rather we not go through the upgrade hell of dev deps and testing all of the flavours of Quasar. This adds no benefit to the end user but adds possible failure points for us.
So please downgrade back all deps.

rstoenescu (changes requested):

ui/lang/ar.js
pdanpdan: Ok, will do, I forgot this is not transpiled

ui/src/plugins/Notify.js
pdanpdan: Vue devtools beta works with vue2, and it:
- shows the top level app
- borks on getting router for these plugins

ui/build/script.build.javascript.js
pdanpdan: Ok

PR #11651, opened 2021-12-12
feat(ts): Improve useDialogPluginComponent types

MT224244:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

When using `useDialogPluginComponent`, you can specify the type of the `onDialogOK` argument.

Also, if you don't specify it, `any` will be inserted, so it won't be broken.

```ts
const { dialogRef, onDialogOK } = useDialogPluginComponent<string>();
onDialogOK("foo"); // OK
onDialogOK(10); // Error
```

And since `ok` in `emitsObject` did not have an argument, I added one.
However, I couldn't think of a way to specify the type here, so it is `any`...

### DIFF
--- a/ui/types/composables.d.ts
+++ b/ui/types/composables.d.ts
@@ -6,15 +6,15 @@ import { Ref } from "vue";
 import { QVueGlobals } from "./globals";
 
 interface useDialogPluginComponent {
-  (): {
+  <T = any>(): {
     dialogRef: Ref<QDialog | undefined>;
     onDialogHide: () => void;
-    onDialogOK: (payload?: any) => void;
+    onDialogOK: (payload?: T) => void;
     onDialogCancel: () => void;
   };
   emits: ['ok', 'hide'];
   emitsObject: {
-    ok: () => true;
+    ok: (payload?: any) => true;
     hide: () => true;
   }
 }


ui/types/composables.d.ts
```diff
@@ -6,15 +6,15 @@ import { Ref } from "vue";
...
     dialogRef: Ref<QDialog | undefined>;
     onDialogHide: () => void;
-    onDialogOK: (payload?: any) => void;
+    onDialogOK: (payload?: T) => void;
     onDialogCancel: () => void;
   };
   emits: ['ok', 'hide'];
   emitsObject: {
-    ok: () => true;
+    ok: (payload?: any) => true;
```
yusufkandemir, line 17: Just some insights and thinking:

`onDialogOK` is basically emitting the given payload, and calls hide to properly hide the dialog. Although the consumers don't really supposed to be emitting this event themselves, we still may consider improving it, so `T` should have been used instead of `any`  if it was possible.

https://github.com/quasarframework/quasar/blob/5526dd9d82b7cbf4ee937c8bfbd030cc6ee3ef50/ui/src/composables/use-dialog-plugin-component.js#L21-L24

The only way to achieve such a thing would be creating the function form of `emitsObject`, but it wouldn't really be worth doing it since it will not be used directly anyways. So, `any` is the best thing possible in the current situation.

yusufkandemir (approved):
Thanks for the contribution. Looks good to me 👍 

It's not possible to get the payload type to `Dialog.create(...).onOk(...)` or `$q.dialog(...).onOk(...)` automatically, at least I am not aware of such possibility in general. But this is still very useful.

PR #11673, opened 2021-12-13
feat(QBreadcrumbs): improvement for breadcrumb elements

pdanpdan:
- add marker classes on breadcrumb el parent for when it's disabled/inactive or last
- don't apply activeClass on disabled breadcrumb els
- really disable last breadcrumb el (instead of using pointer-events)
- add visual indicator for active el when kbd navigating


rstoenescu (changes requested):
Please adapt from https://github.com/quasarframework/quasar/commit/925ba772a8f9c4b486dd614b283f0e5d6563559b

PR #11677, opened 2021-12-14
chore: use consistent --disabled modifier

pdanpdan:
In all other places except QTime/QUploader the modifier is '--disabled'


rstoenescu:
The consistency for CSS classes is to actually use the "ed" suffix (ex: "--bordered" etc).
Furthermore, we cannot add a breaking change like this even if it would be the correct way.

PR #11680, opened 2021-12-14
chore: fix linter complaints about indenting

pdanpdan:


rstoenescu:
I want it exactly how it is right now, for readability purposes. However, the linter rules are unpredictable -- sometimes it wants the ternary operator how it is, sometimes indentation rules take over. Had no time to tweak the rules (or possibly upgrade the linter deps in the hope that they fixed this) since functionality has precedence over nice-to-haves (with no benefit to the end users) atm.

Issue #8578, opened 2021-03-05
QTabs not properly working on RTL mode


hamza-sweid:
**Describe the bug**
When using QTabs in RTL mode the sliding arrows are not working…

**Codepen/jsFiddle/Codesandbox (required)**
https://codepen.io/Hamza-963/pen/WNoKXMO?editors=1111

**To Reproduce**
1. Go to link here: https://codepen.io/Hamza-963/pen/WNoKXMO?editors=1111
2. Click on arrows (left or right), nothing happens
3. Change direction to 'ltr' in data, arrows work normally

**Expected behavior**
When click on right arrow, should slide to right.
When click on left arrow, should slide to left.

**Screenshots**
![arrows](https://user-images.githubusercontent.com/57155313/110050979-b0687b80-7d5d-11eb-9e99-e2c8c00a0786.PNG)

**Platform (please complete the following information):**
Quasar Version:
@quasar/app Version:
Quasar mode:
  [ ] SPA
Tested on:
  [ ] SPA
OS: Windows
Node: v12.18.2
NPM: 6.14.5
Browsers: Chrome

pdanpdan:
It works fine with RTL - you have to include the rtl version of the CSS and the RTL must be set on html tag.

hamza-sweid:
Could u guide me how I can include the rtl version of the CSS, please!

rstoenescu:
https://quasar.dev/options/rtl-support

hamza-sweid:
I have really add it to my project, and all my app turn into RTL mode, navbar, components, and lang in HTML tag is set to 'ar', all my app is fine.
But still Qtab component arrows don't work, I really don't know what's the problem.
If someone can make a codePen which describe that, I will appreciate that.

pdanpdan:
You are right, something is wrong, checking it now.

maggie44:
I noticed there isn't a Qv2 tag on this issue, so just wanted to note that I am having the same problem in Quasar v2. It is a bug that prevents the page from being navigable.

AlaaDq:
@pdanpdan 
@hamza-sweid 
@maggie0002 

i have managed to solve it by editting this global  class in src\css\app.scss file .
`[dir='rtl'] .q-tabs.q-tabs--scrollable{
	direction: ltr !important;
}`


hope it help someone.

pdanpdan:
For Q1 you can use a fixed build.

In package.json replace quasar dependency with:
[...]

PR #11685, opened 2021-12-14
fix(QTabs): fix RTL mode; improve tabs scrolling #8578, #10034

pdanpdan:


rstoenescu:
Can you pls check if `tabs.hasCurrent` is really necessary? Qv2 doesn't actually needs it.
Also, can you remove the `detectBuggyRTL` references and replace them with `rtlHasScrollBug`?

pdanpdan:
`tabs.hasCurrent` is used to disable the tabIndex for the current tab - it's its only role
I see that in v2 this is no longer implemented

rstoenescu:
We only need tabindex for tabs when qtabs has no focus.

PR #11693, opened 2021-12-15
fix(docs): There is no longer a UMD starter kit for v2

hawkeye64:


rstoenescu:
Ha! Good catch. I've removed that starter kit ages ago :))

hawkeye64:
Had a question about it today from Graypes..

rstoenescu:
They should instead use the configurator I wrote.

hawkeye64:
Yes, on the same page.

PR #11811, opened 2021-12-28
(fix) QBreadcrumbs disabledValues typo

exp10der:


rstoenescu:
Hi, `void 0` is actually correct.

PR #11833, opened 2021-12-29
fix(docs): fix breakpoint terminology

hawkeye64:
The new way is the inverse of xs. The old way implied a missing pixel.


rstoenescu:
Good catch!

PR #11947, opened 2022-01-07
fix(QLayout): prevent showing scrollbars on resize

pdanpdan:


rstoenescu (changes requested):
Please adapt from https://github.com/quasarframework/quasar/commit/5db1b5088597da84c9a2f2702dadd10690d66350

PR #11954, opened 2022-01-08
fix(docs): Release names too wide

MilosPaunovic:
On the [Release notes](https://quasar.dev/start/release-notes) page, the `QTab` is too wide when the release name is too long, example is `quasar-v1.17.0 aka "Backport features and fixes from v2"`

Screenshot 2022-01-08 at 15 12 48
Screenshot 2022-01-08 at 15 12 58

Shoul be backported to `v1` docs as well.


rstoenescu:
Hi,

Thanks for contributing!
It would actually be better perf-wise to correctly populate the data from the beginning rather than running a function on each render. So fixed it like this: https://github.com/quasarframework/quasar/commit/4b4aa0d64cd94612a60ce42d916d69a68a3fab45

MilosPaunovic:
Totally agree!

Issue #3551, opened 2019-03-12
[v1 - Request] QTree to make option to disable unselecting already selected item


DmitrijOkeanij:
If I have QTree with standart selection mode (not tiked).

I can select some node and then second click on the same node and it will deselect it and reset QTree selection state to null.

It is very strange. QSelect, for example on 2 ckick on the same item just stay selected on it.

I think it needed to disable 2 click on item with unselecting it.
Or if this behavior is suspected, to do option to disable it.

Why I need it?

Is very usual case when i have to choose one of tree and not null.

Now to make this functionality bu custom is difficult.

My code to make work it:
```
<q-tree @update:selected="categorySelected" 
                  :selected.sync="treeSelectValue"
                  :nodes="nodes" />

categorySelected(key) {
    if(!key)
    {
        let old = this.treeSelectValue;
        this.$nextTick(() => {this.treeSelectValue = old});
    }
}
```

But I think this need to be done more clear, simple and natively has to be natively

gillisig:
I had the same use case and your code worked perfectly.  Thanks @Dmitrij-Polyanin

yarox:
I also had the same problem, but made it work watching the `treeSelectValue` variable:

```
...
watch: {
    treeSelectValue: function (newVal, oldVal) {
      if (!newVal) {
        this.treeSelectValue = oldVal
      }
    }
  },
...
```

Dmitrij-Polyanin:
Very good, more clean then my code.

pdanpdan:
It's very simple:

```html
<q-tree :nodes="nodes" :selected="selected" @update:selected="updateSelected" />
```

```javascript
updateSelected (key) {
  if(key !== null) {
    this.selected = key
}
```

PR #12012, opened 2022-01-12
fix: Adding an extra props "unSelectable" to QTree (fix #7701)

### LINKED ISSUE #7701: When second click to selected node on QTree selected key becomes null
Hi,
I use selected.sync and expanded.sync prop. After second click to selected node than selectedKey value becomes null. I think selected-key must stay the same.

super-seo:
Hi.
Example
1.https://codepen.io/superseo/pen/XWeyBpz
2.https://codepen.io/superseo/pen/oNGQMQO?editors=101

Taken from the manual https://quasar.dev/vue-components/tree

The second click on the selected element always resets the selection. That is, click on the element for the first time - the corresponding Tab Panel opens, click again on the same element - it closes.
But it is necessary that the second click on the same element does not cause deselection.

Adding an extra props "unSelectable" is probably enough. By default (true) - old behavior. If false, then the second click does not deselect. For those cases when such behavior is necessary, without breaking the old logic.

https://codepen.io/superseo/pen/xxXQJLz?editors=101

added a new parameter, for example on PR fix: Adding an extra props "unSelectable" to QTree (fix #7701) #12012
just use <q-tree ... :un-selectable="false" />
The error from the description occurs because of a check on this line
https://github.com/pdanpdan/quasar/blob/8a5a2d799bc4af2dc185ca0ec1614b7dd66737aa/ui/src/components/tree/QTree.js#L607

Same request but for v1
#3551

- [ ] Bugfix
- [*] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [*] No

- [*] It's submitted to the `dev` branch (or `v[X]` branch)
- [*] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)

**Other information:**

### DIFF
--- a/ui/src/components/tree/QTree.js
+++ b/ui/src/components/tree/QTree.js
@@ -46,6 +46,10 @@ export default createComponent({
     controlColor: String,
     textColor: String,
     selectedColor: String,
+    unSelectable: {
+      type: Boolean,
+      default: true
+    },
 
     icon: String,
 
@@ -615,7 +619,11 @@ export default createComponent({
 
       if (hasSelection.value) {
         if (meta.selectable) {
-          emit('update:selected', meta.key !== props.selected ? meta.key : null)
+          if (props.unSelectable) {
+            emit('update:selected', meta.key !== props.selected ? meta.key : null)
+          } else if (meta.key !== props.selected) {
+            emit('update:selected', meta.key || null)
+          }
         }
       }
       else {
--- a/ui/src/components/tree/QTree.js
+++ b/ui/src/components/tree/QTree.js
@@ -46,10 +46,6 @@ export default createComponent({
     controlColor: String,
     textColor: String,
     selectedColor: String,
-    unSelectable: {
-      type: Boolean,
-      default: true
-    },
 
     icon: String,
 
@@ -61,6 +57,8 @@ export default createComponent({
     ticked: Array, // v-model:ticked
     expanded: Array, // v-model:expanded
     selected: {}, // v-model:selected
+
+    noSelectionUnset: Boolean,
 
     defaultExpandAll: Boolean,
     accordion: Boolean,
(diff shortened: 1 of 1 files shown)


rstoenescu:
Hi,

Thanks for contributing!
Changed the implementation though:
1. Boolean props should retain their default "false" value because this allows for easier usage in templates. Just the presence of that prop should be enough (`no-selection-unset` vs `:no-selection-unset="true/false"`).
2. Improved the emit logic.

sbscan:
> Hi,
> 
> Thanks for contributing! Changed the implementation though:
> 
> 1. Boolean props should retain their default "false" value because this allows for easier usage in templates. Just the presence of that prop should be enough (`no-selection-unset` vs `:no-selection-unset="true/false"`).
> 2. Improved the emit logic.

When we use selectable nodes, we can only open the node by clicking the arrow. I think UX will be better if it's flagged selectable:false clicking it should open the node. Specially on mobile trying to click the arrow is not a good practice.

super-seo:
> > Hi,
> > Thanks for contributing! Changed the implementation though:
> > 
> > 1. Boolean props should retain their default "false" value because this allows for easier usage in templates. Just the presence of that prop should be enough (`no-selection-unset` vs `:no-selection-unset="true/false"`).
> > 2. Improved the emit logic.
> 
> When we use selectable nodes, we can only open the node by clicking the arrow. I think UX will be better if it's flagged selectable:false clicking it should open the node. Specially on mobile trying to click the arrow is not a good practice.

Hi.
This edition is not about list disclosure. This is a refinement about the ability not to remove the selection on the second repeated click. You should probably output a separate task/request.

However, in your case with selectable=true you can watch the emit 'update:selected' and use the tree's setExpanded(selected, true) method.

Issue #12032, opened 2022-01-14
q-date navigation-min-year-month error


HyeongRae:
### What happened?

q-date components make error if set default-view='Months', emit-immediately='true' and navigation-min-year-month '2019/12' or '2021/8' or '2020/11' or etc.
In these months, the first day is on a Sunday.

### What did you expect to happen?

not make error.

### Reproduction URL

https://codepen.io/hyeongrae/pen/poWqLaN

### How to reproduce?

1. click date
2. click '<' btn until 2019 year
3. click  DEC
4. find error

### Flavour

Quasar CLI (@quasar/cli | @quasar/app)

### Areas

Components (quasar)

### Platforms/Browsers

Chrome

### Quasar info output

```shell
Operating System - Darwin(21.1.0) - darwin/x64
NodeJs - 16.13.1

Global packages
  NPM - 8.1.2
  yarn - 1.22.17
  @quasar/cli - 1.1.3
  @quasar/icongenie - Not installed
  cordova - Not installed

Important local packages
  quasar - 2.4.3 -- Build high-performance VueJS user interfaces (SPA, PWA, SSR, Mobile and Desktop) in record time
  @quasar/app - 3.2.6 -- Quasar Framework local CLI
  @quasar/extras - 1.12.2 -- Quasar Framework fonts, icons and animations
  eslint-plugin-quasar - Not installed
  vue - 3.2.26 -- The progressive JavaScript framework for building modern web UI.
  vue-router - 4.0.12
  vuex - 4.0.2 -- state management for Vue.js
  electron - Not installed
  electron-packager - Not installed
  electron-builder - Not installed
  @babel/core - 7.16.7 -- Babel compiler core.
  webpack - 5.65.0 -- Packs CommonJs/AMD modules for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand. Support loaders to preprocess files, i.e. json, jsx, es7, css, less, ... and your custom stuff.
  webpack-dev-server - 4.7.1 -- Serves a webpack app. Updates the browser on changes.
  workbox-webpack-plugin - 6.4.2 -- A plugin for your Webpack build process, helping you generate a manifest of local files that workbox-sw should precache.
  register-service-worker - 1.7.2 -- Script for registering service worker, with hooks
  typescript - 4.5.4 -- TypeScript is a language for application scale JavaScript development
  @capacitor/core - Not installed
  @capacitor/cli - Not installed
  @capacitor/android - Not installed
[...]

PR #12041, opened 2022-01-14
fix(QDate): filter out empty daysMap #12032

pdanpdan:
#12032


rstoenescu:
This only masks the real problem and leaves room for many other scenarios to fail

PR #12047, opened 2022-01-14
tests(ui): add unit testing boilerplate and first tests

Evertvdw:

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Testing!

- [ ] Yes
- [x] No

**Other information:**


ui/test/cypress/plugins/unit.js
```diff
@@ -0,0 +1,28 @@
...
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+import { injectDevServer } from '@quasar/quasar-app-extension-testing-e2e-cypress/cct-dev-server'
+
+const {
+  addMatchImageSnapshotPlugin
+} = require('cypress-image-snapshot/plugin')
```
IlCallo: Can't this be imported using `ES` import syntax as we do for `injectDevServer`?

ui/test/cypress/support/commands.d.ts
```diff
@@ -0,0 +1,38 @@
+/// <reference types="Cypress" />
```
IlCallo, line 1: Do you think these commands could be useful for other devs too? If yes, we could add them into the Cypress AE

ui/test/cypress/helpers/create-spec.js
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env node
+const argv = require('yargs/yargs')(process.argv.slice(2))
```
IlCallo, line 2: Can you add a comment here (or into Cypress README) explaining how to use this and when it's needed?

IlCallo (changes requested):
Gave a first brief look and left some comments, will need to check this more in deep, but of course great work :)

rstoenescu:
Awesome! Looking forward to merge this after comments above are resolved.

ui/test/cypress/plugins/unit.js
Evertvdw: Yes it can! Will change

ui/test/cypress/support/commands.d.ts
Evertvdw: The checks for vertical/horizontal position are specific to quasars positioning system for popups. I don't think that is very useful. The `isNotActionable` could be useful. When calling .click() on a covered element it will normally fail the test if that element is covered. This command is the reverse of that, it will fail if the .click() succeeds.

ui/test/cypress/helpers/create-spec.js
Evertvdw: Yes, done!

IlCallo (approved):

IlCallo:
There is a weird problem with a faulty version of Cypress AE I published with the wrong name some months ago, and which apparently I'm not able to unpublish
This happens since we don't have a committed yarn.lock

@rstoenescu if you have permits, could you remove [this version](https://www.npmjs.com/package/@quasar/quasar-app-extension-testing-e2e-cypress/v/4.0.0-beta-4)?
Seems like the `-` instead of `.` of all other versions is making people land on that one instead of newer beta releases

Anyway, since it was due some weeks ago, I directly released Cypress AE v4 as stable and used that one instead :)

---

I had to rebase on latest changes and adapt it to the externalization of vue and vue-router
I also added a `yarn test:unit:ci` for when we'll need to run it on GH pipeline
I ignored `screenshots` and `videos` folders since we don't want to commit those

Everything seems to run pretty smoothly, except a test into QSelect which is failing and the snapshots images which appear to not match anymore, as new ones are created into a different folder now.
@Evertvdw can you check out those 2 things?

PR #12087, opened 2022-01-18
feat(keepAlive activate): set autofocus on activate in QField/QForm; restore scroll on activate in QScrollArea

pdanpdan:


ui/src/components/virtual-scroll/use-virtual-scroll.js
```diff
@@ -691,7 +691,15 @@ export function useVirtualScroll ({
...
 
+  let shouldActivate = false
+
+  onDeactivated(() => {
+    shouldActivate = true
+  })
+
   onActivated(() => {
+    if (shouldActivate !== true) { return }
+
```
rstoenescu, line 702: Why is there a flag needed? onActivated() is always triggered only after an onDeactivated().

Issue #7329, opened 2020-06-30
App extension typescript support


stefanvanherwijnen:
It would be nice if quasar would support app extensions written in typescript.

At least the following should be considered:
- [ ] Make `quasar create my-ext --kit app-extension` ask the user if he wants to write his AE in typescript, and if so, scaffold all required typescript tools and code into the AE.
- [ ] If typescript is used, JS should still be supported. For this the code should probably be transpiled. For files in the `templates` folder there should probably be a way to transpile them to JS after rendering them (a better but more complicated approach than https://github.com/quasarframework/quasar/pull/7297).
- [ ] Since quasar handles the install/index/prompts API files, there should be a way to determine if they are written in typescript and handled accordingly.

webnoob:
@IlCallo Is this something already on the cards or have we ruled it out for some reason?

stefanvanherwijnen:
@IlCallo 
Here is an updated starter-kit which somewhat works for typescript, but I believe the types are missing for the `api` parameter in Quasar so I can't test if it compiles yet:
https://github.com/stefanvanherwijnen/quasar-starter-kit-app-extension

IlCallo:
Yes, I requested this to keep track of the idea.
I think it could be an interesting feature, but would probably need the AE system to be rewritten into TS (or at least provide typings).
Gonna add this to my backlog and check it going forward, even if not in the short term I'm afraid.

@stefanvanherwijnen if you want to spend some time into this stuff, I can assist you to start adding types for AE system as I already did for quasar/app (via wrapper functions I'd say)

stefanvanherwijnen:
@IlCallo 
For some reason I started with the internal types... which made no sense so I just commented them out:
https://github.com/stefanvanherwijnen/quasar/blob/app-extension-types/app/types/app-extension.d.ts

Using the types with the following modified starter kit seems to work for .ts files:
https://github.com/stefanvanherwijnen/quasar-starter-kit-app-extension-typescript

It transpiles the boot files to JS so nothing has to be changed to Quasar internally.
However, I can't figure out how (or if it's possible at all) to transpile TS in .vue files to JS. I tried tsc and babel, but they don't like the HTML part of the Single File Components.
My idea was to be able to write an AE's template SFC's using the composition api and typescript, and transpile them back to JS when TS is not used in Quasar. This would prevent the need to write all template Vue files twice, one for JS and one for TS.

IlCallo:
[...]

PR #12090, opened 2022-01-18
fix(ae): fix file resolution for AEs with mixed dist folders

IlCallo:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
#9758 assumed that a TS-using AE would have all its files into a dist folder referenced into the package.json, but this is not the case for existing AEs which already return a compiled dist (eg. testing AEs), which still have Quasar-specific scripts into `src` folder.

This is especially tricky since a common file name for dist entrypoints is `index`, which is the same used by Quasar for it's index script file. This tricked Quasar to execute the wrong `index.js` file.

This PR reverse the assumption to make the change opt-in: a file is first searched into the default location, and searched into the dist folder only if not found into the default location

### DIFF
--- a/app/lib/app-extension/Extension.js
+++ b/app/lib/app-extension/Extension.js
@@ -125,10 +125,7 @@ module.exports = class Extension {
 
   isInstalled () {
     try {
-      const pkgSrc = this.__getPkgSrc()
-      require.resolve(`${this.packageName}/${pkgSrc}/index`, {
-        paths: [ appPaths.appDir ]
-      })
+      this.__resolvePackageFile('index')
     }
     catch (e) {
       return false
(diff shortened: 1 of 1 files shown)


IlCallo:
@stefanvanherwijnen can you test this change with your setup? I want to merge it asap as it's blocking users to install testing AEs (and possibly other AEs) on new projects

rstoenescu:
Do not merge this. Also, please do test PRs before merging.

rstoenescu:
The support for TS should be accompanied by a PR to quasar-starter-kit-ui and quasar-starter-kit-app-extension repos too. And then a docs mention. Otherwise it's completely hidden for our devs.

stefanvanherwijnen:
I thought both starter kits had the same package.json but this isn't the case:
https://github.com/quasarframework/quasar-starter-kit-app-extension/blob/master/package.json
https://github.com/quasarframework/quasar-starter-kit-ui/blob/master/template/app-extension/package.json

So my getPkgSrc function indeed does not work with the app-extension kit which I didn't realize...

@rstoenescu Before updating the starter kits some things (such as this issue) should be cleared up and fixed first. Right now it is hardcoded that the 'compiled' files are inside `src`, which is not the case for a TS project. Another major issue is the `render()` function which cannot easily support both rendering of JS and TS files without keeping equal copies of both.

@IlCallo 
For now you can just revert to the old behavior of hardcoded `src`. It is confusing but it works fine when tsc outputs to `src`.

rstoenescu:
I've reverted and pushed a version which first looks into /dist/ then in /src. Released q/app v3.2.8 with it.
The package.json is identical. You're looking at the wrong one for app-extension starter kit. https://github.com/quasarframework/quasar-starter-kit-app-extension/blob/master/template/package.json

I understand the reasoning for these changes when it comes to adding support for TS, but a master plan needs to be developed first. My main issue here is that this got merged without being tested and it broke people's AEs, forcing me to do an urgent release with a fix, which is not ok.

Regardless, like I mentioned, the master plan needs to first be developed. From top to bottom, how will this work. There's indeed the `render()` method and possibly other that need a way to be solved. Then we'll have a green light for anyone to implement.

IlCallo:
My fault about merging, I focused mostly on the types fixes when testing the PR (and wrongly didn't check them into the code after the last revision) and didn't spot the problem it would cause to some AEs

Gonna be more cautious next time

stefanvanherwijnen:
I'll also properly test things first next time.

@rstoenescu You're right about the master plan and I opened an issue for it a while ago:  https://github.com/quasarframework/quasar/issues/7329

Personally I came to the conclusion that proper TS support is likely not possible with the current build tools and I ended up switching to Vite. With this PR users should be able to create their own TS project setup and if anyone figures out a proper way to solve all problems it would be great if they would comment on that issue.

Apologies for breaking things 😅

Issue #5432, opened 2019-10-31
QForm.submit and QForm.validate do not respect async rules


holtalanm:
**Describe the bug**
A form with async rules.  Calling submit on that form when one or more of those rules would result in an error still submits the form.

**Codepen/jsFiddle/Codesandbox (required)**
https://codepen.io/holtalanm/pen/eYYeygM

**To Reproduce**
1. Enter a value into the Name field
2. Wait for it to auto-submit
3. Notice age field failed validation, but submit notification was still shown, indicating that the form was submitted.

**Expected behavior**
Age field validation should fail, but submit notification should NOT be shown, as the form should not be submitted when there is failed validation.

Contrast this behavior with synchronous validation:
https://codepen.io/holtalanm/pen/ZEEaoao

**Platform (please complete the following information):**
OS: Windows 10
Browsers: Firefox 70

rstoenescu:
Hi,

Thanks for reporting!
Fix will be available in "quasar" v1.3.0.

holtalanm:
Thank you!

OfekA-IAI:
@rstoenescu This is happening to me with Quasar 2.4.4 - any fix for this?

BenJackGill:
I'm having the same problem.

When my input has an async rule validation the form will submit even though it should throw an error.

OfekA-IAI:
@BenJackGill I wasn't able to find the cause of the issue, but in my case I was able to "solve" it like this:

I had a parent and a child component.
The parent had the form components and the `.validate()` call.
The child had a loading indicator which was triggered by a prop from the parent.

After removing that prop that was changed before and after validation ,`.validate()`  seems to work correctly

I still think there's some other issue here but couldn't spend the time to find out.

Hope this helps

holtalanm:
I ended up 'solving' this issue by rolling my own validation, in the end.

BenJackGill:
In my case the QForm was submitting before the QInput async validation had finished.

So I used the async validation inside the `@submit` of the form instead. Now it works.

This bug still needs looking at though.

@rstoenescu Can we reopen this issue?

rstoenescu:
This works correctly. Just checked it.
Please fill a codepen so I can help.

BenJackGill:
**Describe the bug:** A form with async rules will submit even if it has validation errors. This only happens on first page load and when pressing the enter key. All subsequent form submits work as expected.

**Codepen/jsFiddle/Codesandbox (required)**
[...]

PR #12292, opened 2022-02-01
fix(validation): prevent blur on QField to trigger another validation when one is in progress #5432

pdanpdan:
#5432


rstoenescu:
Partial fix. Not tackling the full problem.

PR #12333, opened 2022-02-03
feat(QTree, QExpansionItem): only render collapsed content when first expanded

pdanpdan:


rstoenescu:
The idea is good for QExpansionItem, but the main concern here is QTree memleak, even though you're using WeakMap. It's especially obvious on SSR builds, where the accumulator will keep on pumping values on each render. Some other mechanism would need to be used for QTree.

pdanpdan:
I cannot reproduce the problem. For me it looks like normal variations in GC in node.

Issue #12466, opened 2022-02-14
Splitter component is not keyboard-accessible


pbb72:
### What happened?

Keyboard users are not able to control the splitter component.

### What did you expect to happen?

Using the tab key to navigator through the document, the splitter component should receive focus at the appropriate moment. This focus state should in some way be visible. When the splitter has focus, it's position should be controllable using the keyboard, for example using arrow keys.

### Reproduction URL

https://codepen.io/pbb/pen/bGYryEB

### How to reproduce?

1. Go to reproduction URL
2. Place focus on the "Before" panel by clicking the mouse in it.
3. Press tab to navigate to the next focusable element.
4. Notice that the splitter is not getting focus.

### Flavour

Quasar CLI (@quasar/cli | @quasar/app)

### Areas

Quasar CLI Commands/Configuration (@quasar/cli | @quasar/app)

rstoenescu:
Fix will be available in v2.25

PR #12474, opened 2022-02-14
feat(QSplitter): add keyboard navigation #12466

pdanpdan:
#12466


yusufkandemir:
@pdanpdan do you remember the reason why this got closed?

pdanpdan:
There were some PRs that were some for V1 and not merged or implemented in other ways.
But on this case I think it was replaced with https://github.com/quasarframework/quasar/pull/14576

Issue #12383, opened 2022-02-07
q-field - custom atributes are removed


vanillamasonry:
### What happened?

q-field rendering remove custom attributes

### What did you expect to happen?

ex:
`<q-field  custom_att="test">  </q-field>
is processed as
<label class="q-field row no-wrap items-start q-field--standard q-field--auto-height" for="f_9e49ae40-33e9-490e-ad07-70dfcbdc5323"><div class="q-field__inner relative-position col self-stretch"><div class="q-field__control relative-position row no-wrap" tabindex="-1"><div class="q-field__control-container col relative-position row no-wrap q-anchor--skip"></div></div></div></label> `

custom_att="test" is missing from label or any other tag

### Reproduction URL

https://codepen.io/Paul_1001/pen/wvPgBaq

### How to reproduce?

1. Go to reproduction link
2. Move the mouse over the field
3. Click inspect element (firefox)
4. See --> label or any other div tag do not contain the  custom_att="test"

### Flavour

UMD

### Areas

Components (quasar)

### Platforms/Browsers

Firefox

soulsam480:
Hey, I think it's expected as `inheritAttrs` is set to false in QField. https://github.com/quasarframework/quasar/blob/160e7247e204ffe3988b422837cdcd08542b35ef/ui/src/components/field/QField.js#L8 . Are you aware of this ?

vanillamasonry:
Yes, but I do not think that this is the normal behavior. In fact I think  that ALL components must inherit custom attributes

soulsam480:
> 

That's true, but it definitely works for me when I pass custom attrs, as I can see `data-cy` attrs in the DOM in my app

rstoenescu:
Improvement will be available in Quasar v2.6.0.

PR #12485, opened 2022-02-14
fix(QField/QSelect): improve attrs on controls #12383

pdanpdan:
#12383

- QField: apply on external label if control slot is not present and if not QInput/QSelect
- QSelect: apply on active element (target)


rstoenescu:
Is there a PR for Qv2 also?

pdanpdan:
Not yet, I have to check the code

rstoenescu:
I'd like to merge them both simultaneously, otherwise I'll forget about it for Qv2.

PR #12578, opened 2022-02-23
fix: MAX_SAFE_INTEGER rounding when doing JSON.parse on a string cont…

bartduisters:
fix: MAX_SAFE_INTEGER rounding when doing JSON.parse on a string containing a number greater than MAX_SAFE_INTEGER

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

When using the Cookie plugin, if a string is stored that contains a number higher than MAX_SAFE_INTEGER, then a string will be returned. In the old behavior a number would be returned that contains a rounding error.

Therefor, if the number in the string is MAX_SAFE_INTEGER or higher, a string will be returned. This then can be handled accordingly in the application itself.

Example:
"69" will return a number 69
MAX_SAFE_INTEGER - 1, "9007199254740990" will return a number "9007199254740990"
MAX_SAFE_INTEGER, "9007199254740991" will return a string "9007199254740991"
MAX_SAFE_INTEGER + 1, "9007199254740992" will return a string "9007199254740992"

Old behavior vs new behavior:
Screenshot 2022-02-23 at 18 41 13

### DIFF
--- a/ui/src/plugins/Cookies.js
+++ b/ui/src/plugins/Cookies.js
@@ -28,7 +28,7 @@ function read (string) {
   string = decode(string.replace(/\+/g, ' '))
 
   try {
-    const parsed = JSON.parse(string, reviverFn)
+    const parsed = JSON.parse(string)
 
     if (parsed === Object(parsed) || Array.isArray(parsed) === true) {
       string = parsed
--- a/ui/src/plugins/Cookies.js
+++ b/ui/src/plugins/Cookies.js
@@ -28,7 +28,14 @@ function read (string) {
   string = decode(string.replace(/\+/g, ' '))
 
   try {
-    string = JSON.parse(string)
+    const value = JSON.parse(string);
+    if (typeof value === 'number') {
+      if (value < Number.MAX_SAFE_INTEGER) {
+        string = value;
+      }
+    } else {
+      string = value;
+    }
   }
   catch (e) {}
 
--- a/ui/src/plugins/Cookies.js
+++ b/ui/src/plugins/Cookies.js
@@ -28,13 +28,10 @@ function read (string) {
   string = decode(string.replace(/\+/g, ' '))
 
   try {
-    const value = JSON.parse(string);
-    if (typeof value === 'number') {
-      if (value < Number.MAX_SAFE_INTEGER) {
-        string = value;
-      }
-    } else {
-      string = value;
+    const parsed = JSON.parse(string, reviverFn)
+
+    if (parsed === Object(parsed) || Array.isArray(parsed) === true) {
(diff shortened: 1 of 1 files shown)


pdanpdan:
~~I think it would be better to parse in a `temp` variable and only assign the the `temp` to `string` if it is object or array~~

Disregard the above, you are not setting a string, you are setting a number.
In your case the right way is to set a string, so you'll get back a string

```javascript
SessionStorage.set('testS', '19007199254740992')
SessionStorage.set('testN', 19007199254740992)

console.log(SessionStorage.getItem('testS'))
console.log(SessionStorage.getItem('testN'))
```

bartduisters:
I've made this PR based on some discussion in Discord. I'll point them to this reply.

bartduisters:
But before I point them to this reply: It is saved as a string. Then it goes through the JSON.parse() function, which then converts it to a number (which has a wrong value because of the MAX_SAFE_INTEGER rounding that happens).

The second one wouldn't even save the correct value, since 19007199254740992 would be evaluated to 19007199254740990.

pdanpdan:
Sorry, I mixed up cookies with storage - you are right that the values are broken. But I think it would be better to go with a predictable conversion, like:

```javascript
try {
    const parsed = JSON.parse(string, reviverFn)

    if (parsed === Object(parsed) || Array.isArray(parsed) === true) {
      string = parsed
    }
  }
  catch (e) {}
```

Do you happen to know (or have links to code) on how different backends process / convert cookie values?
I would expect most of them to stick to string conversion by default

pdanpdan:
Or maybe better align with https://github.com/expressjs/cookie-parser/blob/dda4c5be5a97fbf4552ff54889715581c1c93128/index.js#L83

bartduisters:
I don't have any code. I just saw a comment in the Discord channel, know that this is opensource, and suggested a fix.

That said, I would expect that this is already solved before and that there is a dependency handling all the possible cases.

My suggested fix is just: everything keeps working as before, except for the use case where you have a number that is MAX_SAFE_INTEGER or higher, then you need to do some manual parsing in the frontend application. Which is better than getting an incorrect number.

The function you send from that other repo has the same issue, they just have a `j:` prepended that gets stripped.

With 'Align with', do you mean that I should start a conversation over there about the MAX_SAFE_INTEGER? I'm willing to do so, I just need to know what the expected outcome is for this repo, so that I know what I am working towards.

rstoenescu:
Thanks for contributing!

PR #12620, opened 2022-02-28
fix: Tweak regex for JSFiddle links (fix #12612)

MilosPaunovic:
Closes #12612

PR #12644, opened 2022-03-03
fix: Add Quasar docs URL as valid to GH Actions

MilosPaunovic:
As an extension of the work on [https://github.com/quasarframework/quasar/pull/12620](https://github.com/quasarframework/quasar/pull/12620), I've noticed that there is a significant amount of issues created with links from Quasar offical docs page.

This PR is meant to allow those kinds of links as valid, e.g. not report missing/invalid link comment by GH bot.
If you think it's unnecessary, I'll close it.

### DIFF
--- a/.github/workflows/process-created-issue.yml
+++ b/.github/workflows/process-created-issue.yml
@@ -123,6 +123,10 @@ jobs:
                       labelsToAdd.push('bug/1-hard-to-reproduce');
                       break;
                     }
+                  case 'quasar.dev':
+                    if (/^\/.+$/.test(reproURL.pathname)) {
+                      break;
+                    }
                   default:
                     throw new Error();
                 }


yusufkandemir:
The type of repro links the users should send is clearly stated when creating a "Bug Report"(_it's valid with "Docs Report"_). So, they shouldn't be using Quasar docs links anyways. I don't remember any discussion with valid points either.

Some people are also trying to say "repro not required", "it's straight-forward", etc. which sounded fair at first, but as I triaged more and more issues, I am now even more sure that a repro is always required and useful. Some even noticed afterward that the condition was valid and they should've been created a repro. So, this kind of proves that some people are doing something doesn't necessarily mean it's correct or useful.

Some people are just sending whole documentation page links or just sections. But, this is not really useful when trying to reproduce the issue. Also, it's not really bound with a specific version that can always be consistently reproduced, it's bound to the docs. So, if the docs get a content/code update, or Quasar package update, etc. it can get fixed or get worse without noticing, etc. Maybe the problem was caused by another thing in the docs specific code, layout, etc. But, if it was a Codepen, we could eliminate all other factors and bind the problem to a specific version of Quasar. Then, we could easily play around with it, switch versions to see if that changes anything, and eliminate external factors.

If there is a problem with the docs-specific code, then Docs Report should be used instead, where it's totally legit to send Quasar docs links.

So, to sum up, I don't see any value in that. If you see any valid and useful points, we can discuss them and we might consider allowing Quasar docs URL too.

MilosPaunovic:
Completely agree with the explanation; closing.

PR #12701, opened 2022-03-08
fix(ui): QInfiniteScroll poll.cancel is undefined when debounce prop equal to 0

kovalchukq:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
When debounce prop equal to 0 and immediatePoll function assigned.
https://github.com/quasarframework/quasar/blob/712ccc84239e1673e90aa8db3b86ebc4d12c8160/ui/src/components/infinite-scroll/QInfiniteScroll.js#L155
`poll.cancel() ` is `undefined` in this case
https://github.com/quasarframework/quasar/blob/712ccc84239e1673e90aa8db3b86ebc4d12c8160/ui/src/components/infinite-scroll/QInfiniteScroll.js#L123

### DIFF
--- a/ui/src/components/infinite-scroll/QInfiniteScroll.js
+++ b/ui/src/components/infinite-scroll/QInfiniteScroll.js
@@ -120,7 +120,7 @@ export default createComponent({
         isWorking.value = false
         isFetching.value = false
         localScrollTarget.removeEventListener('scroll', poll, passive)
-        if(poll !== void 0 && poll.cancel) {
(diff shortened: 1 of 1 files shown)


rstoenescu:
Good catch. Thanks for contributing!
This will go into Quasar v2.6.0

PR #12715, opened 2022-03-09
fix/perf(is): improve perf of isPlainObject and avoid problems with Symbol.toStringTag

pdanpdan:
- see https://github.com/quasarframework/quasar/commit/f4c84adca0138eb6dba64c9a7977c4363b07f6f1#commitcomment-68102067

### DIFF
--- a/ui/src/utils/private/is.js
+++ b/ui/src/utils/private/is.js
@@ -120,8 +120,9 @@ export function isDeepEqual (a, b) {
   return a !== a && b !== b // eslint-disable-line no-self-compare
 }
 
+// see https://jsbench.me/tbl0iliyax/1
 export function isPlainObject (v) {
-  return Object.prototype.toString.call(v) === '[object Object]'
+  return v !== null && typeof v === 'object' && (v.constructor === Object || v.constructor === void 0)
 }
 
 export function isDate (v) {


fjwong:
Hello. First of all, thank you very much for the work in Quasar. We use it quite a lot in our projects and it works great.

We have a problem similar to the one described in https://github.com/quasarframework/quasar/commit/f4c84adca0138eb6dba64c9a7977c4363b07f6f1#commitcomment-68102067 with some of our QSelects after upgrading to Quasar v1.18.5.

In our case, this is a selector for switching between mics and cameras during videoconferencing. The `options` we pass to QSelect is an array of [MediaDeviceInfo](https://developer.mozilla.org/en-US/docs/Web/API/MediaDeviceInfo) objects returned by [navigator.mediaDevices.enumerateDevices()](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/enumerateDevices), and we declare this more or less like:

```
<q-select emit-value v-model="selectedDeviceId" :options="availableDevices" :option-value="deviceId"></q-select>
```

The fix implemented here will not work for our case since `MediaDeviceInfo` objects are not plain javascript objects as checked by `isPlainObject` (constructor is not the same), and I assume other people using Quasar to build videoconferencing apps, or that may be using arrays of other non-plain JS objects, will face the same problem.

We can fix this rather easily on our side by either:
* Remapping the `availableDevices` array to an array of plain objects that have exactly the same properties.
* Using a function for the option values like `:option-value="(device) => device.deviceId"`.

And well, we will have to double-check this for all QSelects in our app, and all other components that are now using `isPlainObject` after v1.18.5.

We don't mind doing this, but before we do this though, we just would like to confirm the following:
* Is disallowing the use of non-plain JS objects in QSelect intended behavior? Is this one of those cases that needed to be guarded against when `isPlainObject` was introduced?.
* If it is, would it be possible to make reference in the QSelect docs that `options` needs to be an array of plain JS objects for `option-value` to work?

puppy0cam:
This PR may resolve the issue with `[Symbol.toStringTag]`, but it introduces a new problem in that I can no longer use classes for such data, so it doesn't really solve my problem for the same reason that it won't for @fjwong.
I can think of two solutions for my issue here, though neither of them would likely work for @fjwong. The first solution is to create a new typescript decorator that I need to include in all of my data classes, which would change the `constructor` property of the prototype to `Object` (or alternatively, delete the `constructor` property from the prototype of my class). I'm not completely certain what side effects this may have, but it would not require further changes to quasar's code. The second option would be to use `class extends Object` alongside a change in how `isPlainObject` works.
```js
const PlainObjectSymbol = Symbol();
Object[PlainObjectSymbol] = void 0;

function isPlainObject3 (v) {
  return v !== null && typeof v === 'object' && (v.constructor === void 0 || PlainObjectSymbol in v.constructor);
}
```
This exhibits similar performance to the previous toString based implementation, which you can check from https://jsbench.me/0nl0lu5tp4/1.
image

rstoenescu:
We'll put all concerns to rest. Went with a different approach that fixes the initial issue and also will not mess up with your use-cases. Will be available in Quasar v2.6.0 and v1.18.6

fjwong:
> We'll put all concerns to rest. Went with a different approach that fixes the initial issue and also will not mess up with your use-cases. Will be available in Quasar v2.6.0 and v1.18.6

@rstoenescu Thank you for the quick response.
We'll be looking forward to v1.18.6.

PR #12742, opened 2022-03-10
feat/refactor: Move Pinia to 'src/stores' folder, install Pinia/Vuex if they are missing

yusufkandemir:

- Feature
- Refactor

- No

**Other information:**
Matches better with [Pinia's recommendation](https://pinia.vuejs.org/cookbook/migration-vuex.html#restructuring-modules-to-stores). Since there is not a single store, 'stores' is a better name. @TobyMosque pointed this out via PM on Discord.

Implemented the TODO comment about installing the store package when it's missing.

Also, I abstracted the package manager logic.


rstoenescu:
Excellent work.

This will however introduce some friction for our developers. Especially the ones with less experience. "Oh, I missed the ending 's' character on the folder name. That's why it wasn't working. Thanks!" :)
But it is what it is. We'll go this way.

PR #12789, opened 2022-03-14
fix(app-vite): Update supported Node versions(14.19+) and make lots of fixes

yusufkandemir:

- Bugfix
- Code style update
- Refactor

- No, because the code was already broken on older Node versions

**Other information:**

We are using private class methods which is part of ES2022:
[node.green/#ES2022-features-private-class-methods](https://node.green/#ES2022-features-private-class-methods)
So, we should be using 14.19
+ as the safest minimum option.

Up until now, we experienced a lot of bug reports, discussions threads, Discord messages, etc. which were related to using unstable(odd-numbered), or too new and untested versions of NodeJS. This made us explain the situation various times and recommend using even-numbered NodeJS versions. So, we are now enforcing stable(even-numbered) versions of NodeJS at `engines` level, which will be validated by the package manager.

I set up ESLint to help with noticing unsupported features and such. I tried enabling more best-practice rule sets and noticed that there were various bugs, so I adjusted the rules and fixed those bugs. It should help us avoid doing those mistakes again.


rstoenescu:
Awesome!

PR #12855, opened 2022-03-19
fix(cache): avoid stack overflow because of vue setting nested fns

pdanpdan:


rstoenescu (changes requested):
Questions:
1. Why hasn't this been detected yet?
2. This essentially removes all of the research done for `on: {}` direct caching. So I'm really interested in a reproduction of the problem to understand what changed.
3. Should this PR go through, it'd be better to use `on: { ...cache() }` than directly changing the cache. Directives do not need the overhead.

pdanpdan:
you are right, it's about component reuse and events merging in that case

PR #12869, opened 2022-03-20
Fix wrong "Please add back <div id="q-app"></div>" error.

rafael-telles:
If a project codebase is using single-quotes instead of double-quotes in HTML, user gets a `Please add back <div id="q-app"></div> to /src/index.template.html inside of <body>` error. This PR prevents this error from happening and avoid failing `quasar build`.


rstoenescu:
Thanks for contributing!
Will be available in q/app-webpack v3.4.5

PR #12906, opened 2022-03-23
chore/fix(create-quasar): Setup ESLint and fix lots of problems

yusufkandemir:

- Bugfix
- Linting

- No

**Other information:**

Now, `_` doesn't get turned into `_.`, it will be stripped completely. So, I updated the existing filenames to include the dot in their names, e.g. `_.eslintrc.js`. The main idea behind this change is to allow "disabling" some files like package.json, tsconfig.json, etc. to keep our tools(e.g. ESLint) and IDEs happy.

I configured ESLint to lint both the regular code and the templated code inside `./templates`. https://github.com/ota-meshi/eslint-plugin-lodash-template is a great plugin that can handle Lodash template syntax(_and more_), so we can detect problems in templated code too. After correctly setting everything up, it detected lots of problems and I found some small problems myself(_e.g. old Mustache syntax but valid JS `name: '{{ name }}'`_), so I fixed all of them.

`yarn lint` is all that's needed to lint the files. ~~Some warnings(3 atm) can be ignored, see the [explanation here](https://github.com/ota-meshi/eslint-plugin-lodash-template). I didn't disable the rule, because it still helps to catch more mistakes. If the related feature gets implemented, then we will not have to worry about it.~~ It got implemented and released really fast. (https://github.com/ota-meshi/eslint-plugin-lodash-template/issues/226#issuecomment-1077311302)


rstoenescu:
The lodash eslint plugin is really a nice one. Good job!

PR #12962, opened 2022-03-28
feat(ui/QNotify): Create types for Notify options (fix: #12961)

### LINKED ISSUE #12961: Expose a QNotifyOptions interface for the Notify plugin similar to QDialogOptions
### Discussed in https://github.com/quasarframework/quasar/discussions/12877

<sup>Originally posted by **Basaingeal** March 21, 2022</sup>
The Dialog plugin has a nice `QDialogOptions` interface that I can use if I want to abstract dialog creation into different Vue hooks. 

The options for `notify(opts)` are just defined as part of the notification function type, which makes it difficult to type any kind of options object that will eventually be passed into `notify` but not immediately. 

It would be nice to have a `QNotifyOptions` interface to use, just how there is already one for the Dialog plugin.</div>

yusufkandemir:

- Feature
- Documentation

- No

**Other information:**

Closes #12961.

New types:
```ts
// See the diff of `dist/types/index.d.ts` for more info
import { QNotifyCreateOptions } from "quasar";

export type QNotifyUpdateOptions = Omit<QNotifyCreateOptions, 'group' | 'position'>;
export type QNotifyOptions = Omit<QNotifyCreateOptions, 'ignoreDefaults'>;
```

 📜 Changes for dist/types/index.d.ts
```diff
@@ -443,508 +443,21 @@
    * @param opts Notification options
    * @returns Calling this function with no parameters hides the notification; When called with one Object parameter (the original notification must NOT be grouped), it updates the notification (specified properties are shallow merged with previous ones; note that group and position cannot be changed while updating and so they are ignored)
    */
   create: (
-    opts:
-      | {
-          /**
-           * Optional type (that has been previously registered) or one of the out of the box ones ('positive', 'negative', 'warning', 'info', 'ongoing')
-           */
-          type?: string;
-          /**
-           * Color name for component from the Quasar Color Palette
-           */
-          color?: string;
-          /**
-           * Color name for component from the Quasar Color Palette
-           */
-          textColor?: string;
-          /**
-           * The content of your message
-           */
-          message?: string;
-          /**
-           * The content of your optional caption
-           */
-          caption?: string;
-          /**
-           * Render message as HTML; This can lead to XSS attacks, so make sure that you sanitize the message first
-           */
-          html?: boolean;
-          /**
-           * Icon name following Quasar convention; Make sure you have the icon library installed unless you are using 'img:' prefix; If 'none' (String) is used as value then no icon is rendered (but screen real estate will still be used for it)
-           */
-          icon?: string;
-          /**
-           * Color name for component from the Quasar Color Palette
-           */
-          iconColor?: string;
-          /**
-           * Size in CSS units, including unit name
-           */
-          iconSize?: string;
-          /**
-           * URL to an avatar/image; Suggestion: public folder
-           */
-          avatar?: string;
-          /**
-           * Useful for notifications that are updated; Displays a Quasar spinner instead of an avatar or icon; If value is Boolean 'true' then the default QSpinner is shown
-           */
-          spinner?: boolean | Component;
-          /**
-           * Color name for component from the Quasar Color Palette
-           */
-          spinnerColor?: string;
-          /**
-           * Size in CSS units, including unit name
-           */
-          spinnerSize?: string;
-          /**
-           * Window side/corner to stick to
-           * Default value: bottom
-           */
-          position?:
-            | "top-left"
-            | "top-right"
-            | "bottom-left"
-            | "bottom-right"
-            | "top"
-            | "bottom"
-            | "left"
-            | "right"
-            | "center";
-          /**
-           * Override the auto generated group with custom one; Grouped notifications cannot be updated; String or number value inform this is part of a specific group, regardless of its options; When a new notification is triggered with same group name, it replaces the old one and shows a badge with how many times the notification was triggered
-           * Default value: (message + caption + multiline + actions labels + position)
-           */
-          group?: boolean | string | number;
-          /**
-           * Color name for the badge from the Quasar Color Palette
-           */
-          badgeColor?: string;
-          /**
-           * Color name for the badge text from the Quasar Color Palette
-           */
-          badgeTextColor?: string;
-          /**
-           * Notification corner to stick badge to; If notification is on the left side then default is top-right otherwise it is top-left
-           * Default value: (top-left/top-right)
-           */
-          badgePosition?:
-            | "top-left"
-            | "top-right"
-            | "bottom-left"
-            | "bottom-right";
-          /**
-           * Style definitions to be attributed to the badge
-           */
-          badgeStyle?: VueStyleProp;
-          /**
-           * Class definitions to be attributed to the badge
-           */
-          badgeClass?: VueClassProp;
-          /**
-           * Show progress bar to detail when notification will disappear automatically (unless timeout is 0)
-           */
-          progress?: boolean;
-          /**
-           * Class definitions to be attributed to the progress bar
-           */
-          progressClass?: VueClassProp;
-          /**
-           * Add CSS class(es) to the notification for easier customization
-           */
-          classes?: string;
-          /**
-           * Key-value for attributes to be set on the notification
-           */
-          attrs?: any;
-          /**
-           * Amount of time to display (in milliseconds)
-           * Default value: 5000
-           */
-          timeout?: number;
-          /**
-           * Notification actions (buttons); If a 'handler' is specified or not, clicking/tapping on the button will also close the notification; Also check 'closeBtn' convenience prop
-           */
-          actions?: any[];
-          /**
-           * Function to call when notification gets dismissed
-           */
-          onDismiss?: () => void;
-          /**
-           * Convenience way to add a dismiss button with a specific label, without using the 'actions' prop; If set to true, it uses a label accordding to the current Quasar language
-           */
-          closeBtn?: boolean | string;
-          /**
-           * Put notification into multi-line mode; If this prop isn't used and more than one 'action' is specified then notification goes into multi-line mode by default
-           */
-          multiLine?: boolean;
-          /**
-           * Ignore the default configuration (set by setDefaults()) for this instance only
-           */
-          ignoreDefaults?: boolean;
-        }
-      | string
-  ) => (props?: {
-    /**
-     * Optional type (that has been previously registered) or one of the out of the box ones ('positive', 'negative', 'warning', 'info', 'ongoing')
-     */
-    type?: string;
-    /**
-     * Color name for component from the Quasar Color Palette
-     */
-    color?: string;
-    /**
-     * Color name for component from the Quasar Color Palette
-     */
-    textColor?: string;
-    /**
-     * The content of your message
-     */
-    message?: string;
-    /**
-     * The content of your optional caption
-     */
-    caption?: string;
-    /**
-     * Render message as HTML; This can lead to XSS attacks, so make sure that you sanitize the message first
-     */
-    html?: boolean;
-    /**
-     * Icon name following Quasar convention; Make sure you have the icon library installed unless you are using 'img:' prefix; If 'none' (String) is used as value then no icon is rendered (but screen real estate will still be used for it)
-     */
-    icon?: string;
-    /**
-     * Color name for component from the Quasar Color Palette
-     */
-    iconColor?: string;
-    /**
-     * Size in CSS units, including unit name
-     */
-    iconSize?: string;
-    /**
-     * URL to an avatar/image; Suggestion: use public folder
-     */
-    avatar?: string;
-    /**
-     * Useful for notifications that are updated; Displays a Quasar spinner instead of an avatar or icon; If value is Boolean 'true' then the default QSpinner is shown
-     */
-    spinner?: boolean | Component;
-    /**
-     * Color name for component from the Quasar Color Palette
-     */
-    spinnerColor?: string;
-    /**
-     * Size in CSS units, including unit name
-     */
-    spinnerSize?: string;
-    /**
-     * Show progress bar to detail when notification will disappear automatically (unless timeout is 0)
-     */
-    progress?: boolean;
-    /**
-     * Class definitions to be attributed to the progress bar
-     */
-    progressClass?: VueClassProp;
-    /**
-     * Add CSS class(es) to the notification for easier customization
-     */
-    classes?: string;
-    /**
-     * Key-value for attributes to be set on the notification
-     */
-    attrs?: any;
-    /**
-     * Amount of time to display (in milliseconds)
-     * Default value: 5000
-     */
-    timeout?: number;
-    /**
-     * Notification actions (buttons); If a 'handler' is specified or not, clicking/tapping on the button will also close the notification; Also check 'closeBtn' convenience prop
-     */
-    actions?: any[];
-    /**
-     * Function to call when notification gets dismissed
-     */
-    onDismiss?: () => void;
-    /**
-     * Convenience way to add a dismiss button with a specific label, without using the 'actions' prop; If set to true, it uses a label accordding to the current Quasar language
-     */
-    closeBtn?: boolean | string;
-    /**
-     * Put notification into multi-line mode; If this prop isn't used and more than one 'action' is specified then notification goes into multi-line mode by default
-     */
-    multiLine?: boolean;
-    /**
-     * Ignore the default configuration (set by setDefaults()) for this instance only
-     */
-    ignoreDefaults?: boolean;
-  }) => void;
+    opts: QNotifyCreateOptions | string
+  ) => (props?: QNotifyUpdateOptions) => void;
   /**
    * Merge options into the default ones
-   * @param opts Notification options
+   * @param opts Notification options (See 'opts' param of 'create()' for object properties)
    */
-  setDefaults: (opts: {
-    /**
-     * Optional type (that has been previously registered) or one of the out of the box ones ('positive', 'negative', 'warning', 'info', 'ongoing')
-     */
-    type?: string;
-    /**
-     * Color name for component from the Quasar Color Palette
-     */
-    color?: string;
-    /**
-     * Color name for component from the Quasar Color Palette
-     */
-    textColor?: string;
-    /**
-     * The content of your message
-     */
-    message?: string;
-    /**
-     * The content of your optional caption
-     */
-    caption?: string;
-    /**
-     * Render message as HTML; This can lead to XSS attacks, so make sure that you sanitize the message first
-     */
-    html?: boolean;
-    /**
-     * Icon name following Quasar convention; Make sure you have the icon library installed unless you are using 'img:' prefix; If 'none' (String) is used as value then no icon is rendered (but screen real estate will still be used for it)
-     */
-    icon?: string;
-    /**
-     * Color name for component from the Quasar Color Palette
-     */
-    iconColor?: string;
-    /**
-     * Size in CSS units, including unit name
-     */
-    iconSize?: string;
-    /**
-     * URL to an avatar/image; Suggestion: use public folder
-     */
-    avatar?: string;
-    /**
-     * Useful for notifications that are updated; Displays a Quasar spinner instead of an avatar or icon; If value is Boolean 'true' then the default QSpinner is shown
-     */
-    spinner?: boolean | Component;
-    /**
-     * Color name for component from the Quasar Color Palette
-     */
-    spinnerColor?: string;
-    /**
-     * Size in CSS units, including unit name
-     */
-    spinnerSize?: string;
-    /**
-     * Window side/corner to stick to
-     * Default value: bottom
-     */
-    position?:
-      | "top-left"
-      | "top-right"
-      | "bottom-left"
-      | "bottom-right"
-      | "top"
-      | "bottom"
-      | "left"
-      | "right"
-      | "center";
-    /**
-     * Color name for the badge from the Quasar Color Palette
-     */
-    badgeColor?: string;
-    /**
-     * Color name for the badge text from the Quasar Color Palette
-     */
-    badgeTextColor?: string;
-    /**
-     * Notification corner to stick badge to; If notification is on the left side then default is top-right otherwise it is top-left
-     * Default value: (top-left/top-right)
-     */
-    badgePosition?: "top-left" | "top-right" | "bottom-left" | "bottom-right";
-    /**
-     * Style definitions to be attributed to the badge
-     */
-    badgeStyle?: VueStyleProp;
-    /**
-     * Class definitions to be attributed to the badge
-     */
-    badgeClass?: VueClassProp;
-    /**
-     * Show progress bar to detail when notification will disappear automatically (unless timeout is 0)
-     */
-    progress?: boolean;
-    /**
-     * Class definitions to be attributed to the progress bar
-     */
-    progressClass?: VueClassProp;
-    /**
-     * Add CSS class(es) to the notification for easier customization
-     */
-    classes?: string;
-    /**
-     * Key-value for attributes to be set on the notification
-     */
-    attrs?: any;
-    /**
-     * Amount of time to display (in milliseconds)
-     * Default value: 5000
-     */
-    timeout?: number;
-    /**
-     * Notification actions (buttons); If a 'handler' is specified or not, clicking/tapping on the button will also close the notification; Also check 'closeBtn' convenience prop
-     */
-    actions?: any[];
-    /**
-     * Function to call when notification gets dismissed
-     */
-    onDismiss?: () => void;
-    /**
-     * Convenience way to add a dismiss button with a specific label, without using the 'actions' prop; If set to true, it uses a label accordding to the current Quasar language
-     */
-    closeBtn?: boolean | string;
-    /**
-     * Put notification into multi-line mode; If this prop isn't used and more than one 'action' is specified then notification goes into multi-line mode by default
-     */
-    multiLine?: boolean;
-  }) => void;
+  setDefaults: (opts: QNotifyOptions) => void;
   /**
    * Register a new type of notification (or override an existing one)
    * @param typeName Name of the type (to be used as 'type' prop later on)
-   * @param typeOpts Notification options
+   * @param typeOpts Notification options (See 'opts' param of 'create()' for object properties)
    */
-  registerType: (
-    typeName: string,
-    typeOpts: {
-      /**
-       * Optional type (that has been previously registered) or one of the out of the box ones ('positive', 'negative', 'warning', 'info', 'ongoing')
-       */
-      type?: string;
-      /**
-       * Color name for component from the Quasar Color Palette
-       */
-      color?: string;
-      /**
-       * Color name for component from the Quasar Color Palette
-       */
-      textColor?: string;
-      /**
-       * The content of your message
-       */
-      message?: string;
-      /**
-       * The content of your optional caption
-       */
-      caption?: string;
-      /**
-       * Render message as HTML; This can lead to XSS attacks, so make sure that you sanitize the message first
-       */
-      html?: boolean;
-      /**
-       * Icon name following Quasar convention; Make sure you have the icon library installed unless you are using 'img:' prefix; If 'none' (String) is used as value then no icon is rendered (but screen real estate will still be used for it)
-       */
-      icon?: string;
-      /**
-       * Color name for component from the Quasar Color Palette
-       */
-      iconColor?: string;
-      /**
-       * Size in CSS units, including unit name
-       */
-      iconSize?: string;
-      /**
-       * URL to an avatar/image; Suggestion: use public folder
-       */
-      avatar?: string;
-      /**
-       * Useful for notifications that are updated; Displays a Quasar spinner instead of an avatar or icon; If value is Boolean 'true' then the default QSpinner is shown
-       */
-      spinner?: boolean | Component;
-      /**
-       * Color name for component from the Quasar Color Palette
-       */
-      spinnerColor?: string;
-      /**
-       * Size in CSS units, including unit name
-       */
-      spinnerSize?: string;
-      /**
-       * Window side/corner to stick to
-       * Default value: bottom
-       */
-      position?:
-        | "top-left"
-        | "top-right"
-        | "bottom-left"
-        | "bottom-right"
-        | "top"
-        | "bottom"
-        | "left"
-        | "right"
-        | "center";
-      /**
-       * Color name for the badge from the Quasar Color Palette
-       */
-      badgeColor?: string;
-      /**
-       * Color name for the badge text from the Quasar Color Palette
-       */
-      badgeTextColor?: string;
-      /**
-       * Notification corner to stick badge to; If notification is on the left side then default is top-right otherwise it is top-left
-       * Default value: (top-left/top-right)
-       */
-      badgePosition?: "top-left" | "top-right" | "bottom-left" | "bottom-right";
-      /**
-       * Style definitions to be attributed to the badge
-       */
-      badgeStyle?: VueStyleProp;
-      /**
-       * Class definitions to be attributed to the badge
-       */
-      badgeClass?: VueClassProp;
-      /**
-       * Show progress bar to detail when notification will disappear automatically (unless timeout is 0)
-       */
-      progress?: boolean;
-      /**
-       * Class definitions to be attributed to the progress bar
-       */
-      progressClass?: VueClassProp;
-      /**
-       * Add CSS class(es) to the notification for easier customization
-       */
-      classes?: string;
-      /**
-       * Key-value for attributes to be set on the notification
-       */
-      attrs?: any;
-      /**
-       * Amount of time to display (in milliseconds)
-       * Default value: 5000
-       */
-      timeout?: number;
-      /**
-       * Notification actions (buttons); If a 'handler' is specified or not, clicking/tapping on the button will also close the notification; Also check 'closeBtn' convenience prop
-       */
-      actions?: any[];
-      /**
-       * Function to call when notification gets dismissed
-       */
-      onDismiss?: () => void;
-      /**
-       * Convenience way to add a dismiss button with a specific label, without using the 'actions' prop; If set to true, it uses a label accordding to the current Quasar language
-       */
-      closeBtn?: boolean | string;
-      /**
-       * Put notification into multi-line mode; If this prop isn't used and more than one 'action' is specified then notification goes into multi-line mode by default
-       */
-      multiLine?: boolean;
-    }
-  ) => void;
+  registerType: (typeName: string, typeOpts: QNotifyOptions) => void;
 }
 
 export interface Platform {
   /**
@@ -13747,8 +13260,146 @@
 import { WebStorageGetItemMethodType } from "./api";
 import { WebStorageGetIndexMethodType } from "./api";
 import { WebStorageGetKeyMethodType } from "./api";
 import { WebStorageGetAllKeysMethodType } from "./api";
+export interface QNotifyCreateOptions {
+  /**
+   * Optional type (that has been previously registered) or one of the out of the box ones ('positive', 'negative', 'warning', 'info', 'ongoing')
+   */
+  type?: string;
+  /**
+   * Color name for component from the Quasar Color Palette
+   */
+  color?: string;
+  /**
+   * Color name for component from the Quasar Color Palette
+   */
+  textColor?: string;
+  /**
+   * The content of your message
+   */
+  message?: string;
+  /**
+   * The content of your optional caption
+   */
+  caption?: string;
+  /**
+   * Render the message as HTML; This can lead to XSS attacks, so make sure that you sanitize the message first
+   */
+  html?: boolean;
+  /**
+   * Icon name following Quasar convention; Make sure you have the icon library installed unless you are using 'img:' prefix; If 'none' (String) is used as value then no icon is rendered (but screen real estate will still be used for it)
+   */
+  icon?: string;
+  /**
+   * Color name for component from the Quasar Color Palette
+   */
+  iconColor?: string;
+  /**
+   * Size in CSS units, including unit name
+   */
+  iconSize?: string;
+  /**
+   * URL to an avatar/image; Suggestion: use public folder
+   */
+  avatar?: string;
+  /**
+   * Useful for notifications that are updated; Displays a Quasar spinner instead of an avatar or icon; If value is Boolean 'true' then the default QSpinner is shown
+   */
+  spinner?: boolean | Component;
+  /**
+   * Color name for component from the Quasar Color Palette
+   */
+  spinnerColor?: string;
+  /**
+   * Size in CSS units, including unit name
+   */
+  spinnerSize?: string;
+  /**
+   * Window side/corner to stick to
+   * Default value: bottom
+   */
+  position?:
+    | "top-left"
+    | "top-right"
+    | "bottom-left"
+    | "bottom-right"
+    | "top"
+    | "bottom"
+    | "left"
+    | "right"
+    | "center";
+  /**
+   * Override the auto generated group with custom one; Grouped notifications cannot be updated; String or number value inform this is part of a specific group, regardless of its options; When a new notification is triggered with same group name, it replaces the old one and shows a badge with how many times the notification was triggered
+   * Default value: (message + caption + multiline + actions labels + position)
+   */
+  group?: boolean | string | number;
+  /**
+   * Color name for the badge from the Quasar Color Palette
+   */
+  badgeColor?: string;
+  /**
+   * Color name for the badge text from the Quasar Color Palette
+   */
+  badgeTextColor?: string;
+  /**
+   * Notification corner to stick badge to; If notification is on the left side then default is top-right otherwise it is top-left
+   * Default value: (top-left/top-right)
+   */
+  badgePosition?: "top-left" | "top-right" | "bottom-left" | "bottom-right";
+  /**
+   * Style definitions to be attributed to the badge
+   */
+  badgeStyle?: VueStyleProp;
+  /**
+   * Class definitions to be attributed to the badge
+   */
+  badgeClass?: VueClassProp;
+  /**
+   * Show progress bar to detail when notification will disappear automatically (unless timeout is 0)
+   */
+  progress?: boolean;
+  /**
+   * Class definitions to be attributed to the progress bar
+   */
+  progressClass?: VueClassProp;
+  /**
+   * Add CSS class(es) to the notification for easier customization
+   */
+  classes?: string;
+  /**
+   * Key-value for attributes to be set on the notification
+   */
+  attrs?: any;
+  /**
+   * Amount of time to display (in milliseconds)
+   * Default value: 5000
+   */
+  timeout?: number;
+  /**
+   * Notification actions (buttons); If a 'handler' is specified or not, clicking/tapping on the button will also close the notification; Also check 'closeBtn' convenience prop
+   */
+  actions?: any[];
+  /**
+   * Function to call when notification gets dismissed
+   */
+  onDismiss?: () => void;
+  /**
+   * Convenient way to add a dismiss button with a specific label, without using the 'actions' prop; If set to true, it uses a label according to the current Quasar language
+   */
+  closeBtn?: boolean | string;
+  /**
+   * Put notification into multi-line mode; If this prop isn't used and more than one 'action' is specified then notification goes into multi-line mode by default
+   */
+  multiLine?: boolean;
+  /**
+   * Ignore the default configuration (set by setDefaults()) for this instance only
+   */
+  ignoreDefaults?: boolean;
+}
+
+import { QNotifyUpdateOptions } from "./api";
+import { QNotifyOptions } from "./api";
 import { VueStyleObjectProp } from "./api";
 import { ValidationRule } from "./api";
 import { QRejectedEntry } from "./api";
 import { SliderMarkerLabels } from "./api";
@@ -13850,244 +13501,10 @@
      * @param opts Notification options
      * @returns Calling this function with no parameters hides the notification; When called with one Object parameter (the original notification must NOT be grouped), it updates the notification (specified properties are shallow merged with previous ones; note that group and position cannot be changed while updating and so they are ignored)
      */
     notify: (
-      opts:
-        | {
-            /**
-             * Optional type (that has been previously registered) or one of the out of the box ones ('positive', 'negative', 'warning', 'info', 'ongoing')
-             */
-            type?: string;
-            /**
-             * Color name for component from the Quasar Color Palette
-             */
-            color?: string;
-            /**
-             * Color name for component from the Quasar Color Palette
-             */
-            textColor?: string;
-            /**
-             * The content of your message
-             */
-            message?: string;
-            /**
-             * The content of your optional caption
-             */
-            caption?: string;
-            /**
-             * Render message as HTML; This can lead to XSS attacks, so make sure that you sanitize the message first
-             */
-            html?: boolean;
-            /**
-             * Icon name following Quasar convention; Make sure you have the icon library installed unless you are using 'img:' prefix; If 'none' (String) is used as value then no icon is rendered (but screen real estate will still be used for it)
-             */
-            icon?: string;
-            /**
-             * Color name for component from the Quasar Color Palette
-             */
-            iconColor?: string;
-            /**
-             * Size in CSS units, including unit name
-             */
-            iconSize?: string;
-            /**
-             * URL to an avatar/image; Suggestion: public folder
-             */
-            avatar?: string;
-            /**
-             * Useful for notifications that are updated; Displays a Quasar spinner instead of an avatar or icon; If value is Boolean 'true' then the default QSpinner is shown
-             */
-            spinner?: boolean | Component;
-            /**
-             * Color name for component from the Quasar Color Palette
-             */
-            spinnerColor?: string;
-            /**
-             * Size in CSS units, including unit name
-             */
-            spinnerSize?: string;
-            /**
-             * Window side/corner to stick to
-             * Default value: bottom
-             */
-            position?:
-              | "top-left"
-              | "top-right"
-              | "bottom-left"
-              | "bottom-right"
-              | "top"
-              | "bottom"
-              | "left"
-              | "right"
-              | "center";
-            /**
-             * Override the auto generated group with custom one; Grouped notifications cannot be updated; String or number value inform this is part of a specific group, regardless of its options; When a new notification is triggered with same group name, it replaces the old one and shows a badge with how many times the notification was triggered
-             * Default value: (message + caption + multiline + actions labels + position)
-             */
-            group?: boolean | string | number;
-            /**
-             * Color name for the badge from the Quasar Color Palette
-             */
-            badgeColor?: string;
-            /**
-             * Color name for the badge text from the Quasar Color Palette
-             */
-            badgeTextColor?: string;
-            /**
-             * Notification corner to stick badge to; If notification is on the left side then default is top-right otherwise it is top-left
-             * Default value: (top-left/top-right)
-             */
-            badgePosition?:
-              | "top-left"
-              | "top-right"
-              | "bottom-left"
-              | "bottom-right";
-            /**
-             * Style definitions to be attributed to the badge
-             */
-            badgeStyle?: VueStyleProp;
-            /**
-             * Class definitions to be attributed to the badge
-             */
-            badgeClass?: VueClassProp;
-            /**
-             * Show progress bar to detail when notification will disappear automatically (unless timeout is 0)
-             */
-            progress?: boolean;
-            /**
-             * Class definitions to be attributed to the progress bar
-             */
-            progressClass?: VueClassProp;
-            /**
-             * Add CSS class(es) to the notification for easier customization
-             */
-            classes?: string;
-            /**
-             * Key-value for attributes to be set on the notification
-             */
-            attrs?: any;
-            /**
-             * Amount of time to display (in milliseconds)
-             * Default value: 5000
-             */
-            timeout?: number;
-            /**
-             * Notification actions (buttons); If a 'handler' is specified or not, clicking/tapping on the button will also close the notification; Also check 'closeBtn' convenience prop
-             */
-            actions?: any[];
-            /**
-             * Function to call when notification gets dismissed
-             */
-            onDismiss?: () => void;
-            /**
-             * Convenience way to add a dismiss button with a specific label, without using the 'actions' prop; If set to true, it uses a label accordding to the current Quasar language
-             */
-            closeBtn?: boolean | string;
-            /**
-             * Put notification into multi-line mode; If this prop isn't used and more than one 'action' is specified then notification goes into multi-line mode by default
-             */
-            multiLine?: boolean;
-            /**
-             * Ignore the default configuration (set by setDefaults()) for this instance only
-             */
-            ignoreDefaults?: boolean;
-          }
-        | string
-    ) => (props?: {
-      /**
-       * Optional type (that has been previously registered) or one of the out of the box ones ('positive', 'negative', 'warning', 'info', 'ongoing')
-       */
-      type?: string;
-      /**
-       * Color name for component from the Quasar Color Palette
-       */
-      color?: string;
-      /**
-       * Color name for component from the Quasar Color Palette
-       */
-      textColor?: string;
-      /**
-       * The content of your message
-       */
-      message?: string;
-      /**
-       * The content of your optional caption
-       */
-      caption?: string;
-      /**
-       * Render message as HTML; This can lead to XSS attacks, so make sure that you sanitize the message first
-       */
-      html?: boolean;
-      /**
-       * Icon name following Quasar convention; Make sure you have the icon library installed unless you are using 'img:' prefix; If 'none' (String) is used as value then no icon is rendered (but screen real estate will still be used for it)
-       */
-      icon?: string;
-      /**
-       * Color name for component from the Quasar Color Palette
-       */
-      iconColor?: string;
-      /**
-       * Size in CSS units, including unit name
-       */
-      iconSize?: string;
-      /**
-       * URL to an avatar/image; Suggestion: use public folder
-       */
-      avatar?: string;
-      /**
-       * Useful for notifications that are updated; Displays a Quasar spinner instead of an avatar or icon; If value is Boolean 'true' then the default QSpinner is shown
-       */
-      spinner?: boolean | Component;
-      /**
-       * Color name for component from the Quasar Color Palette
-       */
-      spinnerColor?: string;
-      /**
-       * Size in CSS units, including unit name
-       */
-      spinnerSize?: string;
-      /**
-       * Show progress bar to detail when notification will disappear automatically (unless timeout is 0)
-       */
-      progress?: boolean;
-      /**
-       * Class definitions to be attributed to the progress bar
-       */
-      progressClass?: VueClassProp;
-      /**
-       * Add CSS class(es) to the notification for easier customization
-       */
-      classes?: string;
-      /**
-       * Key-value for attributes to be set on the notification
-       */
-      attrs?: any;
-      /**
-       * Amount of time to display (in milliseconds)
-       * Default value: 5000
-       */
-      timeout?: number;
-      /**
-       * Notification actions (buttons); If a 'handler' is specified or not, clicking/tapping on the button will also close the notification; Also check 'closeBtn' convenience prop
-       */
-      actions?: any[];
-      /**
-       * Function to call when notification gets dismissed
-       */
-      onDismiss?: () => void;
-      /**
-       * Convenience way to add a dismiss button with a specific label, without using the 'actions' prop; If set to true, it uses a label accordding to the current Quasar language
-       */
-      closeBtn?: boolean | string;
-      /**
-       * Put notification into multi-line mode; If this prop isn't used and more than one 'action' is specified then notification goes into multi-line mode by default
-       */
-      multiLine?: boolean;
-      /**
-       * Ignore the default configuration (set by setDefaults()) for this instance only
-       */
-      ignoreDefaults?: boolean;
-    }) => void;
+      opts: QNotifyCreateOptions | string
+    ) => (props?: QNotifyUpdateOptions) => void;
     platform: Platform;
     screen: Screen;
     sessionStorage: SessionStorage;
   }
```

### DIFF
--- a/ui/src/plugins/Notify.json
+++ b/ui/src/plugins/Notify.json
@@ -376,9 +376,29 @@
             "actions": {
               "type": "Array",
               "tsType": "QNotifyActions",
-              "desc": "Notification actions (buttons); If a 'handler' is specified or not, clicking/tapping on the button will also close the notification; Also check 'closeBtn' convenience prop",
+              "desc": "Notification actions (buttons); Unless 'noDismiss' is true, clicking/tapping on the button will close the notification; Also check 'closeBtn' convenience prop",
+              "definition": {
+                "handler": {
+                  "type": "Function",
+                  "desc": "Function to be executed when the button is clicked/tapped",
+                  "params": null,
+                  "returns": null,
+                  "examples": [
+                    "() => console.log('button clicked')"
+                  ]
+                },
+                "noDismiss": {
+                  "type": "Boolean",
+                  "desc": "Do not dismiss the notification when the button is clicked/tapped"
+                },
+                "...": {
+                  "type": "Any",
+                  "desc": "Any other QBtn prop expect 'onClick'(use 'handler' instead)",
+                  "examples": [ "label: 'Learn more'", "color: 'primary'" ]
+                }
+              },
               "examples": [
-                "[ { label: 'Show', handler: (Function), attrs: { 'aria-label': 'Button label' } }, { icon: 'map', handler: (Function), color: 'yellow' }, { label: 'Learn more', noDismiss: true, handler: (Function) } ]"
+                "[ { label: 'Show', handler: () => {}, 'aria-label': 'Button label' }, { icon: 'map', handler: () => {}, color: 'yellow' }, { label: 'Learn more', noDismiss: true, handler: () => {} } ]"
               ]
             },
 
--- a/ui/src/plugins/Notify.json
+++ b/ui/src/plugins/Notify.json
@@ -375,7 +375,7 @@
                 },
                 "...": {
                   "type": "Any",
-                  "desc": "Any other QBtn prop expect 'onClick'(use 'handler' instead)",
+                  "desc": "Any other QBtn prop expect 'onClick' (use 'handler' instead)",
                   "examples": [ "label: 'Learn more'", "color: 'primary'" ]
                 }
               },
--- a/ui/src/plugins/Notify.json
+++ b/ui/src/plugins/Notify.json
@@ -64,7 +64,7 @@
       },
 
       "spinner": {
-        "type": [ "Boolean", "Component" ],
+        "type": "Boolean",
         "desc": "Useful for notifications that are updated; Displays a Quasar spinner instead of an avatar or icon; If value is Boolean 'true' then the default QSpinner is shown",
         "examples": [ true, "QSpinnerBars" ]
       },
@@ -167,24 +167,6 @@
         "desc": "Amount of time to display (in milliseconds)",
         "default": 5000,
         "examples": [ 2500 ]
-      },
-
-      "actions": {
-        "type": "Array",
-        "desc": "Notification actions (buttons); If a 'handler' is specified or not, clicking/tapping on the button will also close the notification; Also check 'closeBtn' convenience prop",
-        "examples": [
-          "[ { label: 'Show', handler: (Function), attrs: { 'aria-label': 'Button label' } }, { icon: 'map', handler: (Function), color: 'yellow' }, { label: 'Learn more', noDismiss: true, handler: (Function) } ]"
-        ]
-      },
-
-      "onDismiss": {
-        "type": "Function",
-        "desc": "Function to call when notification gets dismissed",
-        "params": null,
-        "returns": null,
-        "examples": [
-          "() => { console.log('Dismissed') }"
-        ]
       },
 
       "closeBtn": {
--- a/ui/src/plugins/Notify.json
+++ b/ui/src/plugins/Notify.json
@@ -36,7 +36,7 @@
 
       "html": {
         "type": "Boolean",
-        "desc": "Render the message as HTML; This can lead to XSS attacks so make sure that you sanitize the message first"
+        "desc": "Render the message as HTML; This can lead to XSS attacks, so make sure that you sanitize the message first"
       },
 
       "icon": {
@@ -64,8 +64,9 @@
       },
 
       "spinner": {
-        "type": "Boolean",
-        "desc": "Useful for notifications that are updated; Displays the default Quasar spinner instead of an avatar or icon"
+        "type": [ "Boolean", "Component" ],
+        "desc": "Useful for notifications that are updated; Displays a Quasar spinner instead of an avatar or icon; If value is Boolean 'true' then the default QSpinner is shown",
+        "examples": [ true, "QSpinnerBars" ]
       },
 
       "spinnerColor": {
@@ -86,8 +87,7 @@
           "top-left", "top-right",
           "bottom-left", "bottom-right",
           "top", "bottom", "left", "right", "center"
-        ],
-        "examples": [ "top-right" ]
+        ]
       },
 
       "group": {
@@ -112,21 +112,24 @@
         "values": [
           "top-left", "top-right",
           "bottom-left", "bottom-right"
-        ],
-        "examples": [ "bottom-right" ]
+        ]
       },
       "badgeStyle": {
-        "type": "String",
+        "type": [ "String", "Array", "Object" ],
+        "tsType": "VueStyleProp",
         "desc": "Style definitions to be attributed to the badge",
         "examples": [
-          "background-color: #ff0000"
+          "background-color: #ff0000",
+          "{ backgroundColor: '#ff0000' }"
         ]
       },
       "badgeClass": {
-        "type": "String",
+        "type": [ "String", "Array", "Object" ],
+        "tsType": "VueClassProp",
         "desc": "Class definitions to be attributed to the badge",
         "examples": [
-          "my-special-class"
+          "my-special-class",
+          "{ 'my-special-class': <condition> }"
         ]
       },
 
@@ -135,10 +138,12 @@
         "desc": "Show progress bar to detail when notification will disappear automatically (unless timeout is 0)"
       },
       "progressClass": {
-        "type": "String",
+        "type": [ "String", "Array", "Object" ],
+        "tsType": "VueClassProp",
         "desc": "Class definitions to be attributed to the progress bar",
         "examples": [
-          "my-special-class"
+          "my-special-class",
+          "{ 'my-special-class': <condition> }"
         ]
       },
 
@@ -164,9 +169,27 @@
         "examples": [ 2500 ]
       },
 
+      "actions": {
+        "type": "Array",
+        "desc": "Notification actions (buttons); If a 'handler' is specified or not, clicking/tapping on the button will also close the notification; Also check 'closeBtn' convenience prop",
+        "examples": [
+          "[ { label: 'Show', handler: (Function), attrs: { 'aria-label': 'Button label' } }, { icon: 'map', handler: (Function), color: 'yellow' }, { label: 'Learn more', noDismiss: true, handler: (Function) } ]"
+        ]
+      },
+
+      "onDismiss": {
+        "type": "Function",
+        "desc": "Function to call when notification gets dismissed",
+        "params": null,
+        "returns": null,
+        "examples": [
+          "() => { console.log('Dismissed') }"
+        ]
+      },
+
       "closeBtn": {
         "type": [ "Boolean", "String" ],
-        "desc": "Convenient way to add a dismiss button with a specific label, without using the 'actions' convoluted prop",
+        "desc": "Convenient way to add a dismiss button with a specific label, without using the 'actions' prop; If set to true, it uses a label according to the current Quasar language",
         "examples": [ "Close me" ]
       },
 
@@ -184,6 +207,8 @@
       "params": {
         "opts": {
           "type": [ "Object", "String" ],
+          "tsType": "QNotifyCreateOptions",
+          "autoDefineTsType": true,
           "required": true,
           "desc": "Notification options",
           "definition": {
@@ -215,7 +240,7 @@
 
             "html": {
               "type": "Boolean",
-              "desc": "Render message as HTML; This can lead to XSS attacks, so make sure that you sanitize the message first"
+              "desc": "Render the message as HTML; This can lead to XSS attacks, so make sure that you sanitize the message first"
             },
 
             "icon": {
@@ -234,7 +259,7 @@
 
             "avatar": {
               "type": "String",
-              "desc": "URL to an avatar/image; Suggestion: public folder",
+              "desc": "URL to an avatar/image; Suggestion: use public folder",
               "examples": [
                 "(public folder) img/something.png",
                 "(relative path format) require('./my_img.jpg')",
@@ -368,7 +393,7 @@
 
             "closeBtn": {
               "type": [ "Boolean", "String" ],
-              "desc": "Convenience way to add a dismiss button with a specific label, without using the 'actions' prop; If set to true, it uses a label accordding to the current Quasar language",
+              "desc": "Convenient way to add a dismiss button with a specific label, without using the 'actions' prop; If set to true, it uses a label according to the current Quasar language",
               "examples": [ "Close me" ]
             },
 
@@ -390,150 +415,10 @@
         "params": {
           "props": {
             "type": "Object",
+            "tsType": "QNotifyUpdateOptions",
             "required": false,
-            "desc": "Notification properties that will be shallow merged to previous ones in order to update the non-grouped notification",
-            "definition": {
-              "type": {
-                "type": "String",
-                "desc": "Optional type (that has been previously registered) or one of the out of the box ones ('positive', 'negative', 'warning', 'info', 'ongoing')",
-                "examples": [ "negative", "custom-type" ]
-              },
-
-              "color": {
-                "extends": "color"
-              },
-
-              "textColor": {
-                "extends": "color"
-              },
-
-              "message": {
-                "type": "String",
-                "desc": "The content of your message",
-                "examples": [ "John Doe pinged you" ]
-              },
-
-              "caption": {
-                "type": "String",
-                "desc": "The content of your optional caption",
-                "examples": [ "5 minutes ago" ]
-              },
-
-              "html": {
-                "type": "Boolean",
-                "desc": "Render message as HTML; This can lead to XSS attacks, so make sure that you sanitize the message first"
-              },
-
-              "icon": {
-                "extends": "icon"
-              },
-
-              "iconColor": {
-                "extends": "color",
-                "addedIn": "v2.5.5"
-              },
-
-              "iconSize": {
-                "extends": "size",
-                "addedIn": "v2.5.5"
-              },
-
-              "avatar": {
-                "type": "String",
-                "desc": "URL to an avatar/image; Suggestion: use public folder",
-                "examples": [
-                  "(public folder) img/something.png",
-                  "(relative path format) require('./my_img.jpg')",
-                  "(URL) https://some-site.net/some-img.gif"
-                ]
-              },
-
-              "spinner": {
-                "type": [ "Boolean", "Component" ],
-                "desc": "Useful for notifications that are updated; Displays a Quasar spinner instead of an avatar or icon; If value is Boolean 'true' then the default QSpinner is shown",
-                "__exemption": [ "examples" ]
-              },
-
-              "spinnerColor": {
-                "extends": "color",
-                "addedIn": "v2.5.5"
-              },
-
-              "spinnerSize": {
-                "extends": "size",
-                "addedIn": "v2.5.5"
-              },
-
-              "progress": {
-                "type": "Boolean",
-                "desc": "Show progress bar to detail when notification will disappear automatically (unless timeout is 0)"
-              },
-              "progressClass": {
-                "type": [ "String", "Array", "Object" ],
-                "tsType": "VueClassProp",
-                "desc": "Class definitions to be attributed to the progress bar",
-                "examples": [
-                  "my-special-class",
-                  "{ 'my-special-class': <condition> }"
-                ]
-              },
-
-              "classes": {
-                "type": "String",
-                "desc": "Add CSS class(es) to the notification for easier customization",
-                "examples": [
-                  "my-notif-class"
-                ]
-              },
-
-              "attrs": {
-                "type": "Object",
-                "desc": "Key-value for attributes to be set on the notification",
-                "examples": [ "{ role: 'alertdialog' }" ],
-                "__exemption": [ "definition" ]
-              },
-
-              "timeout": {
-                "type": "Number",
-                "desc": "Amount of time to display (in milliseconds)",
-                "default": 5000,
-                "examples": [ 2500 ]
-              },
-
-              "actions": {
-                "type": "Array",
-                "desc": "Notification actions (buttons); If a 'handler' is specified or not, clicking/tapping on the button will also close the notification; Also check 'closeBtn' convenience prop",
-                "examples": [
-                  "[ { label: 'Show', handler: (Function), attrs: { 'aria-label': 'Button label' } }, { icon: 'map', handler: (Function), color: 'yellow' }, { label: 'Learn more', noDismiss: true, handler: (Function) } ]"
-                ]
-              },
-
-              "onDismiss": {
-                "type": "Function",
-                "desc": "Function to call when notification gets dismissed",
-                "params": null,
-                "returns": null,
-                "examples": [
-                  "() => { console.log('Dismissed') }"
-                ]
-              },
-
-              "closeBtn": {
-                "type": [ "Boolean", "String" ],
-                "desc": "Convenience way to add a dismiss button with a specific label, without using the 'actions' prop; If set to true, it uses a label accordding to the current Quasar language",
-                "examples": [ "Close me" ]
-              },
-
-              "multiLine": {
-                "type": "Boolean",
-                "desc": "Put notification into multi-line mode; If this prop isn't used and more than one 'action' is specified then notification goes into multi-line mode by default"
-              },
-
-              "ignoreDefaults": {
-                "type": "Boolean",
-                "desc": "Ignore the default configuration (set by setDefaults()) for this instance only"
-              }
-            }
+            "desc": "Notification properties that will be shallow merged to previous ones in order to update the non-grouped notification; (See 'opts' param of 'create()' for object properties, except 'group' and 'position')",
+            "__exemption": [ "definition", "examples" ]
           }
         },
         "returns": null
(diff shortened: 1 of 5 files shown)


yusufkandemir:
@Basaingeal, as you are the one that [requested this feature](https://github.com/quasarframework/quasar/discussions/12877#discussion-3953270), please let us know what you think of the solution. Would this work for you?

Basaingeal:
Yes. This looks like this would work.

rstoenescu (changes requested):
Looks good, but there is an issue with `quasarConfOptions`. It's no mistake that it differs from all the other definitions. For example, you can't configure something with a Function, because this definition would be part of quasar.config.js which is not UI code -- then the App CLI generator comes and creates .quasar folder and things start to fall apart.
So please leave quasarConfOptions with the original definitions and touch the other parts only.

ui/src/plugins/Notify.json
```diff
@@ -390,150 +418,10 @@
         "params": {
           "props": {
             "type": "Object",
+            "tsType": "QNotifyUpdateOptions",
             "required": false,
-            "desc": "Notification properties that will be shallow merged to previous ones in order to update the non-grouped notification",
```
rstoenescu, line 394: Sorry for not pointing this earlier, but we shouldn't remove these as well, even though there is a tsType defined now.

JSONs are not only for helping TS defs. They are consumed as standalone too. Example 1: docs API card; Example 2: quasar describe command.

rstoenescu (changes requested):

ui/src/plugins/Notify.json
yusufkandemir: That's why I added "reference"/"pointer" messages like the following:
> See 'opts' param of 'create()' for object properties, except 'group' and 'position'

We are already using this approach in a lot of places. One example is that in `actions`, we are pointing out `QBtn` props, because 1. it's easy to get lost as there are lots of nested properties, 2. it's really hard to keep JSON API definitions in sync because we can't re-use whole structures, unlike TS. Checking out the API card for `Notify.create` is really hard, because there are lots of properties in parameters, and there are lots more in the return type as well, most are the same, hard to spot the difference by eye. Definitions, descriptions, etc. were already out-of-sync in this file. So, a simple explanation and a pointer are more helpful and simple IMO, both for users and maintainers. But, if you still think we should add it, I'll add them back.

rstoenescu: Missed the "see opts param". It's not a perfect solution but it's pretty good because besides the maintenance benefit it also decreases the size of the JSON file -- and I'm constantly researching ways to decrease the footprint of the docs (as some users that don't use the PWA are complaining about high load times).

As for the docs API cards, I'm planning some tweaking to improve the UX.

Maybe some thoughts for force keeping defs in sync: we do an AST analysis to catch possible problems, but it only goes so far. Once a composable is used the analysis does not cover it (same for any other imported file). For someone who has a bit of time to enhance the AST analysis this would be something really great for the framework. Then we should do something about the TS defs which are manually defined so as to not get out of sync with UI code and JSON.

yusufkandemir: I don't have a lot of ideas about optimizing the JSON files. I will keep this problem in mind to see if I can think of anything.

UX improvements sound great.

For keeping things in sync: I was thinking of a TS-first approach(_and experimenting with it_). As TypeScript + JSDoc/TSDoc is more detailed and comprehensive than JSON API, we can auto-generate JSON API from TypeScript. We have some extra metadata such as `category`, `addedIn`, etc. in JSON API, some of them are available in JSDoc/TSDoc specs, some of them can be easily added as custom tags. We still need to preserve&support JSON API because we are using it in our docs, the describe command, in AEs, etc. Furthermore, since we will use JSON API as the common API, we can progressively migrate and enhance types. So, we don't have to convert all the types at once. I am also thinking of writing a script to convert an existing JSON API file to a separate `.d.ts` in the related component/composable/etc. folder, to use it as a base. It's easy because currently, we can convert JSON -> TS too.

By adopting this, we can provide generic types(_e.g. `QSelect<TModel, TOption>`_), have more detailed types, have an easier time when creating/syncing types, thus reducing the maintenance costs, etc.

A short example of what I have(_and aiming to have_) after a limited time of working on this idea:
```ts
// /ui/src/components/select/QSelect.d.ts

/**
 * @public
 * @docsUrl https://v2.quasar.dev/vue-components/select
 */
export interface QSelectProps<Model = any, Option = QSelectOption<Model>>
  extends UseFieldProps<Model>,
    UseVirtualScrollProps<QSelect> {
  /**
   * Model of the component.
   * Must be Array if using 'multiple' prop.
   * Either use this property (along with a listener for 'update:modelValue' event) OR use v-model directive
   *
   * @category model
   * @see https://v2.quasar.dev/vue-components/select#the-model
   * @example
   * ```
   * v-model="myModel"
   * ```
   */
  modelValue: Model;

  /**
   * Available options that the user can select from.
   * For best performance freeze the list of options.
   *
   * @category options
   * @defaultValue `[]`
   * @example String options
   * ```
   * :options="[ 'BMW', 'Samsung Phone' ]"
   * ```
   * @example Object options
   * ```
   * :options="[ { label: 'BMW', value: 'car' }, { label: 'Samsung Phone', value: 'phone' } ]"
   * ```
   */
  options?: Option[];

  /**
   * Property of option which holds the 'value'.
   * If using a function then for best performance, reference it from your scope and do not define it inline.
   *
   * @category options
   * @defaultValue 'value'
   * @param option The current option being processed.
   * @returns Value of the current option.
   * @example Property name form
   * ```
   * option-value="modelNumber"
   * ```
   * @example Function form
   * ```
   * :option-value="(item) => item === null ? null : item.modelNumber"
   * ```
   */
  optionValue?: FnOrKey<Option, Model>;

// ...

// /ui/src/composables/private/use-field.d.ts
export interface UseFieldProps<T = any> extends UseValidateProps<T> {
  /**
   * Used to specify the name of the control.
   * Useful if dealing with forms.
   * If not specified, it takes the value of 'for' prop, if it exists.
   */
  name?: string;
// ...
```
We can also reference these types in JS source code in form of JSDocs, and improve the DX in our source code too. Should we plan to migrate our code-base to TS, we can use these types. So, I see a lot of value in this. Please let me know what you think.

Basaingeal:
I think there's a mistake in the `QNotifyCreateOptions` interface. The `actions` property is defined as `QNotifyActions[]`, which is an array of arrays. I think it's should either just be `QNotifyActions` or `QNotifyAction[]`

ui/src/plugins/Notify.json
rstoenescu: Sounds really juicy. Let me think about it a bit and will come back to you shortly.

PR #12972, opened 2022-03-29
fix(app-vite): Fix SSR publicPath handling

yusufkandemir:

- Bugfix

- No

**Other information:**

### DIFF
--- a/app-vite/lib/modes/ssr/ssr-devserver.js
+++ b/app-vite/lib/modes/ssr/ssr-devserver.js
@@ -307,7 +307,7 @@ class SsrDevServer extends AppDevserver {
     const { publicPath } = this.#appOptions
 
     publicPath !== '/' && app.use((req, res, next) => {
-      const pathname = new URL(req.url).pathname || '/'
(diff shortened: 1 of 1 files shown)


app-vite/lib/modes/ssr/ssr-devserver.js
```diff
@@ -296,12 +304,10 @@ class SsrDevServer extends AppDevserver {
       })
     })
 
-    await injectMiddlewares(middlewareParams)
```
rstoenescu, line 299: Please remove the publicPath handling altogether.

If the middlewares run after publicPath then we kill the possibility of SSR to be able to supply other paths (eg. /api). On the other hand, if the middlewares run first, then the publicPath logic won't be able to run since the render middleware catches every route and ends the socket with html.

rstoenescu (changes requested):

app-vite/lib/modes/ssr/ssr-devserver.js
rstoenescu: On second thought, the middleware catches everything for the publicPath in use. So forget about the comment above.

PR #13049, opened 2022-04-03
feat(ui): update XMLHttpRequest to support IE6 and older

Hashen110:
[Using XMLHttpRequest in IE6](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest_in_IE6)

Need review

### DIFF
--- a/ui/dev/src/pages/components/ajax-bar.vue
+++ b/ui/dev/src/pages/components/ajax-bar.vue
@@ -72,7 +72,15 @@
 import { LoadingBar } from 'quasar'
 
 function sendXhr (url) {
-  const xhr = new XMLHttpRequest()
+  let xhr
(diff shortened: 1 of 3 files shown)


pdanpdan:
Quasar v2 and Vue3 do not support IE at all :)

rstoenescu:
The minimum version of IE supported by Quasar v1 was IE11. The earlier versions have 0% marketshare. But that was Quasar v1, based on Vue 2. We're now at Quasar v2, which uses Vue 3. Vue 3 does not have support for IE at all due to technical reasons. Microsoft itself is trying to get rid of IE and has ended support for it, finally.

It makes no sense to support IE6 in 2022. And even if this was the case, this PR won't do it, not even by a long shot.

PR #13075, opened 2022-04-05
fix(q-layout): Update scroll regardless of scroll prevented #12994

### LINKED ISSUE #12994: Fixed header with reveal prop doesn't reshow itself on page navigation while loading plugin is active
### What happened?

When scrolled down on a page with the fixed header hidden due to the reveal prop and clicking on a link that navigates to another page. If the page has a q loading plugin active then the header will not reappear. This is occuring do the fact the q loading plugin sets qScrollPrevented to true. If qScrollPrevented is true when the onPageScroll function is called in QLayout via the onScroll event then the $Layout.scroll is never updated which the QHeader watches.

For the QHeader to not appear disappeared on the page navigated to.

### Reproduction URL

https://github.com/kalvenschraut/vitejs-vite-uswlyh/tree/master

### How to reproduce?

I had a lot of issues reproducing consistently on the stackblitz site but I have narrowed down what exactly the problem by debugging in my applications code. I feel like the scroll object needs to be updated regardless if scroll is prevented for cases like mine.

### Flavour

### Areas

Components (quasar), Plugins (quasar)

### Platforms/Browsers

Chrome

### Quasar info output

```Shell
System:
    OS: Linux 5.10 Gentoo/Linux
    CPU: (4) x64 Intel(R) Core(TM) i5-6500 CPU @ 3.20GHz
    Memory: 6.65 GB / 31.33 GB
    Container: Yes
    Shell: 5.1.8 - /bin/bash
  Binaries:
    Node: 16.10.0 - /usr/bin/node
    npm: 6.14.15 - /usr/bin/npm
    Watchman: 4.9.0 - /usr/bin/watchman
  Browsers:
    Chromium: 95.0.4638.69
    Firefox: 91.3.0esr
  npmPackages:
    @quasar/extras: ^1.13.4 => 1.13.4
    @quasar/quasar-ui-qcalendar: 4.0.0-beta.13 => 4.0.0-beta.13
    @quasar/vite-plugin: ^1.0.7 => 1.0.7
    @vitejs/plugin-vue: ^2.3.1 => 2.3.1
    quasar: 2.6.2 => 2.6.2
```

### Relevant log output

_No response_

### Additional context

_No response_

kalvenschraut:
Fixes #12994

Problem as I described in my issue is due to the fact the loading plugin uses the prevent scroll plugin. Due to this if the loader is ever active when a scroll event would be triggered the scroll position change would be ignored. This issue was specifically a problem for me clicking on a button at the bottom of a long page which then redirected to a page with no scroll. My pages uses the loading plugin and suspense to show a global spinner and the next page is loaded. In this scenario since I use the reveal header prop when I am scrolled all the way down to click on the navigation button, the header is hidden. Then the loader appears and sets prevent scroll true, scroll event fires since page body has changed and it isn't long enough to need a scroll bar. Due to the previous steps the header never reappears since the prevent scroll true is preventing the updating of the scroll object it is watching. 

As I mentioned in my issue I couldn't consistently reproduce this outside of my applications environment. Seems to be a very specific race condition which are always fun to try and reproduce. Hope I described it thoroughly enough so the problem can be easily seen just by code inspection alone. 

Alternative solution to this problem would be to only update scroll if it has a changed value when scrolling is prevented but not sure if that is really any better then always updating it on change event. Open to other possible ways to fix my scenario, figured I would make a PR for mine since my issue wasn't getting any response.

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No? I don't think this is breaking but can't be sure

**Other information:**

### DIFF
--- a/ui/src/components/layout/QLayout.js
+++ b/ui/src/components/layout/QLayout.js
@@ -69,7 +69,7 @@ export default createComponent({
     ))
 
     function onPageScroll (data) {
-      if (props.container === true || document.qScrollPrevented !== true) {
+      if (props.container === true) {
         const info = {
           position: data.position.top,
           direction: data.direction,


rstoenescu:
Hi,

Thank you for contributing!
Unfortunately, this fixes your specific issue but breaks the exact problem that was fixed by adding that bit of code.

Will look at your ticket and try to find a fix for that scenario instead.

PR #13099, opened 2022-04-06
feat(ui): add new type to web storage

Hashen110:
`undefined` is also a type


rstoenescu:
Hi. It's handled implicitly.

PR #13105, opened 2022-04-07
feat(ui): add new props to QImg

Hashen110:
Review before merge this


rstoenescu:
It's just now that it struck me reviewing this PR. I totally forgot that Qv1's QImg works totally different. This shouldn't be ported back because it makes little sense :( Many apologies for requesting it when I shouldn't.

PR #13115, opened 2022-04-07
adding email validation rule

aymenBox:
#### Reference Issues/PRs
None
#### what does this implement fix? Explain your changes.
Adding Email validation Pattern rules in ui/types/api/validation.d.ts and ui/src/utils/patterns.js 

#### any other comments 
if this implementation is helpful i would like to continue working on other validation pattern rules to be used with q-input


pdanpdan:
The correct regex for RFC822 is
http://www.ex-parrot.com/~pdw/Mail-RFC822-Address.html

Something that works most of the time is
https://emailregex.com/

See https://stackoverflow.com/questions/201323/how-can-i-validate-an-email-address-using-a-regular-expression

aymenBox:
@pdanpdan thank you, I will change the regex to RFC 5322 for email validation if that's acceptable.

rstoenescu:
Added some clarifications as comments.
We didn't add an email rule from the beginning because it opens up Pandora's box. However, I think that at least a basic validation will do, once it is clear that we won't be accepting "tweaks" to it since devs can write their own rules anyway.

Thanks for contributing!

aymenBox:
@rstoenescu Thank you and I'm happy to be helpful, and i agree other than ( time , date , email , ... ) validation which are standard any other validation can be written with costume rules by devs, i constantly use email validation that's way i made this pull request. 
however should i make a pull request to the documentation page to add the email string?

rstoenescu:
No need to PR docs, because they are pointing directly to this file. Which is why I also added comments.

Thanks for contributing!

PR #13118, opened 2022-04-07
chore(ui): add prefix for text-overflow

Hashen110:
Default browserslist configuration doesn't prefix `text-overflow`


pdanpdan:
What browser requires prefix for this?

rstoenescu:
Hi,

This is not required. The whole point is for developers to adapt the browserslist to their requirements and the build tools will do their part. But moreover, for Quasar v2 I highly doubt there is still a modern browser that requires this.

Issue #13013, opened 2022-03-31
q/app-vite: Report when lint error/warning has been fixed


rstoenescu:
### Discussed in https://github.com/quasarframework/quasar/discussions/12982

<div type='discussions-op-text'>

<sup>Originally posted by **jdheeter** March 30, 2022</sup>
During development while running the Vite Dev server it's normal for the dev server to pick up some invalid code and throw an error like `vite] Internal server error: Transform failed with 1 error:`

The problem is that after the error is resolved, there is no feedback in the dev server log. When using webpack it's very obvious when the dev server has refreshed and is now serving correctly or still stuck on an error.

While the Vite dev server is working correctly, it's not always obvious from a developers perspective, because the errors are not followed by any success messages in the logs.</div>

rstoenescu:
Will be available in q/app-vite beta.12

maggie44:
It would be nice to have an indicator when the dev environment is up and the 'open browser' setting is set to false. 

When open browser automatically is set to true:

```
 .d88888b.
d88P" "Y88b
888     888
888     888 888  888  8888b.  .d8888b   8888b.  888d888
888     888 888  888     "88b 88K          "88b 888P"
888 Y8b 888 888  888 .d888888 "Y8888b. .d888888 888
Y88b.Y8b88P Y88b 888 888  888      X88 888  888 888
 "Y888888"   "Y88888 "Y888888  88888P' "Y888888 888
       Y8b


 » Reported at............ 16/04/2022 23:22:50
 » App dir................ /Users/arranmagee/Documents/GitHub.nosync/balena-device-ui/ui
 » App URL................ http://localhost:9000/
                           http://192.168.0.36:9000/
 » Dev mode............... spa
 » Pkg quasar............. v2.6.6
 » Pkg @quasar/app-vite... v1.0.0-beta.14
 » Browser target......... es2019|edge88|firefox78|chrome87|safari13.1

 App • Opening default browser at http://localhost:9000/
```

When open browser is set to false:

```
 .d88888b.
d88P" "Y88b
888     888
888     888 888  888  8888b.  .d8888b   8888b.  888d888
888     888 888  888     "88b 88K          "88b 888P"
888 Y8b 888 888  888 .d888888 "Y8888b. .d888888 888
Y88b.Y8b88P Y88b 888 888  888      X88 888  888 888
 "Y888888"   "Y88888 "Y888888  88888P' "Y888888 888
       Y8b


[...]

PR #13198, opened 2022-04-17
fix(app): Indicator that server is running even if browser not opening

MilosPaunovic:
> It would be nice to have an indicator when the dev environment is up and the 'open browser' setting is set to false.

https://github.com/quasarframework/quasar/issues/13013#issuecomment-1100763792


rstoenescu:
Information is already available in the banner at App URL

PR #13322, opened 2022-05-03
fix(QItem): take active prop into account if both route and active given

Tofandel:

- [x] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
This ensures that if the `active` prop is given it will always be used, otherwise if using a route the active prop has no use and the  logic to determine if active has no way to be customized

Use case: to be able to mark the item as active in a related route (not necessarily a children route, so `exact` does not fit this use case)

```vue
<QItem :to="{name: 'some.route.list'}" :active="/^some.route/.test($route.name)"/>
```

Currently the previous code would not result in the item being active on route `some.route.edit` even though it's clearly what's wanted, this PR fixes this

### DIFF
--- a/ui/src/components/item/QItem.js
+++ b/ui/src/components/item/QItem.js
@@ -20,7 +20,10 @@ export default createComponent({
       default: 'div'
     },
 
-    active: Boolean,
+    active: {
+      type: Boolean,
+      default: null
+    },
 
     clickable: Boolean,
     dense: Boolean,
@@ -58,7 +61,7 @@ export default createComponent({
       + (props.dense === true ? ' q-item--dense' : '')
       + (isDark.value === true ? ' q-item--dark' : '')
       + (
-        hasLink.value === true && props.active === undefined
+        hasLink.value === true && props.active === null
           ? linkClass.value
           : (
               props.active === true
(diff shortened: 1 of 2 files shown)


ui/src/components/item/QItem.js
```diff
@@ -58,7 +58,7 @@ export default createComponent({
       + (props.dense === true ? ' q-item--dense' : '')
       + (isDark.value === true ? ' q-item--dark' : '')
       + (
-        hasLink.value === true
+        hasLink.value === true && props.active === undefined
```
pdanpdan: `props.active` is guaranteed to be either `true` or `false` - never `undefined` (and `undefined` does not comply with repo's ESLint rules - use `void 0`)

You must change line 23 from `active: Boolean,` to

```javascript
active: {
  type: Boolean,
  default: null
}
```

Then use `&& props.active === null`

pdanpdan (changes requested):

ui/src/components/item/QItem.js
pdanpdan: Also you have to change the json definition for QItem

Tofandel: Arg yes, indeed I forgot about this specificity of the `Boolean` vue type

pdanpdan (approved):

rstoenescu:
Thanks for contributing!

PR #13371, opened 2022-05-06
sass variables for files with only @use statements

disfated:
scss-transform will fail currently for files like this:
```scss
@use 'A';
@use 'B'; // no new line here
```
gives
```
[vite] Internal server error: @use rules must be written before any other rules.
```

- [x] Bugfix

- [x] No


rstoenescu:
Thanks for contributing!

PR #13383, opened 2022-05-09
Setting causing to long tab when window larger than 1440px. if you have 13 tabs and full-width is 1440 then the lasts tabs are not shown.(13 x 128 = 1664)

sebastienroul:


sebastienroul:
If you have 13 tab and tabs full-width is 1440px then the lasts tabs are not shown.(13 x 128 = 1664).
This setting is maybe old, but nobody can guess the min-width, cause it depends on the number of tab in the tabs

With the setting : min-width
image

Without : 
image

As you can see, tab-content are perfectly feeting the q-tabs width, instead of overfowing.

rstoenescu:
Thanks for contributing!
This will go into 2.7.1

PR #13442, opened 2022-05-16
fix(ui): replace deprecated `keyCode` with `key`

Hashen110:


pdanpdan:
I would not change this, I think keyCode is used uniformly in all codebase

PR #13444, opened 2022-05-16
fix(ui): use optionsDense prop to check virtual scroll item size

kovalchukq:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
While using QSelect with next props:
```
dense=true
options-dense=false
```
And you have more items than QSelect min-height - menu is jumping. 
I think we must use optionsDense prop to check virtual scroll item size.


rstoenescu:
Thank you for contributing!

PR #13466, opened 2022-05-19
fix(app-vite): Add ssr context to SSR mode

MilosPaunovic:
Closes #13059


rstoenescu (changes requested):
Hi,

There is no "ctx" when creating the project folder. At best, this should just include the folder directly, unless something breaks. But it needs testing first. Would you be so kind to make the adjustments and test, pls?

MilosPaunovic:
Included directly, tested by creating new project with all the same options as [here](https://github.com/quasarframework/quasar/discussions/13059#discussion-3986144), works as expected.

rstoenescu:
Good work!

PR #13467, opened 2022-05-19
fix(QRadio): doesn't work normally when value is not proxy object

### LINKED ISSUE #13457: QOptionGroup with objects as values for the options does not reflect the selected option in UI
### What happened?

Using a default (radio-button) q-option-group with objects a as values for the options the modelValue is changed correctly on selection but the UI does not reflect the state.

The radio button of the current selection should be filled (reflect the selection).

### Reproduction URL

https://codepen.io/suchwerk123/pen/WNMjNva?editors=101

### How to reproduce?

1. Make a selection
2. See that the model value changed
3. The radio button of the selected option still is not filled

### Flavour

### Areas

Components (quasar)

### Platforms/Browsers

_No response_

### Quasar info output

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

volvachev:
Fixes #13457

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/ui/src/components/radio/QRadio.js
+++ b/ui/src/components/radio/QRadio.js
@@ -1,4 +1,4 @@
-import { h, ref, computed, getCurrentInstance } from 'vue'
+import { h, ref, computed, getCurrentInstance, toRaw } from 'vue'
 
 import QIcon from '../icon/QIcon.js'
 
@@ -64,7 +64,7 @@ export default createComponent({
     const rootRef = ref(null)
     const { refocusTargetEl, refocusTarget } = useRefocusTarget(props, rootRef)
 
-    const isTrue = computed(() => props.modelValue === props.val)
+    const isTrue = computed(() => toRaw(props.modelValue) === toRaw(props.val))
 
     const classes = computed(() =>
       'q-radio cursor-pointer no-outline row inline no-wrap items-center'
--- a/ui/src/components/checkbox/use-checkbox.js
+++ b/ui/src/components/checkbox/use-checkbox.js
@@ -61,11 +61,12 @@ export default function (type, getInner) {
     props.val !== void 0 && Array.isArray(props.modelValue)
   )
 
-  const index = computed(() => (
-    modelIsArray.value === true
-      ? props.modelValue.findIndex(opt => toRaw(opt) === toRaw(props.val))
+  const index = computed(() => {
+    const val = toRaw(props.val)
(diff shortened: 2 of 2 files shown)


ui/src/components/radio/QRadio.js
```diff
@@ -64,7 +64,11 @@ export default createComponent({
     const rootRef = ref(null)
     const { refocusTargetEl, refocusTarget } = useRefocusTarget(props, rootRef)
 
-    const isTrue = computed(() => props.modelValue === props.val)
+    const isTrue = computed(() => {
```
pdanpdan: You don't need to check for `isProxy`, you can directly check `toRaw(props.modelValue) === toRaw(props.val)`

pdanpdan (changes requested):

ui/src/components/radio/QRadio.js
volvachev: Yeah, sure

pdanpdan (approved):

pdanpdan:
Can you please also check in `use-checkbox.js` - the branch for QCheckbox/QToggle?

It needs changing in the `index`/`isTrue`/`isFalse` computeds

```
  const index = computed(() => (
    modelIsArray.value === true
      ? props.modelValue.findIndex(opt => toRaw(opt) === toRaw(props.val))
      : -1
  ))

  const isTrue = computed(() => {
    return modelIsArray.value === true
      ? index.value > -1
      : toRaw(props.modelValue) === toRaw(props.trueValue)
  })

  const isFalse = computed(() => (
    modelIsArray.value === true
      ? index.value === -1
      : toRaw(props.modelValue) === toRaw(props.falseValue)
  ))
```

volvachev:
> Can you please also check in `use-checkbox.js` - the branch for QCheckbox/QToggle?

Hi, you are right, same problem.

pdanpdan:
Good job from my pov.
Now I think we should wait for Razvan to decide on what we do, because this problem might hit anywhere in the framework

rstoenescu:
This will go into 2.7.1

rstoenescu:
Thanks for contributing!

suchwerk:
Yes, thank you all! For the initial workaround and fix!

PR #13106, opened 2022-04-07
feat(app-vite&app-webpack): Add PNPM support

GaliMu:
**What kind of change does this PR introduce?** 

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** 

- [ ] Yes
- [x] No

**The PR fulfills these requirements:**

- [x] It's submitted to the `dev` branch (or `v[X]` branch)
- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)
- [ ] It's been tested on a Cordova (iOS, Android) app
- [ ] It's been tested on a Electron app
- [ ] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs)  or explained in the PR's description.

If adding a **new feature**, the PR's description includes:
- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to [start a new feature discussion](https://github.com/quasarframework/quasar/discussions/new?category=ideas-proposals) first and wait for approval before working on it)

**Other information:**
PNPM can only be correctly identified and used when executing a command, and it is uncertain whether the command can be executed correctly.

PR #12809, opened 2022-03-15
docs(cli): add pnpm instructions

soulsam480:
**What kind of change does this PR introduce?** 

- [ ] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** 

- [ ] Yes
- [x] No

**The PR fulfills these requirements:**

- [x] It's submitted to the `dev` branch (or `v[X]` branch)
- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)
- [ ] It's been tested on a Cordova (iOS, Android) app
- [ ] It's been tested on a Electron app
- [x] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs)  or explained in the PR's description.

If adding a **new feature**, the PR's description includes:
- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to [start a new feature discussion](https://github.com/quasarframework/quasar/discussions/new?category=ideas-proposals) first and wait for approval before working on it)

**Other information:**
As the new Quasar CLI is working properly with pnpm, it makes sense to mention it in the docs. This PR adds pnpm starter commands where npm and yarn are mentioned.

PR #13615, opened 2022-06-05
feat(create-quasar): PNPM support

yusufkandemir:

- Feature

- No

**Other information:**

`shamefully-hoist=true` is for allowing app code to access dependencies in `@quasar/app-*` packages. One example would be `@quasar/fastclick`, which is used for iOS on PWA and Cordova. We inject this into auto-generated entry files, which are actually part of the app, so it will not be able to access the dependency defined in `@quasar/app-*` packages and will result in errors. We can choose to make it a peer dependency of `@quasar/app-*` packages and make it part of the starter kit. But, since that's not the only thing, we would be making the starter kits confusing and bloated, and they will be harder to update/manage, if it becomes not needed, we can't simply remove it, etc.

For app-webpack there are lots of packages, so I don't see really see hope in a "shamefully-hoist-less" app-webpack. ~~For app-vite however, if I am not missing something, the only dependency with a runtime reference is `@quasar/fastclick`, so we can decide to move it to devland to be able to go "shamefully-hoist-less".~~ There are also `express`, `compress`, etc. and their respective `@types/*` packages too. So, we would need to go on a longer route, which is probably not a good idea, at least for now.

Nuxt is also doing [`shamefully-hoist=true`](https://github.com/nuxt/starter/tree/90f48a598c32f38982801c35b2a61be7ac923776#setup), in fact, their starter kit only has `nuxt` as a dev dependency, the rest including vue and vue-router is shared(_not using exact pins_). It would have at least been a problem in terms of import suggestions(_like we had in #9235_), but that's not a problem because they offer auto-imports.

`strict-peer-dependencies=false` is for avoiding the need for installing internal libraries like webpack, vite, etc. on devland. Installing webpack, vite, etc. loaders/plugins start enforcing this, so it's annoying without any real benefit at the app level for a framework like this IMO.

Needs these for full PNPM support experience:
- #13609
- #13106
- #12809
- More documentation changes
- Messages/logs inside q-app messages/instructions. An instruction like "_Run `npm run x` or `yarn x`_" should also cover PNPM. Ideally, we should only display the instructions for the package manager the user is using.

### DIFF
--- a/app-vite/package.json
+++ b/app-vite/package.json
@@ -113,6 +113,13 @@
     "npm": ">= 6.14.12",
     "yarn": ">= 1.17.3"
   },
+  "pnpm": {
+    "peerDependencyRules": {
+      "ignoreMissing": [
+        "rollup"
+      ]
+    }
+  },
   "volta": {
     "node": "lts"
   },
--- a/app-vite/package.json
+++ b/app-vite/package.json
@@ -72,7 +72,6 @@
     "minimist": "^1.2.6",
     "open": "^8.4.0",
     "ouch": "^2.0.0",
-    "postcss": "^8.4.4",
     "register-service-worker": "^1.7.2",
     "rollup-plugin-visualizer": "^5.5.4",
     "sass": "1.32.12",
--- a/create-quasar/templates/app/quasar-v2/js-webpack/BASE/_.npmrc
+++ b/create-quasar/templates/app/quasar-v2/js-webpack/BASE/_.npmrc
@@ -1,0 +1,2 @@
+shamefully-hoist=true
+strict-peer-dependencies=false
--- a/create-quasar/templates/app/quasar-v2/js-webpack/BASE/_.npmrc
+++ b/create-quasar/templates/app/quasar-v2/js-webpack/BASE/_.npmrc
@@ -1,2 +1,3 @@
+# pnpm-related options
 shamefully-hoist=true
 strict-peer-dependencies=false
--- a/create-quasar/index.js
+++ b/create-quasar/index.js
@@ -88,6 +88,7 @@ async function run () {
             ]
             : [
               { title: 'Yes, use Yarn (recommended)', value: 'yarn' },
+              { title: 'Yes, use PNPM', value: 'pnpm' },
               { title: 'Yes, use NPM', value: 'npm' },
               { title: 'No, I will handle that myself', value: false }
             ]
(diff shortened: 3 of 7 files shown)


app-vite/package.json
```diff
@@ -72,7 +72,6 @@
     "minimist": "^1.2.6",
     "open": "^8.4.0",
     "ouch": "^2.0.0",
-    "postcss": "^8.4.4",
```
IlCallo, line 75: Isn't this removal a breaking change? Can you explain why you needed to move it to the project `package.json`?

create-quasar/templates/app/quasar-v2/js-webpack/BASE/_.npmrc
```diff
@@ -0,0 +1,2 @@
+shamefully-hoist=true
```
IlCallo, line 2: I'd say `.npmrc` file should be scaffolded only when using PNPM
But I guess the problem here is that Quasar asks about the package manager to use only after the scaffolding process took place

What about adding at least a comment on the top to mark this is only useful when using PNPM and can be deleted otherwise?

app-vite/package.json
yusufkandemir: [`4b75327` (#13615)](https://github.com/quasarframework/quasar/pull/13615/commits/4b7532724a52e84a62fe7eb34749b7435350d45b)

> chore(app-vite): Move postcss to devland
We are not directly using it in app-vite, so we can move it to devland.
This helps to avoid missing peer deep errors/warnings from autoprefixer,
which is installed on the default template, and is basically required.
Existing apps will not be affected because Vite also ships postcss as
a dependency.

create-quasar/templates/app/quasar-v2/js-webpack/BASE/_.npmrc
yusufkandemir: I also thought about that. Then, I thought some people may want to migrate to PNPM, or it could be part of a project where multiple package managers are welcome(_open-source apps, others?_). Also, as you said the package manager is asked after scaffolding. Then, I decided to always include it since it's not a big deal.

We can put `utils.runningPackageManager` to `scope`, then use it when it's defined. Then, we can scaffold this file when it's PNPM or undefined. But, I guess adding comments would be a simpler addition, and would still cover all of the things I mentioned. So, I will add a comment.

app-vite/package.json
IlCallo: Sorry, didn't think about checking the commits, that's clear now

IlCallo (approved):

PR #13624, opened 2022-06-06
feat(app-vite&app-webpack): Add TS template for PWA mode (part of #8572)

yusufkandemir:

- Feature

- No

**Other information:**
Part of #8572, supersedes #8705.

`quasar mode add pwa` will now use the TS template if the app has TS.

Side note: When using Volar with Takeover mode, `process` gives errors due to https://github.com/johnsoncodehk/volar/issues/1341. It still builds and works correctly.


IlCallo (commented):
Is it correct that [custom-service-worker.ts](https://github.com/quasarframework/quasar/pull/13624/files#diff-24a213fe8611e11b2c75ef2a90d4c9ecf1f8ae9dd11c38e87b1aa17069a3ccc0) has much less stuff in it for webpack than Vite? Is it due to additional configuration needed for webpack? Or do we have additional optimizations for Vite that we don't have for Webpack?

yusufkandemir:
@IlCallo yes, it's not just about its TS-variant though, the original JS version is also thin. It is most probably due to Vite being newer. @rstoenescu probably took the opportunity to create a more feature-rich template for the Vite one, the same one we use for our docs. We can probably add the same stuff to the Webpack variant after implementing `process.env.PWA_FALLBACK_HTML` for app-webpack. It's practically out of the scope of this PR though. We can deal with it later, after getting some context from @rstoenescu.

IlCallo:
Yes, not into the scope of this PR, just wanted to check if it was something left out by mistake
I'll test this out tomorrow and merge if everything works fine 👍🏻

IlCallo:
Everything is working fine
The only thing which bugs me is that we always scaffold the eslint override file
I never saw a codebase with TS and without linting, but since we allow people to choose when they setup the project, I'd not scaffold it if not required

If there's an easy way to do so, let's fix this. Otherwise people without eslint can just delete the file

yusufkandemir:
@IlCallo first world problems 😂 I also thought about this, if it's been thought about twice and not really hard to implement, it might be a good idea. Done 👍

Issue #12016, opened 2022-01-12
QBtn with QIcon doesn't render properly as of V2.4.8


jaybo:
### What happened?

I know a bug was fixed in V2.4.8 which had caused icons to be oversized (which I tested and verified),  but there is a new issue where it appears that buttons with icons are still not displaying their full content.

![SvgBug](https://user-images.githubusercontent.com/1607323/149209784-e2628069-4d65-4d51-8ff4-6d818180ada6.png)
None of my code in this area has changed in years, but this new effect just appeared.  

### What did you expect to happen?

Make buttons with icons appear like they always have.

### Reproduction URL

https://codepen.io/jaybonomad/pen/WNZYYdG

### How to reproduce?

1. The .png in the codepen is different from the above image since I can't upload a png to codepen without a pro account.  I've substituted the Twitter logo with the same dimensions and it reproduces the same problem.

### Flavour

Quasar CLI (@quasar/cli | @quasar/app)

### Areas

Components (quasar), SPA Mode, PWA Mode

### Quasar info output

```shell
(base) d:\devGit\dzweb\ClientApp>quasar info

Operating System - Windows_NT(10.0.22000) - win32/x64
NodeJs - 16.1.0

Global packages
  NPM - 7.24.1
  yarn - 1.22.5
  @quasar/cli - 1.2.2
  @quasar/icongenie - 2.3.3
  cordova - Not installed

Important local packages
  quasar - 2.4.8 -- Build high-performance VueJS user interfaces (SPA, PWA, SSR, Mobile and Desktop) in record time
  @quasar/app - 3.2.6 -- Quasar Framework local CLI
  @quasar/extras - 1.12.3 -- Quasar Framework fonts, icons and animations
  eslint-plugin-quasar - Not installed
  vue - 3.2.24 -- The progressive JavaScript framework for buiding modern web UI.
  vue-router - 4.0.12
  vuex - Not installed
  electron - Not installed
  electron-packager - Not installed
  electron-builder - Not installed
  @babel/core - 7.15.4 -- Babel compiler core.
  webpack - 5.58.1 -- Packs CommonJs/AMD modules for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand. Support loaders to preprocess files, i.e. json, jsx, es7, css, less, ... and your custom stuff.
  webpack-dev-server - 4.7.1 -- Serves a webpack app. Updates the browser on changes.
[...]

PR #13651, opened 2022-06-09
fix(QIcon): fix QIcon shrink without breaking aspect #12016, #13521

pdanpdan:


Tofandel:
This breaks the icons of `.q-timeline__dot` which have a special styling with `width: 100%`
Before
image
After
image


Only the case with svg icons

PR #13652, opened 2022-06-09
feat(app-vite&app-webpack): Add TS template for SSR mode (part of #8572)

yusufkandemir:

- Feature

- No

**Other information:**
Part of #8572, supersedes #8646.


app-webpack/templates/ssr/ts/middlewares/render.ts
```diff
@@ -0,0 +1,54 @@
...
+  // over to Vue and Vue Router to render our page
+  app.get(resolve.urlPath('*'), (req, res) => {
+    res.setHeader('Content-Type', 'text/html');
+
+    render(/* the ssrContext: */ { req, res })
+      .then(html => {
+        // now let's send the rendered html to the client
+        res.send(html);
+      })
+      .catch(err => {
```
IlCallo: We should add the error type here as `(err: : RenderError)`

The interface exists both on Vite and Webpack app, so you can import it as
`import { RenderError } from '@quasar/app-vite';`
`import { RenderError } from '@quasar/app-webpack';`

IlCallo (changes requested):

IlCallo:
I tested all combinations and everything seems to work fine, you can merge it after fixing the render error type thing, good job 💪🏻

app-webpack/templates/ssr/ts/middlewares/render.ts
yusufkandemir: Good catch! Will do.

ESLint and strict TS would normally go crazy about this, but when working on the Electron TS template, I discovered that since these files are not included in the root tsconfig, they are not really "acting as usual". I think we should either directly include them into the root tsconfig's `include` section or create tsconfigs for each folder, then connect them to the root using `references`. Otherwise, we'll not get the full benefits of TS detection and strict linting.
Here is an example I found while working on Electron TS template:
```ts
// The root level of electron-main.ts
let mainWindow: BrowserWindow | undefined;
// mainWindow: BrowserWindow
mainWindow.loadURL() // will end up with "can't access `loadURL` of undefined" at runtime, but TS doesn't care
```
So, IMO that is unacceptable and we should address that. If you have any preferred solution among those two or have another solution, please let me know.

IlCallo: This is due to the usage of `include` option of tsconfig into Quasar Vite CLI
Quasar Webpack CLI only uses `exclude` instead, which avoid the problem
We should also be able to correctly extend a tsconfig preset provided by app-vite package now, instead of copying the whole configuration to devland, which would allow us to fix those kind of problems on our side without asking for manual changes to the users

PR #13664, opened 2022-06-09
feat(extras): Define exports field in package.json

kalvenschraut:
Typescript recently released 4.7 which allows compiling to esm output for node. With this I have to set moduleResolution to use nodenext. With the new esm module resolution this requires extensions to be used on everything, a way to get around this is to setup exports object mapping the import paths to the files. You can also setup conditional mapping so that require will pull in the cjs version, import the esm version and where the types are located. I went through and added the export mapping for all the different icon imports.

I had to do this due to an error I was receiving about not able to import my desired icons while trying to run tests with vitest  due to it missing the extension (the import from your code due to me marking quasar/extras as an external package at build time). I added all of them for the different icon imports to map to each index file of it. Note that this forces all possible imports to be defined in the exports field, so for backwards compatibility `"./*": "./*"` was added, but ideally quasar will explicitly state what can all be imported in the exports field and not need the wildcard allowing everything. Guessing this change would be looked at when quasar 3.0 is being looked at.

https://nodejs.org/api/packages.html#package-entry-points the spec defined by nodejs.

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/extras/package.json
+++ b/extras/package.json
@@ -9,104 +9,104 @@
   "exports": {
     ".": "./index.js",
     "./bootstrap-icons": {
+      "types": "./bootstrap-icons/index.d.ts",
       "import": "./bootstrap-icons/index.mjs",
-      "require": "./bootstrap-icons/index.js",
-      "types": "./bootstrap-icons/index.d.ts"
+      "require": "./bootstrap-icons/index.js"
     },
     "./eva-icons": {
+      "types": "./eva-icons/index.d.ts",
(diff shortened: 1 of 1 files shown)


radoslavirha:
Hi, any update on this? Thanks

PR #13665, opened 2022-06-09
feat(ui):  export field in package.json and .mjs extension for esm

kalvenschraut:
Similar to #13664  define exports field, but the main thing blocking me causing me to submit this PR is that your esm bundled output has the extensions of .js when it needs .mjs so node can recognize it as an esm module. Otherwise my tests with vitest refused to import it.

In the exports field on package.json I am rewritting the export of the old file with the .js file to the .mjs file in case anyone was directly importing it so it is backwards compatible. I then define the conditional import maps for when just importing from `quasar` and finally added `"./*": "./*"` to ensure all direct imports still work to not cause any issues with backwards compatibility. As noted in my other PR, ideally only the things that can be imported would be defined instead, Likely in quasar 3 is planned to be made.

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


kalvenschraut:
Just noticed that the build script is changing all .mjs extensions to .js. I added a check specifically for the quasar.esm file, otherwise   if for some reason quasar doesn't want to change its quasar.esm.prod.js file to quasar.esm.prod.mjs can there be another file generated called quasar.prod.mjs? Mjs should imply that it is esm already.

kalvenschraut:
@rstoenescu is this something that can be looked at? A recent update, not sure what is causing my tests to fail because vitest is trying to use the CJS SSR bundle. Adding the export mapping like I have done so here fixes the problem for me.

kalvenschraut:
@rstoenescu I need some kind of resolution on this as I currently am forced to maintain my fork with these changes otherwise my unit tests are failing.

Due to my pushing for supporting OSS, my organization sponsors you and quasar with that one of your benefits was priority fixes/features.

rstoenescu:
@kalvenschraut
Hi. Sorry for the delay. Lots going on under the covers.
Can we connect on [Discord](https://chat.quasar.dev) pls? Would be easier to reach a resolution.

Issue #13700, opened 2022-06-15
QSelect scroll position goes to top when pushing more options


gcarpi:
### What happened?

I'm using the example of [Dynamic Loading Options](https://quasar.dev/vue-components/select#example--dynamic-loading-options) and when I populate the select with more options the scroll is set to the top.

I created an [issue](https://github.com/quasarframework/quasar/issues/13694) reporting a similar problem. In that issue the scroll is sent to the first selected option and I managed to solve this problem using some virtual scroll methods, as you can see below:

```js
async $_onVirtualScroll ({ index, ref }) {
  const lastIndex = this.filteredOptions.length - 1

  if (index === lastIndex && this.$_canFetchOptions()) {
    await this.$_loadMoreOptions()

    this.$nextTick(() => { // This code fixes the issue #13694
      ref.reset()
      ref.refresh(lastIndex)
    })
  }
 }
```

But this code generated a new problem, sometimes when I push new options for the select, the scroll is sent to the top and I can't scroll down anymore because I'm always thrown up.

It's hard to reproduce because it only happens a few times, so I'll leave a video below as a demonstration:

https://user-images.githubusercontent.com/19765074/173937317-9c638f66-c25a-459c-af94-a12e12b7b134.mov

### What did you expect to happen?

I would like to have some explanation of why this is happening and if I can prevent this problem from happening.

### Reproduction URL

https://codepen.io/guilhermecarpi/pen/gOvqwPr?editors=101

### How to reproduce?

I can't reproduce this issue outside of my project

### Flavour

Quasar CLI (@quasar/cli | @quasar/app)

### Areas

Components (quasar)

### Platforms/Browsers

Chrome

### Additional context

Quasar version: 1.19.1

DouglasCalora:
Please, someone help!! 😢

rstoenescu:
Fix will be available in Quasar v2.7.2 and 1.19.2

Issue #13694, opened 2022-06-14
QSelect start on selected option (issue with Dynamic Loading Options)


gcarpi:
### What happened?

I'm using the example of [Dynamic Loading Options](https://quasar.dev/vue-components/select#example--dynamic-loading-options) and when I populate the select with more options the scroll is reset to the first selected.

This happens because there is a [feature that performs this action](https://github.com/quasarframework/quasar/pull/4427), but I believe that usability is affected when we want to pull more options for the select (lazy loading of options).

![Simulation](https://user-images.githubusercontent.com/19765074/173612332-3ba0a9d7-90c6-44b6-b111-c1604df92593.gif)

**OBS:** This behavior happens in v2 too 

### What did you expect to happen?

My suggestion is to create a property to disable this action, preventing the current behavior of QSelect from being changed.

### Reproduction URL
https://codepen.io/guilhermecarpi/pen/gOvqwPr?editors=101

### How to reproduce?

1. Go to the reproduction URL;
2. Choose for example option 30;
3. Perform the scroll event to search for more options;
4. When the options are incremented, notice that the scroll is returned to the first selected one.

### Flavour

Quasar CLI (@quasar/cli | @quasar/app)

### Areas

Components (quasar)

### Platforms/Browsers

Chrome

DouglasCalora:
Any idea if this will be fixed?

rstoenescu:
Fix will be available in Quasar v2.7.2 and 1.19.2

PR #13711, opened 2022-06-16
fix(QSelect): fix menu scroll when options list length increases #13700, #13694

pdanpdan:


gcarpi:
This works! I tested!

I look forward to seeing this fix in a new version!

Thank you very much!

PR #13717, opened 2022-06-17
Remove !important - it works without

bartduisters:

- [ ] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
It works without !important, no need to introduce it.

### DIFF
--- a/docs/src/pages/style/color-palette.md
+++ b/docs/src/pages/style/color-palette.md
@@ -63,10 +63,10 @@ If you want to use your own colors for your components (let's say we are adding 
 
 ```css
 .text-brand {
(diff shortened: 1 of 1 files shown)


rstoenescu:
Hi,

Thanks for contributing.

It works only in *some* scenarios, but not all. What we want is to be thorough and make it work in all possible scenarios.
We're also trying to write Quasar's CSS in BEM form and there are plenty of cases where a color defined without the "!important" part will not have any effect. The order of definitions also matter.

PR #13758, opened 2022-06-22
feat(app-vite&app-webpack): Add TS template for Electron mode (part of #8572)

yusufkandemir:

- Feature

- No

**Other information:**
Part of #8572, supersedes #8661, #8647, and #8575.


IlCallo:
This, as SSR one, will have the small problem of giving errors when being used into Vite-based projects with the faulty "include" property of tsconfig we recently removed from the create template, but I don't think there's an efficient way to fix this problem and it's rare for some projects to add a mode after a while they setup the project, so the impact should be low

PR #13809, opened 2022-06-28
feat(docs): Copying prop names to clipboard

MilosPaunovic:
Props in the API cards of the documentation are now clickable, and click would copy prop name to clipboard and show Notify message about it, same as the one when anchor header link is copied.

### DIFF
--- a/docs/src/assets/page-utils.js
+++ b/docs/src/assets/page-utils.js
@@ -43,6 +43,17 @@ export function copyHeading (id) {
   })
 }
 
+export function copyPropName (propName) {
+  copyToClipboard(propName)
+
+  Notify.create({
+    message: 'Prop name has been copied to clipboard.',
+    position: 'top',
+    actions: [{ icon: 'cancel', color: 'white', dense: true, round: true }],
+    timeout: 2000
+  })
+}
+
 // eslint-disable-next-line
 const specialRE = /[\s·/_\\,:;\.\(\)\[\]]+/g
 const andRE = /&/g
--- a/docs/src/assets/page-utils.js
+++ b/docs/src/assets/page-utils.js
@@ -43,17 +43,6 @@ export function copyHeading (id) {
   })
 }
 
-export function copyPropName (propName) {
-  copyToClipboard(propName)
-
-  Notify.create({
-    message: 'Prop name has been copied to clipboard.',
-    position: 'top',
-    actions: [{ icon: 'cancel', color: 'white', dense: true, round: true }],
-    timeout: 2000
-  })
-}
-
 // eslint-disable-next-line
 const specialRE = /[\s·/_\\,:;\.\(\)\[\]]+/g
 const andRE = /&/g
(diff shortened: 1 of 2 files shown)


docs/src/components/DocApiEntry.js
```diff
@@ -78,10 +79,15 @@ function getExtendedNameDiv (label, level, type, required, addedIn) {
   const suffix = `${type ? ` : ${type}` : ''}${required ? ' - required!' : ''}`
 
   const child = [
-    h(QBadge, {
-      color: NAME_PROP_COLOR[ level ],
+    h(QChip, {
```
IlCallo: Seems like you're trying to force a QChip to behave as a QBadge here
Is it because QBadge won't allow you to react to click events?

MilosPaunovic: Yup, that was the only reason.

yusufkandemir: `QBadge` click event seems to work correctly: https://codepen.io/yusufkandemir/pen/BarzojJ?editors=101
I wonder what's wrong.

Some random thought: `QChip` sounds somewhat more semantical than `QBadge` in this context.

docs/src/assets/page-utils.js
```diff
@@ -43,6 +43,17 @@ export function copyHeading (id) {
...
+export function copyPropName (propName) {
+  copyToClipboard(propName)
+
+  Notify.create({
+    message: 'Prop name has been copied to clipboard.',
+    position: 'top',
+    actions: [{ icon: 'cancel', color: 'white', dense: true, round: true }],
+    timeout: 2000
+  })
+}
```
yusufkandemir: `copyHeading` which is kind of similar to this helper, is used in 3 different places. But, this one is only being used in one place. So, I would put this into where it's used.

yusufkandemir:
I wonder if we should create an anchor system for this too. I think it's common to point people out to some API entry. But, we can only send screenshots, or send the link to API and make some explanation, etc. So, I think a URL that will open the related tab and category, then scroll into the related entry would be nice. Not the most important thing in the world, also probably a bit hard to implement, but could be a nice addition.

docs/src/components/DocApiEntry.js
IlCallo: Yeah, that was my same doubt, AFAIK you can just listen to click even on a `QBadge`, to keep the code consistent and avoid forcing a `QChip` to look like a `QBadge`

IlCallo:
That seems an interesting idea, but I'd keep the scope of this PR narrow and add that feature in a follow up PR

docs/src/components/DocApiEntry.js
MilosPaunovic: I'm gonna take another look at this and report back.

MilosPaunovic: I've made some mistakes first time, as `QBadge` is now behaving perfectly on listening to `click` event.

docs/src/assets/page-utils.js
MilosPaunovic: Agreed, moved logic to the only place where it's used.

IlCallo:
Cool! Thanks Milos 😁 
Do you want to take a shot to Yusuf suggestion about the anchor system?

MilosPaunovic:
Yup, I've added it to my notes; it just a question of spare time, probably sometime during next week.

PR #13812, opened 2022-06-29
Fix(ui): Check if metaKey or altKey are pressed before tabs navigation

kovalchukq:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
You are not able no navigate between pages with browser hotkeys: (alt | command + leftArrow | rightArrow) when focused on QTab.

I think we need to check - if altKey or metaKey are pressed - we do default browser behavior (back and forward)

Else If those keys are not pressed - we do QTab navigation like it was before


rstoenescu:
Thanks for contributing!

Issue #12515, opened 2022-02-18
QUploader component's background colour changes slightly when 'text-color' prop is set


vijay-singh-16:
### When 'text-color' prop is NOT set:

![no-text-color](https://user-images.githubusercontent.com/60729120/154657596-4f985598-d8d9-4496-a433-1713f43596db.jpg)

### When 'text-color' prop is set:

![grey-5-text-color](https://user-images.githubusercontent.com/60729120/154657609-55ee9e6e-5aa3-4753-930c-55967fa3035f.jpg)

This happens with any 'color' prop value.

### What did you expect to happen?

The background colour should not change when 'text-color' prop is set.

### Reproduction URL

https://codepen.io/vijay-singh-16/pen/MWOQegE?editors=101

### How to reproduce?

Refer reproduction URL.

### Flavour

Vue CLI Plugin (vue-cli-plugin-quasar)

### Areas

Components (quasar)

### Platforms/Browsers

Firefox, Chrome, Microsoft Edge

Saeid-Za:
Hello There!
I've created a pull-request that should resolve this issue, until then, there is a workaround, put this chunk of css into your global style file.
```scss
  .q-uploader__header::before{
    background: unset !important;
  }
```

rstoenescu:
Fix will be available in Quasar v2.7.5 and 1.19.4

PR #13844, opened 2022-07-01
fix(QUploader): issue with q-uploader background inconsistency (#12515)

Saeid-Za:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
When using `currentColor` as background color of  `<div class="q-uploader__header" ::before>`, background color would change slightly as the text-color of q-uploader changes.
from [W3 Specifications](https://www.w3.org/TR/css-color-3/#currentcolor):
> The value of the ‘[color](https://www.w3.org/TR/css-color-3/#color)’ property. The used value of the ‘[currentColor](https://www.w3.org/TR/css-color-3/#currentColor-def)’ keyword is the computed value of the ‘[color](https://www.w3.org/TR/css-color-3/#color)’ property. If the ‘[currentColor](https://www.w3.org/TR/css-color-3/#currentColor-def)’ keyword is set on the ‘[color](https://www.w3.org/TR/css-color-3/#color)’ property itself, it is treated as ‘color: inherit’. 

There is no color on `::before` element, from my understanding, color would be inherited from parent element (`<div class="q-uploader__header">`), which is being configured from `text-color` prop of `QUploader`, I think this is why when changing text-color property, we see a change on q-uploader header background.
but background-color of parent (`.q-uploader__header` class) is being applied to `::before` element too, thus making a weird situation where both the background and color of parent element (`<div class="q-uploader__header">`) would affect the background of the child (`<div class="q-uploader__header" ::before>`)
I've removed the `currentColor` from parent element.

### DIFF
--- a/ui/src/components/uploader/QUploader.sass
+++ b/ui/src/components/uploader/QUploader.sass
@@ -33,7 +33,6 @@
       bottom: 0
       left: 0
       pointer-events: none
-      background: currentColor
       opacity: .04
 
   &__header


rstoenescu:
Hi,

Thanks for contributing!
The change is not quite right, as removing the background makes the whole pseudo-selector useless. But more than this, it should remain as-is for the q-uploader__file class.
So pushed a fix to remove it from q-uploader__header only.

PR #13898, opened 2022-07-07
feat(app-vite): Add TS template for BEX mode (part of #8572) 

yusufkandemir:

- Feature

- No

**Other information:**

Part of #8572.

@rstoenescu due to the structure of BEX in app-webpack, it's not possible to directly&fully support TS. `js/background.js` and `js/content-script.js` in the manifest don't get processed at all, they are not a part of the build chain. So, naturally, you can't make those files TS. It's possible to make them separate entry points to make them compiled, but there is another problem. Dev server is designed to compile some stuff into `src-bex/www`, so that on development we can use `src-bex` folder as the unpacked extension, and the source code would be directly accessed without any processing, etc. So, the time you try to load the extension the browser naturally throws errors about .ts files.
So, we might be able to make some changes to make TS properly, but I think it might mean diverging JS-TS template structures&docs considerably. I think you already know all this because you created a superior structure for BEX in app-vite. So, I was wondering if you have a good solution that doesn't change everything too much. This can be added later so I am skipping it altogether now since I couldn't have decided on a good approach to this.


IlCallo:
I added this as a reference to Qv3 project board, as changing BEX right now for app-webpack would require doing breaking changes

app-vite/templates/bex/default/manifest-v2/background.js
```diff
@@ -11,20 +11,25 @@ chrome.browserAction.onClicked.addListener((/* tab */) => {
...
       chrome.storage.local.get(null, items => {
-        const result = []
-
-        // Group the items up into an array to take advantage of the bridge's chunk splitting.
-        for (const itemKey in items) {
-          result.push(items[itemKey])
-        }
-        respond(result)
+        // Group the values up into an array to take advantage of the bridge's chunk splitting.
+        respond(Object.values(items))
```
IlCallo, line 28: Shouldn't this be done on manifes-v3 js version too?

yusufkandemir: Good catch! Thanks.

PR #13948, opened 2022-07-13
chore: run tests on both nodejs 14 and 16

Hashen110:


MilosPaunovic:
Hi @Hashen110, thanks for your contribution!

However, when implementing this, there was a discussion whether to create a matrix of `Node` versions or not - decision was to fix it to `Node@14` as it is the minimal supported version right now which will save us some minutes on GH action runners. The idea behind it was that if it works with `v14`, it's highly likely that it will work with newer versions too.

Hashen110:
Sure :)

yusufkandemir:
Thanks for the suggestion. This would be a good idea for testing Node-based packages like `@quasar/app-vite` and `@quasar/app-webpack`, but we don't have any tests for them, at least not yet 🙂. As @MilosPaunovic pointed out, we decided it's not really needed for UI tests. So, I am closing this PR.

PR #14117, opened 2022-08-02
fix(vite-plugin): let Quasar be pre-bundled in dev mode

philfontaine:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

A bit of context here: https://github.com/quasarframework/quasar/discussions/14084

But TLDR:

- One of the goals of @quasar/vite-plugin is to make quasar tree-shakeable. This is not necessary in dev and makes development slower. It's faster to have a single pre-bundled dependency than make 1XX requests for each component and composable. (See https://vitejs.dev/guide/dep-pre-bundling.html#the-why)
- This PR disables this behavior in development mode, but keeps the current behavior as is for production.

In my actual use case, my first page load went from 300 requests to 170 requests so it's a nice improvement.

I tested this PR by building @quasar/vite-plugin locally then copying the dist/index.js in the node_modules/@quasar/vite-plugin/dist folder of my project.

### DIFF
--- a/vite-plugin/src/plugin.js
+++ b/vite-plugin/src/plugin.js
@@ -58,19 +58,26 @@ function getScssTransformsPlugin (opts) {
 }
 
 function getScriptTransformsPlugin (opts) {
+  let config
+
   return {
     name: 'vite:quasar:script',
+    configResolved(resolvedConfig) {
+      config = resolvedConfig
+    },
     transform (src, id) {
-      if (vueTransformRegex.test(id) === true) {
-        return {
-          code: vueTransform(src, opts.autoImportComponentCase),
-          map: null // provide source map if available
+      if (config.mode !== 'development') {
+        if (vueTransformRegex.test(id) === true) {
+          return {
+            code: vueTransform(src, opts.autoImportComponentCase),
+            map: null // provide source map if available
+          }
         }
-      }
-      else if (jsTransformRegex.test(id) === true) {
-        return {
-          code: jsTransform(src),
-          map: null // provide source map if available
+        else if (jsTransformRegex.test(id) === true) {
+          return {
+            code: jsTransform(src),
+            map: null // provide source map if available
+          }
         }
       }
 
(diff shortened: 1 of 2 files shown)


vite-plugin/src/plugin.js
```diff
@@ -58,19 +58,26 @@ function getScssTransformsPlugin (opts) {
...
     name: 'vite:quasar:script',
+    configResolved(resolvedConfig) {
+      config = resolvedConfig
+    },
     transform (src, id) {
-      if (vueTransformRegex.test(id) === true) {
-        return {
-          code: vueTransform(src, opts.autoImportComponentCase),
-          map: null // provide source map if available
+      if (config.mode !== 'development') {
```
philfontaine, line 69: This check here could (in theory) throw an exception if configResolved is never called, not sure how you wanted to handle that.

rstoenescu:
Hi,

This actually breaks some of the Quasar modes. We must point to the actual src/ files in order for correctness of code (which has flags) to be preserved. Otherwise the precompiled file(s) will be used, which are pre-built without flags (because each one has very specific use-case).

philfontaine:
Hi @rstoenescu 

From what I understand, this breaks SSR and PWA in dev mode only, correct? 

Either way, it seems to work fine in DEV + SPA. Would you be interested in an updated PR with a check on the quasar mode?

If you are not interested I can simply write my own plugin.

yusufkandemir:
@rstoenescu we can start producing another ESM entry with untouched feature flags to be replaced later at build time, just like vue, vue-i18n, etc.

rstoenescu:
@yusufkandemir working on something that might just work

rstoenescu:
For reference: https://github.com/quasarframework/quasar/commit/f1e23fc8e822b160b7193c81435aa507d85749fb

PR #14184, opened 2022-08-12
feat(create-quasar): add "defaults" option

### LINKED ISSUE #5537: Preset option in quasar create
I want to automate the creation of quasar projects. Therefore a configuration for the starter template over JSON would be very nice. The Vue CLI supports [presets](https://cli.vuejs.org/guide/plugins-and-presets.html#presets) and makes it easy to automate the creation.

I know I could create a vue cli project but I want to use the quasar cli. Are there any plans for the future to implement this? Or are there reasons why this isn't implemented yet?

IlCallo:
Partially resolves #5537 , it's a first step but more flags could be added later on

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
See https://github.com/quasarframework/quasar/issues/5537#issuecomment-1212348698

### DIFF
--- a/create-quasar/utils/index.js
+++ b/create-quasar/utils/index.js
@@ -218,6 +218,26 @@ module.exports.ensureOutsideProject = function () {
   }
 }
 
+const defaultProjectFolder = 'quasar-project'
+
+module.exports.promptsDefaults = {
+  projectType: 'app',
+  projectFolder: defaultProjectFolder,
+  overwrite: 'n',
+  packageManager: module.exports.runningPackageManager ?? 'yarn',
+  quasarVersion: 'v2',
+  scriptType: 'js',
+  engine: 'webpack',
+  productName: 'Quasar App',
+  name: module.exports.inferPackageName(defaultProjectFolder),
+  description: 'A Quasar Project',
+  author: getGitUser(),
+  typescriptConfig: 'composition',
+  css: 'scss',
+  preset: ['lint'],
+  lintConfig: 'prettier',
+}
+
 const QUASAR_VERSIONS = [
   { title: 'Quasar v2 (Vue 3 | latest and greatest)', value: 'v2', description: 'recommended' },
   { title: 'Quasar v1 (Vue 2)', value: 'v1' }
@@ -248,7 +268,7 @@ module.exports.commonPrompts = {
     type: 'text',
     name: 'productName',
     message: 'Project product name: (must start with letter if building mobile apps)',
-    initial: 'Quasar App',
+    initial: module.exports.promptsDefaults.productName,
     validate: val =>
       val && val.length > 0 || 'Invalid product name'
   },
@@ -257,7 +277,7 @@ module.exports.commonPrompts = {
     type: 'text',
     name: 'description',
     message: 'Project description:',
-    initial: 'A Quasar Project',
+    initial: module.exports.promptsDefaults.description,
     format: module.exports.escapeString,
     validate: val =>
       val.length > 0 || 'Invalid project description'
@@ -266,7 +286,7 @@ module.exports.commonPrompts = {
   author: {
     type: 'text',
     name: 'author',
-    initial: () => getGitUser(),
+    initial: module.exports.promptsDefaults.author,
     message: 'Author:'
   },
 
--- a/create-quasar/index.js
+++ b/create-quasar/index.js
@@ -13,10 +13,17 @@ console.log(
 
 const utils = require('./utils')
 
+const parseArgs = require('minimist')
+const argv = parseArgs(process.argv.slice(2), {
+  boolean: ['defaults'],
+})
+
+if(argv.defaults) {
+  require('prompts').override(utils.promptsDefaults)
+}
+
 // should error out if already inside of a Quasar project
 utils.ensureOutsideProject()
-
-const defaultProjectFolder = 'quasar-project'
 
 async function run () {
   const scope = {}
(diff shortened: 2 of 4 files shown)


create-quasar/utils/index.js
```diff
@@ -218,6 +218,26 @@ module.exports.ensureOutsideProject = function () {
...
+  engine: 'webpack',
+  productName: 'Quasar App',
+  name: module.exports.inferPackageName(defaultProjectFolder),
+  description: 'A Quasar Project',
+  author: getGitUser(),
+  typescriptConfig: 'composition',
+  css: 'scss',
+  preset: ['lint'],
+  lintConfig: 'prettier',
+}
```
yusufkandemir, line 239: `defaults` might become more useful for all kinds of people if it doesn't choose every option possible. It could set the following:
```suggestion
module.exports.promptsDefaults = {
  projectType: 'app',
  overwrite: 'n',
  packageManager: module.exports.runningPackageManager ?? 'yarn',
  quasarVersion: 'v2',
  productName: 'Quasar App',
  name: module.exports.inferPackageName(defaultProjectFolder),
  description: 'A Quasar Project',
  author: getGitUser(),
}
```
The most common aim will be creating a Quasar v2(_or whatever is the latest stable major version_) **app**. The user would highly likely enter a folder name to avoid collisions and to be able to find the project by its name. Some of the rest such as the project name(_which is usually the same as the folder name_), author(_inferred from Git_), etc. are all the defaults, the other options would be to just select between Vite and Webpack, the CSS flavor, sub features, etc. which has lots of combinations compared to the options in the suggested change.

If we merge this as is, lots of people may not find it useful as it's not even possible to enter a custom folder name.

yusufkandemir: We could also try to infer the product name by converting the folder name to the capitalized case, e.g. `quasar-app` -> `Quasar App` to make it more useful than using `Quasar App`

IlCallo: The point of `--defaults` flag is to totally avoid user interaction, useful for people trying to automate Quasar projects creation, it's not meant to speed up people creating projects, we would need other flags for that
The name itself explains it: it accepts all the defaults values
If you prefer, we can call it `-y` as it is in many other CLI tools, I used `defaults` because it should be easier to understand

Applying these changes would leave some questions open for interaction and that would defy the primary purpose of the flag

I'm more on the side of adding the `defaults` flag and then provide some other flag to customize the choices
Maybe it's better to define those other flags before merging this at this point

---

A suggestion to solve the folder name is to take the first argument from the command line for that, many other CLIs do that
So it would become `yarn create quasar my-project-folder --defaults`

I agree that the product name should be inferred by the folder name

yusufkandemir: It seems this is solving an artificial "problem" of the ability to create projects without any interaction. This is good at solving that problem, but I just don't think it's practical and doubt that anybody who wanted the presets option will be happy with this option. But, if it's gonna be implemented like this I think `-y` would be better for this.

Accepting the project name as the first argument is a good idea in any case. So, we should definitely do that. I think most, if not all, scaffolding tools use that approach.

About adding more flags, as you said this PR is not really useful unless there are more flags. But, my preferred way would be to implement the defaults approach I mentioned, which will save some time for most users and is practical. Then, we could add more flags to answer more questions. If someone wants, they could provide all the flags to create a project without any interaction, in the exact way they want. This allows room for flexibility, while accepting all then overriding the rest is more limiting and would require the user to provide almost every flag possible to be practical, IMO.

IlCallo: > It seems this is solving an artificial "problem" of the ability to create projects without any interaction

This may be true, we should probably seek some validation from the community before merging this in

alanpoulain:
This PR would be very useful to test programmatically a generated Quasar app.

IlCallo:
@alanpoulain could you check out https://github.com/quasarframework/quasar/issues/5537 , describe your use case and see if the workaround there works for you?
We're looking for more feedback before adding this feature in, it's not clear if this PR would match real use cases

yusufkandemir:
This will be included in create-quasar v4
```
  Description
    Scaffolds Quasar Apps & App Extensions
    Version 4.0.0

  Usage
    $ create-quasar [dir] [options]

    # examples:
    $ create-quasar my-app --template app --engine vite-3 --defaults
    $ create-quasar my-ae --template ae --preset prompts --preset oxlint --defaults

  Options
    --template, -t  Type of project to create: app | ae
    --overwrite, -o Overwrite existing dir if it exists
    --preset        Preset to apply (can be used multiple times)
                    - template "app" presets:
                      typescript, sass, oxlint, eslint, i18n, pinia, fbr (filename-based routing)
                    - template "ae" presets:
                      typescript, oxlint, prompts, install, uninstall
    --name          Name of the project for package.json (must be a valid npm package name)
    --author        Author name for package.json
    --no-git        Do not initialize a git repository
    --install, -i   When invoked through a package manager it's a boolean (eg. --install)
                    Otherwise, the package manager to auto-install with:
                      --install pnpm
                      --install yarn
                      --install npm
                      --install bun

    --engine, -e    (ONLY for template "app") Quasar App Local CLI to use
                      Please note that these do NOT refer to the version of
                      Vite, but rather to the @quasar/app-vite version to use:
                        vite-3 or vite-2
    --product       (ONLY for template "app") Product name for the app

    --defaults, -d  Use default values for the remaining non-specified options
    --no-color      Disable colored output
    --help, -h      Displays this message
```

PR #14213, opened 2022-08-17
Fix sass deprecation warning

bartduisters:
A community member mentioned a deprecation warning in SASS.
This commit migrates the SASS Division: https://sass-lang.com/documentation/breaking-changes/slash-div

### DIFF
--- a/i18n/cs.js
+++ b/i18n/cs.js
@@ -1,0 +1,96 @@
+export default {
+  lang: 'cs',
+  label: {
+    clear: 'Vymazat',
+    ok: 'OK',
+    cancel: 'Zrušit',
+    close: 'Zavřít',
+    set: 'Nastavit',
+    select: 'Vybrat',
+    reset: 'Resetovat',
+    remove: 'Odstranit',
+    update: 'Upravit',
+    create: 'Vytvořit',
+    save: 'Uložit',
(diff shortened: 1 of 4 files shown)


rstoenescu:
Hi,

Thanks for contributing!
This has been discussed many times. The "sass" package needs to be pinned to v1.32.12 otherwise, indeed some deprecation warnings will appear. However, it is not easy to convert to the new syntax without adding breaking changes to our developer's apps (a definitely no-go) -- hence the pinning of the "sass" package which should not be touched (manually installed) by our devs (supplied by Quasar CLI).

rstoenescu:
The pinning is also explained in the docs.

bartduisters:
Alright. I mentioned it to the community.
However, I don't find v1.32.12 referenced in the docs, if you have the link available, it's handy to have as a response on this issue.

MilosPaunovic:
> Alright. I mentioned it to the community. However, I don't find v1.32.12 referenced in the docs, if you have the link available, it's handy to have as a response on this issue.

https://quasar.dev/start/vite-plugin#installation - Check the green tip.

bartduisters:
That page doesn't pop up in the search results when looking for 'sass'. Which is what I tried to find the page :D

PR #14241, opened 2022-08-22
fix(ui/uploader-core): Expose public fields

yusufkandemir:

- Bugfix
- Feature

- No

**Other information:**
They were made available on the uploader scope where it is advertised as the QUploader instance itself. This makes it feel like they are accessible on the instance itself.
So, I am exposing those fields in the component instance too, aiming to unify both references.

### DIFF
--- a/ui/src/components/uploader/uploader-core.js
+++ b/ui/src/components/uploader/uploader-core.js
@@ -1,4 +1,4 @@
-import { h, ref, isRef, computed, watch, provide, onBeforeUnmount, getCurrentInstance } from 'vue'
+import { h, ref, computed, watch, provide, onBeforeUnmount, getCurrentInstance, unref } from 'vue'
 
 import QBtn from '../btn/QBtn.js'
 import QIcon from '../icon/QIcon.js'
@@ -11,6 +11,7 @@ import useFile, { useFileProps, useFileEmits } from '../../composables/private/u
 import { stop } from '../../utils/event.js'
 import { humanStorageSize } from '../../utils/format.js'
 import { uploaderKey } from '../../utils/private/symbols.js'
+import { injectMultipleProps } from '../../utils/private/inject-obj-prop.js'
 
 function getProgressLabel (p) {
   return (p * 100).toFixed(2) + '%'
@@ -464,6 +465,18 @@ export function getRenderer (getPlugin) {
 
   // expose public methods
   Object.assign(proxy, publicMethods)
+
+  injectMultipleProps(proxy, {
+    canAddFiles: () => canAddFiles.value,
+    canUpload: () => canUpload.value,
+    uploadSizeLabel: () => uploadSizeLabel.value,
+    uploadProgressLabel: () => uploadProgressLabel.value,
+
+    ...Object.entries(state).reduce((acc, [ key, val ]) => {
+      acc[ key ] = () => unref(val)
+      return acc
+    })
+  })
 
   return () => {
     const children = [
--- a/ui/src/components/uploader/uploader-core.js
+++ b/ui/src/components/uploader/uploader-core.js
@@ -473,7 +473,7 @@ export function getRenderer (getPlugin) {
     ...Object.entries(state).reduce((acc, [ key, val ]) => {
       acc[ key ] = () => unref(val)
       return acc
-    })
+    }, {})
   })
 
   return () => {
--- a/ui/src/components/uploader/uploader-core.js
+++ b/ui/src/components/uploader/uploader-core.js
@@ -453,9 +453,7 @@ export function getRenderer (getPlugin) {
     }
 
     for (const key in state) {
-      acc[ key ] = isRef(state[ key ]) === true
-        ? state[ key ].value
-        : state[ key ]
+      acc[ key ] = unref(state[ key ])
     }
 
     // TODO: (Qv3) Put the QUploader instance under `ref`
--- a/ui/src/components/uploader/uploader-core.js
+++ b/ui/src/components/uploader/uploader-core.js
@@ -443,8 +443,6 @@ export function getRenderer (getPlugin) {
     abort: state.abort
   }
 
-  // TODO: the result of this computed, especially the dynamic part, isn't currently typed
-  // This result in an error with Volar when accessing the state (eg. files array)
   const slotScope = computed(() => {
     const acc = {
       canAddFiles: canAddFiles.value,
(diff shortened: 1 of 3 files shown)


yusufkandemir:
I also saw and answered some questions in the community and now I am sure that these were exposed in v1 and somehow got forgotten in v2. So, this is definitely a bug.

IlCallo:
Seems like in v1 they were data on the component instance
https://github.com/quasarframework/quasar/blob/19eead2dd8f581417b6452f4fb9f035f0c106d31/ui/src/components/uploader/QUploaderBase.js#L44-L55

The component instance itself was then provided as the slots scope
https://github.com/quasarframework/quasar/blob/19eead2dd8f581417b6452f4fb9f035f0c106d31/ui/src/components/uploader/QUploaderBase.js#L314
https://github.com/quasarframework/quasar/blob/19eead2dd8f581417b6452f4fb9f035f0c106d31/ui/src/components/uploader/QUploaderBase.js#L351

Back at that time there wasn't a clear distinction between internal state and public API, that's probably why the whole component instance were provided
When the distinction came to be together with Composition API, these properties had been hidden into the component and only exposed via the slot scope

https://github.com/quasarframework/quasar/blob/605f477b1b52d3a68e7bff4ee76c766510644805/ui/src/components/uploader/uploader-core.js#L447-L465

https://github.com/quasarframework/quasar/blob/605f477b1b52d3a68e7bff4ee76c766510644805/ui/src/components/uploader/uploader-core.js#L338
https://github.com/quasarframework/quasar/blob/605f477b1b52d3a68e7bff4ee76c766510644805/ui/src/components/uploader/uploader-core.js#L375

That same state is then provided as helpers when creating a custom uploader
https://github.com/quasarframework/quasar/blob/605f477b1b52d3a68e7bff4ee76c766510644805/ui/src/components/uploader/uploader-core.js#L91

Eg [the Axios one](https://gist.github.com/IlCallo/96b0ca99d7400e5a795e8b918dbc2982#file-axios-uploader-plugin-ts-L92-L98) I created

It seems to me that it has been done on purpose to avoid people relying on those implementations details from outside the component, as they could be confusing: people would easily expect them to be reactive references (as they are internally and when provided to a custom adapter) but won't be when provided from the component instance, and will be exposed as readonly props instead

I feel their usage should be limited only to advanced use cases, when creating a new adapter and accessing them via `helpers` property, and that the docs are the ones out of sync in this case

IlCallo:
To add up: I'm not totally against this, as there are already other cases where we export intermediate internal refs as readonly (QTable and fields), but I don't really see the need to do so in this case.
If we add them into the component instance itself, we need to add a section/note somewhere into QUploader docs to describe those props more in depth and specifying that those exposed as the component public API are readonly, while the ones exposed when creating a new adapter are actually meant to be used

What do you think?

yusufkandemir:
> They were made available on the uploader scope where it is advertised as the QUploader instance itself.

What I mean by this is the part in QUploader's JSON API which added almost 2 years ago: https://github.com/quasarframework/quasar/commit/403fbfb5d816d1f245425c6fdf110e2e192c5de1

https://github.com/quasarframework/quasar/blob/403fbfb5d816d1f245425c6fdf110e2e192c5de1/ui/src/components/uploader/QUploader.json#L67-L77

If the intention was to not expose them on purpose, then the explanation should be updated to reflect that. If the explanation is true, then this means these fields should be exposed. Let's leave aside the original intention(_since we can't really know it_), and discuss the way it should be done. As the users can access these in slot scope, it's already a part of the public API, so I wouldn't see a big problem exposing these from the instance as well. Regarding whether it would have proper use cases, here is a use case I got from SirPrise#2223 our Discord server:

https://discord.com/channels/415874313728688138/806171748046733332/1014546856497389578
> In my old program the uploader was part of a form, that gets submitted like this: <form @submit.prevent.stop="save()">
> 
> Id like to access the "inserted files" (into Uploader) in methods triggered by this save() method. Is there no way? So to do it analog to the old program (refresh files in backend when form submits) is it not possible to use Uploader Component https://quasar.dev/vue-components/uploader ? I am a bit confused because looking at quasar.dev the Q-Uploader API seems quite sophisticated and accessing the files put into it programatically seems like a rather basic use case to me. Thank you for your reply! Im still pretty rookie, loving quasar, any input on this topic is gold to me. I am trying to figure out a path to take from here. Does it seem like a good idea to program a custom uploader component / write some logic for that in the component?

They noticed this while upgrading from Quasar v1 to v2. In this case, using template refs makes sense, and would be the simplest option. On the other hand, it would still be possible with a custom uploader component, which would be a longer and less Vue-conventional way of doing it. But, either way would work. Please let me know what you think about this use case.

> people would easily expect them to be reactive references

IMO, this wouldn't be a concern since we will start documenting them in #14243. We can explicitly state there that the fields are read-only. It's technically possible to make public fields writeable as well, so we can add a JSON API prop to state whether a field is read-only or not, then display it in the Docs API Card.

> If we add them into the component instance itself, we need to add a section/note somewhere into QUploader docs to describe those props more in depth and specifying that those exposed as the component public API are readonly, while the ones exposed when creating a new adapter are actually meant to be used

You can check if their individual descriptions are good enough in #14243. About the difference between them in the slot scope/template ref and `createUploaderComponent`, I agree we should clearly state that if they need to write to these props or need complex logic, they need to use `createUploaderComponent`.

IlCallo:
Yeah, the fact that they are already exposed somewhere is what make me think it can be ok, if we document it properly.
Can you add some docs on that topic?

ui/src/components/uploader/uploader-core.js
```diff
@@ -442,30 +443,24 @@ export function getRenderer (getPlugin) {
...
+      acc[ key ] = () => unref(val)
+      return acc
+    }, {})
   })
 
-  // expose public methods
-  Object.assign(proxy, publicMethods)
+  // TODO: (Qv3) Put the instance under `ref` property for consistency and flexibility
+  // computed(() => ({ ref: proxy, /* new stuff can be added if needed */ }))
+  const slotScope = computed(() => proxy)
```
rstoenescu: Hi, putting the proxy under computed() works, but it makes the whole proxy Object reactive. Besides the perf downside, the implications are huge and we can bump into all sorts of trouble. Better to build an object containing all the stuff then "inject" it into proxy and the slotScope computed.

rstoenescu (changes requested):

ui/src/components/uploader/uploader-core.js
yusufkandemir: @rstoenescu #14967

Issue #13217, opened 2022-04-20
Types of QUploader instance properties are missing


paya-cz:
### What happened?

The `QUploader` scoped param is not fully typescript-ready. I used the [official example for QUploader custom header](https://quasar.dev/vue-components/uploader#example--custom-header) and [although some properties are already properly typed](https://github.com/quasarframework/quasar/issues/11090), others from the example are missing, notably:

- files
- queuedFiles
- uploadedFiles
- isUploading
- canAddFiles
- canUpload
- uploadSizeLabel
- uploadProgressLabel

![Screenshot 2022-04-20 at 16 39 17](https://user-images.githubusercontent.com/10402560/164256670-3ba39a6f-077d-4106-bb5a-519ff846de31.png)

### What did you expect to happen?

The official custom header example (and others) should work as-is with TypeScript. All the properties available on scoped params should be properly typed.

### Reproduction URL

https://github.com/paya-cz/quasar-scoped-param-typescript-issue

### How to reproduce?

1. Clone the repo.
2. Look into `src/components/ExampleComponent.vue` using Visual Studio Code
3. Observe TypeScript errors in the `<template>` section

Notes:
- I didn't use Codepen / jsFiddle because they are not setup with TypeScript
- The project is just standard `npm init quasar` with Quasar v2, TypeScript, CLI+Vite
- There is just one single commit beyond `npm init quasar` that you can inspect to see the diff, but it is just a copy/paste of the QUploader example

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

TypeScript Support

yusufkandemir:
QInput is also missing instance property types #11510. I am linking it here because they are related.

rstoenescu:
Fix will be available in Quasar v2.8

paperisque:
not fixed 2.8.2!

yusufkandemir:
@paperisque it would have been more helpful if you share what's not fixed. But, `isUploading` seems to be missing, so I added it in #14399. If something else is missing please let us know.

yusufkandemir:
For anyone else that comes across this: Volar doesn't automatically pick up changes to `node_modules`. So, make sure to restart VS Code or just the Volar extension after installing/upgrading dependencies. With 2.8.2, you will still get an error regarding `isUploading`, but everything else will work properly. Starting from v2.8.3(_not released yet_), everything should work fine.

paperisque:
[...]

PR #14243, opened 2022-08-23
feat(docs&api&ts): Add new top-level section 'fields'

### LINKED ISSUE #11510: The prop 'hasError' generates a TS error, types that are generated based on the docs are incorrect
First of all I deeply apologize for my bad English. I wanted to use the internal validation from quasar and my code is below this text also what I've learned is from the docs here : 
https://quasar.dev/vue-components/input#example--form-validation
But The prop 'hasError' (that is used for internal validation on inputRef and it gives 'true' if the inputs has error) generates a TS error that says `Property 'hasError' does not exist on type QInput` and I guess there is a problem with the types that are generated based on the docs.

As you can see in the simple code below, I have to use a // @ts-expect-error because the code `inuputRef.value?.hasError` generates an error and it also gives me what I wanted for validations at the same time.
the TS error says : `Property 'hasError' does not exist on type QInput`
but the code is correct and the property 'hasError' does exist on type QInput and it's just the TS types that are generated based on the docs and they are incorrect I guess.

Thanks in advance  🙏 

My example code that shows the problem but first remove the // @ts-expect-error from it :

```
<template>
  <q-page class="flex flex-center" padding>
    <form @submit.prevent.stop="onSubmit">
      <q-card class="full-width" style="max-width: 350px">
        <q-card-section class="bg-primary text-white">
          <div class="text-h6">login</div>
          <div class="text-subtitle2">
           subtext
          </div>
        </q-card-section>

        <q-separator />

        <q-card-actions vertical>
          <q-input
            ref="usernameRef"
            v-model="username"
            standout="bg-teal text-white"
            label="username"
            class="q-my-md"
            hint="this is username"
            lazy-rules
            :rules="usernameRules"
          >
            <template #prepend>
              <q-icon name="fingerprint" />
            </template>
          </q-input>

          <q-input
            ref="passwordRef"
            v-model="password"
            standout="bg-teal text-white"
            label="password"
            class="q-mb-md"
            hint="this is password"
            :type="isPwd ? 'password' : 'text'"
            lazy-rules
            :rules="passwordRules"
          >
            <template #prepend>
              <q-icon name="password" />
            </template>
            <template #append>
              <q-icon
                :name="isPwd ? 'visibility_off' : 'visibility'"
                class="cursor-pointer"
                @click="isPwd = !isPwd"
              />
            </template>
          </q-input>

          <q-btn
            color="secondary"
            icon="login"
            label="login"
            class="q-mb-sm"
            @click="onSubmit"
          />
        </q-card-actions>
      </q-card>
    </form>
  </q-page>
</template>

<script lang="ts">
import { defineComponent, ref } from 'vue';
import { useQuasar, QInput } from 'quasar';

export default defineComponent({
  name: 'Login',
  setup() {
    const $q = useQuasar();

    const username = ref<string>('');
    const usernameRef = ref<QInput | null>(null);

    const usernameRules = [
      (value: string) =>
        (value !== null && value !== '') ||
        'username cant be null',
    ];

    const password = ref<string>('');
    const passwordRef = ref<QInput | null>(null);

    const passwordRules = [
      (value: string) =>
        (value !== null && value !== '') || 'password cant be null',
    ];

    const isPwd = ref<boolean>(true);

    function onSubmit(): void {
      usernameRef.value?.validate();
      passwordRef.value?.validate();

      // @ts-expect-error     I have to temporary use a (ts-expect-error). Using the prop 'hasError' generates a ts error but it's correct and the docs are incorrect in this point it seems, and there is a problem with the types that are generated based on the docs I guess.
      if (usernameRef.value?.hasError || passwordRef.value?.hasError) {
        $q.notify({
          icon: 'warning',
          color: 'negative',
          message: 'please fix the errors first the press it again',
        });
      } else {
        $q.notify({
          icon: 'done',
          color: 'positive',
          message: 'succeed',
        });
      }
    }

    return {
      username,
      usernameRef,
      usernameRules,
      password,
      passwordRef,
      passwordRules,
      isPwd,
      onSubmit,
    };
  },
});
</script>
```

yusufkandemir:

- Feature
- Documentation

- No

**Other information:**
Closes #11510, #13217

Depends on #14241 because QUploader isn't exposing the fields.

This adds a top-level property called "fields" to our JSON API system. I analyzed all components that have public instance fields and added JSON API content for them. The docs API cards now have a section called "fields" at the end if applicable. We are also generating the types for fields in the instance type interfaces, e.g. `QUploader`

### DIFF
--- a/ui/src/components/uploader/QUploader.json
+++ b/ui/src/components/uploader/QUploader.json
@@ -152,5 +152,53 @@
         }
       }
     }
+  },
+
+  "fields": {
+    "files": {
+      "type": "Array",
+      "desc": "List of all files",
+      "__exemption": [ "examples" ]
+    },
+
+    "queuedFiles": {
+      "type": "Array",
+      "desc": "List of files that are waiting to be uploaded",
+      "__exemption": [ "examples" ]
+    },
+
+    "uploadedFiles": {
+      "type": "Array",
+      "desc": "List of files that have been uploaded",
+      "__exemption": [ "examples" ]
+    },
+
+    "uploadedSize": {
+      "type": "Number",
+      "desc": "Size of all uploaded files in bytes",
+      "examples": [ 1024 ]
+    },
+
+    "uploadSizeLabel": {
+      "type": "String",
+      "desc": "Label for the size total of all files",
+      "examples": [ "1.0MB" ]
+    },
+
+    "uploadProgressLabel": {
+      "type": "String",
+      "desc": "Label for the upload progress (in %)",
+      "examples": [ "52.76%" ]
+    },
+
+    "canAddFiles": {
+      "type": "Boolean",
+      "desc": "Whether new files can be added to the list"
+    },
+
+    "canUpload": {
+      "type": "Boolean",
+      "desc": "Whether the files can be uploaded"
+    }
   }
 }
--- a/docs/src/components/DocApiEntry.js
+++ b/docs/src/components/DocApiEntry.js
@@ -277,7 +277,7 @@ function copyPropName (propName) {
 
 const describe = {}
 
-describe.props = props => {
+const describePropsLike = props => {
   const child = []
 
   for (const propName in props) {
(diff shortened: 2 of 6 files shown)


ui/src/components/uploader/QUploader.json
```diff
@@ -152,5 +152,53 @@
         }
       }
     }
+  },
+
+  "fields": {
```
IlCallo, line 157: If we don't merge in https://github.com/quasarframework/quasar/pull/14241 as I suggested to avoid, these definitions need to be moved into scoped slots params

rstoenescu:
Excellent work @yusufkandemir !
Will make a few changes and push them today. I'm aware this PR depends on the QUploader one. Will try to make some tweaks to that one as well.

PR #11347, opened 2021-11-15
fix/refactor(ts&api): Rework the 'required' logic

yusufkandemir:
**What kind of change does this PR introduce?**

- Bugfix
- Feature
- Refactor
- Build-related changes

**Does this PR introduce a breaking change?**

- No

**The PR fulfills these requirements:**

- It's submitted to the `dev` branch

**Other information:**

See https://github.com/quasarframework/quasar/pull/11109#issuecomment-964220192

**Notable Changes:**
- Reworking the 'required' logic simplified the code and allowed us to fix a lot of stuff around it. A lot of callbacks and deeper properties now have the correct 'required' status.
- As a result of the unification of the regular and method prop generation logic, we now have more `@param` and `@returns` tags generated.
- Replaced some of the generic object types with more fitting alternatives like Element and Error.
- Fixed the parameter type of `LoadingBar.setDefaults`
- Simplified and improved some API definitions, descriptions, etc.

```diff
 📜 Changes for dist/types/index.d.ts

Index: /home/yusuf/dev/github/quasarframework/quasar/ui/dist/types/index.d.ts
===================================================================
--- /home/yusuf/dev/github/quasarframework/quasar/ui/dist/types/index.d.ts
+++ /home/yusuf/dev/github/quasarframework/quasar/ui/dist/types/index.d.ts
@@ -26,9 +26,9 @@
   isActive: boolean;
   /**
    * The DOM element used as root for fullscreen, otherwise 'null'
    */
-  activeEl: LooseDictionary;
+  activeEl: Element;
   /**
    * Request going into Fullscreen (with optional target)
    * @param target Optional Element of target to request Fullscreen on
    * @returns A Promise with the outcome (true -> validation was a success, false -> invalid models detected)
@@ -350,11 +350,11 @@
    */
   increment: (amount?: number) => void;
   /**
    * Set the inner QAjaxBar's props
-   * @param ...props QAjaxBar component props
+   * @param props QAjaxBar component props
    */
-  setDefaults: (...props: any[]) => void;
+  setDefaults: (props: LooseDictionary) => void;
 }
 
 export interface LocalStorage {
   /**
@@ -2343,8 +2343,9 @@
      */
     btnProps: LooseDictionary;
     /**
      * Default trigger when clicked/tapped on
+     * @param evt JS event object
      */
     onClick: (evt: LooseDictionary) => void;
   }) => VNode[];
 }
@@ -2974,16 +2975,22 @@
    */
   yearsInMonthView?: boolean | undefined;
[...]

PR #14345, opened 2022-09-08
fix(ui/QTable): Use correct definition for @request's filter

yusufkandemir:

- Bugfix
- Documentation

- No

Anyone who was misled by the API docs(_code examples were correct_) and has passed a function instead of a string/object **should be warned**. **Their code will still work** since they can send a function filter to `requestServerInteraction` and expect the function they've sent in `onRequest` and use it as they wish, but **we should advise against this**. That is because if they don't pass in a function when calling `requestServerInteraction` or QTable triggers `@request` itself, they will receive QTable's `filter` prop, which can be string/object/undefined.

- Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs) or explained in the PR's description.

**Other information:**

In the past, I have made some investigations about the type and had a discussion with @rstoenescu:
[`ab3708d` (#11347)](https://github.com/quasarframework/quasar/pull/11347/commits/ab3708d8a01e5fda0ed286085f76a40f4292a935)

But, yesterday, someone from the community [posted this problem](https://discord.com/channels/415874313728688138/807654245640962059/1017061771313418260). After some investigation and discussion today, I am pretty confident that JSON API and docs examples are wrong. I created this PR to fix&clarify this problem.

@metalsadman also tagging you here since we had a [brief discussion about this on our Discord server](https://discord.com/channels/415874313728688138/807654245640962059/1017328496852336710).

### DIFF
--- a/docs/public/examples/QTable/Synchronizing.vue
+++ b/docs/public/examples/QTable/Synchronizing.vue
@@ -1,6 +1,7 @@
 <template>
   <div class="q-pa-md">
     <q-table
+      ref="tableRef"
       title="Treats"
       :rows="rows"
       :columns="columns"
@@ -8,8 +9,8 @@
       v-model:pagination="pagination"
       :loading="loading"
       :filter="filter"
-      @request="onRequest"
       binary-state-sort
+      @request="onRequest"
     >
       <template v-slot:top-right>
         <q-input borderless dense debounce="300" v-model="filter" placeholder="Search">
@@ -90,6 +91,7 @@ const originalRows = [
 
 export default {
   setup () {
+    const tableRef = ref()
     const rows = ref([])
     const filter = ref('')
     const loading = ref(false)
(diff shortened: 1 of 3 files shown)


engineertdog:
Thanks for the PR, Yusuf. One comment though. 

`tableRef.value.requestServerInteraction` should be the following instead, because `tableRef` can be undefined. 

`tableRef.value?.requestServerInteraction`

yusufkandemir:
@engineertdog that's not entirely correct. You wouldn't be able to trigger server interaction and always have an empty table otherwise. `tableRef` will always be defined in the `onMounted` hook and anywhere that gets called everything is settled. That's how template refs work. It would only be `undefined` if you didn't return it(_which would be a bigger mistake_) or you have `v-if` that evaluates to `false` on the `q-table`. So, in this example, it will always be undefined, so it's correct.

PR #14386, opened 2022-09-14
fix(QField): prevent focus when clicking on bottom slot #14383

### LINKED ISSUE #14383: input element - weird behavior (unexpected focus) when clicked area below element
### What happened?

if your input have validation rules (and hence ui element for notification message below input)
input element got focused when user clicks below input. There is no indication why clicking below would make input focused. Also - there is no focusing if input doesn't have validation rules. I mean this behavior is confusing and unexpected for user.
(same behavior is when Select have validation rule)

no focus in input element

### Reproduction URL

https://codepen.io/yureckey/pen/gOzwqNv

### How to reproduce?

1. go to provided link
2. click in an area under any input element

### Flavour

UMD

### Areas

Components (quasar)

### Platforms/Browsers

Firefox, Chrome

### Quasar info output

```Shell
latest
```

### Relevant log output

_No response_

### Additional context

_No response_

pdanpdan:
close #14383

### DIFF
--- a/ui/src/components/field/QField.js
+++ b/ui/src/components/field/QField.js
@@ -18,8 +18,6 @@ function getTargetUid (val) {
   return val === void 0 ? `f_${uid()}` : val
 }
 
-const preventClickFocus = { click: prevent }
-
 export default Vue.extend({
   name: 'QField',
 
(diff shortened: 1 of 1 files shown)


rstoenescu:
Do not cache the "on" prop as we'll bump into trouble. Rather use a direct reference for the click evt

PR #14388, opened 2022-09-14
feat(a11y): improve QExpansionItem, QSelect, QChip, QDialog, QMenu, QFab, QBtnDropdown #14338

### LINKED ISSUE #14338: [a11y] QExpansionItem - use of 'aria-expanded' and 'aria-label'
### What happened?

The expansion item currently does not provide an `aria-expanded` indicating the state of the item. Also we should be able to add an `aria-label` attribute to everything with a tabindex of 0  within that component so we can signal to screen readers what will happen when using that button/link.

Expansion item should be accessible

### Reproduction URL

https://codepen.io/Mett-codepens/pen/wvjKjpx

### How to reproduce?

Use the element inspector inside the codepen to see the issue. The element that toggles the expansion item should have an `aria-expanded` property.

### Flavour

UMD

### Areas

Components (quasar)

### Platforms/Browsers

Firefox, Chrome, Safari, Microsoft Edge, iOS, Android, Electron

### Quasar info output

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

pdanpdan:
close #14338

### DIFF
--- a/ui/lang/sr.js
+++ b/ui/lang/sr.js
@@ -15,8 +15,8 @@ export default {
     search: 'Traži',
     filter: 'Filter',
     refresh: 'Osveži',
-    expand: function (label) { return label ? `Прошири "${label}"` : 'Проширити' },
-    collapse: function (label) { return label ? `Скупи "${label}"` : 'Колапс' }
+    expand: function (label) { return label ? `Proširi "${label}"` : 'Proširiti' },
+    collapse: function (label) { return label ? `Skupi "${label}"` : 'Kolaps' }
   },
   date: {
     days: 'Nedelja_Ponedeljak_Utorak_Sreda_Četvrtak_Petak_Subota'.split('_'),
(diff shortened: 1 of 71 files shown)


ui/lang/sr.js
```diff
@@ -14,7 +14,9 @@ export default {
     create: 'Dodaj',
     search: 'Traži',
     filter: 'Filter',
-    refresh: 'Osveži'
+    refresh: 'Osveži',
+    expand: function (label) { return label ? `Proširi "${label}"` : 'Proširiti' },
+    collapse: function (label) { return label ? `Skupi "${label}"` : 'Kolaps' }
```
MilosPaunovic, line 19: `Kolaps` should be changed to `Skupiti`, as it is the opposite of `Proširiti` and better translation of english `Collapse` in this context. The current word will most often be translated to have non related meaning.

rstoenescu: Thanks for looking at this! Pushed the fix.

PR #14403, opened 2022-09-15
feat(ui): Offer callable builtin validation rules

yusufkandemir:

- Feature
- Code style update

- No

**Other information:**

It will now be possible to create rules by composing builtin rules or providing error messages for them.
```html
<q-input
  :rules="[
    (val, rules) => rules.email(val) || 'Please enter a valid email address'
  ]"
/>
```

### DIFF
--- a/ui/src/composables/private/use-validate.js
+++ b/ui/src/composables/private/use-validate.js
@@ -144,7 +144,7 @@ export default function (focused, innerLoading) {
       let res
 
       if (typeof rule === 'function') {
-        res = rule(val)
+        res = rule(val, testPattern)
       }
       else if (typeof rule === 'string' && testPattern[ rule ] !== void 0) {
         res = testPattern[ rule ](val)
--- a/ui/src/composables/private/use-validate.json
+++ b/ui/src/composables/private/use-validate.json
@@ -32,7 +32,8 @@
       "desc": "Array of Functions/Strings; If String, then it must be a name of one of the embedded validation rules",
       "examples": [
         ":rules=\"[ val => val.length <= 3 || 'Please use maximum 3 characters' ]\"",
-        ":rules=\"[ 'fulltime' ]\""
+        ":rules=\"[ 'fulltime' ]\"",
+        ":rules=\"[ (val, rules) => rules.email(val) || 'Please enter a valid email address' ]\""
       ],
       "category": "behavior"
     },
--- a/ui/types/api/validation.d.ts
+++ b/ui/types/api/validation.d.ts
@@ -1,4 +1,5 @@
 // Keep in sync with ui/src/utils/patterns.js
 export type EmbeddedValidationRule = "date" | "time" | "fulltime" | "timeOrFulltime" | "email" | "hexColor" | "hexaColor" | "hexOrHexaColor" | "rgbColor" | "rgbaColor" | "rgbOrRgbaColor" | "hexOrRgbColor" | "hexaOrRgbaColor" | "anyColor";
 
-export type ValidationRule<T = any> = EmbeddedValidationRule | ((value: T) => boolean | string | Promise<boolean | string>);
+type EmbeddedValidationRuleFn<T = any> = (value: T) => boolean
+export type ValidationRule<T = any> = EmbeddedValidationRule | ((value: T, rules: Record<EmbeddedValidationRule, EmbeddedValidationRuleFn<T>>) => boolean | string | Promise<boolean | string>);
--- a/ui/types/api/validation.d.ts
+++ b/ui/types/api/validation.d.ts
@@ -1,5 +1,24 @@
 // Keep in sync with ui/src/utils/patterns.js
-export type EmbeddedValidationRule = "date" | "time" | "fulltime" | "timeOrFulltime" | "email" | "hexColor" | "hexaColor" | "hexOrHexaColor" | "rgbColor" | "rgbaColor" | "rgbOrRgbaColor" | "hexOrRgbColor" | "hexaOrRgbaColor" | "anyColor";
+export type EmbeddedValidationRule =
+  | "date"
+  | "time"
+  | "fulltime"
+  | "timeOrFulltime"
+  | "email"
+  | "hexColor"
+  | "hexaColor"
+  | "hexOrHexaColor"
+  | "rgbColor"
+  | "rgbaColor"
+  | "rgbOrRgbaColor"
+  | "hexOrRgbColor"
+  | "hexaOrRgbaColor"
+  | "anyColor";
 
-type EmbeddedValidationRuleFn<T = any> = (value: T) => boolean
-export type ValidationRule<T = any> = EmbeddedValidationRule | ((value: T, rules: Record<EmbeddedValidationRule, EmbeddedValidationRuleFn<T>>) => boolean | string | Promise<boolean | string>);
+type EmbeddedValidationRuleFn<T = any> = (value: T) => boolean;
+export type ValidationRule<T = any> =
+  | EmbeddedValidationRule
+  | ((
+      value: T,
+      rules: Record<EmbeddedValidationRule, EmbeddedValidationRuleFn<T>>
+    ) => boolean | string | Promise<boolean | string>);


rstoenescu:
Nice one!

Tofandel:
Uhh this is a breaking change


Imagine this validation rule
```js
const required = (v: any, msg = 'Required field') => {
    return (Array.isArray(v) ? v.length > 0 : !!v) || msg;
};
```

This would have been fine to use as `:rules=[required]` and `:rules=[(v) => required(v, 'Some message')]` but the first usage is not valid anymore because now msg will be an object

It bit me and I was trying to figure out why my fields were not required anymore

I'm inclined to say please revert and instead simply suggest to import the rules using `import { patterns } from 'quasar'`

And use like
```
<q-input
  :rules="[
    (val) => patterns.testPatterns.email(val) || 'Please enter a valid email address'
  ]"
/>
```
I don't think they warrant being parameters of this function as they're not documented and don't bring any context for validation (they are utils and can be imported)

pdanpdan:
I don't understand what you mean. You could not have used that form of function with a `msg` when using directly in rules.

Tofandel:
Yes you could that's the point you can use it both like this and like this

`:rules=[required]` and `:rules=[(v) => required(v, 'Some message')]`

`msg` has a default value, so it's much shorter to write when you don't need to change it instead of having to write `:rules=[(v) => required(v) || 'Required field']`

But adding parameters breaks this

Now the only usable form of the former rule is `:rules=[(v) => required(v)]` which is also less readable

Tofandel:
@rstoenescu Can we get this reverted before it becomes an issue to revert because the new feature is being used please?

This is breaking all of my projects using rules like this

Given that those rules can be imported like this
```js

const { testPattern: rules } = patterns;

<q-input
  :rules="[
    (val) => rules.email(val) || 'Please enter a valid email address'
  ]"
/>
```

I see very little benefit in this breaking change

phgrund:
I second @Tofandel in this one
This feature broke all of my projects too, since I rely on some extra arguments

Example:
```js
export const required = (val, msg) => {
  return !!val || msg
}
```
```html
<q-input :rules="[required]" />
```

danbars:
Is there a verdict on this one? 
Is it going to be reverted?

rstoenescu:
Hi @Tofandel @phgrund @danbars ,

We cannot revert this so late after releasing it. But the main point is that it's not a breaking change. Sure, in your case it might seem so, but you were supplying non-API compliant functions to begin with. Quasar's API was: supply a function with one parameter; then we enhanced it with another optional param.

Now I empathize with your scenario, I do understand the predicament your in, but you have a few options:
* Add a new second param to your validation functions too then adjust their calls throughout the code; this will however leave you hanging if (for any reason, a third parameter is ever added to the Quasar API)
* (recommended) Create a Quasar API compliant wrapper for your validation functions and export/import that instead

Tofandel:
@rstoenescu 14 days is not very late and is still in a revertable time frame, there is no documentation on the rules and I highly doubt anybody is using this at all

Adding parameters to an API in js is a breaking change. Typescript will tell you the same because your code will compile fine and stop compiling after adding a parameter with a different type
In js, functions are polymorphic and as much as you want it not to be, the function is API compliant, that you did not anticipate it's usage is fine and understandable, but what matters is how you deal with it and denying it won't help.

Semver is quite specific on that and if an api change can break existing code it's by definition a breaking change 

I know I'm not the only dummy who thought about using this shortcut, and like this I'm sure you'll break a lot of projects and get a lot more support tickets in the following weeks as people start to upgrade and realise none of their forms work as expected, I'm always on the front row because I use unit tests with renovate and know straight away when something is broken, in this case it went past me because it was a silent breakage that was not accounted for, the fields just stopped being validated and we noticed it in staging of one of our project and then had to fix 4 more projects with this issue

Quasar is very popular and just realise that you probably still have hundreds of thousands of projects that are going to upgrade in the following weeks or months, vs the maybe 1-3 new projects that started using this (and are more likely to see the revert because in the build phase), reverting here is for the greater good

Semver has guidelines about this kind of issues, one of them is to introduce a new major with the new api and revert it on the the current version, in any case new features should normally be in minor version and not patch version like this one happened

In the end it's up to you, personnally I think rules should forever stay simple, they just don't need extra parameters

I'm merely using the projects at my own risks, I love it and praise it but not adhering to semver is costing the trust I have in it

rstoenescu:
I agree on everything but the definition of breaking change.
Conceptually, following the same logic, adding a new prop to a component would fall under this category as well. Imagine you were using a component with the official props and you supplied some bogus ones as well, then we release a prop with a matching name and the component suddenly does something that it wasn't supposed to.

We follow the semver notation, however with a small twist. For small/unimportant/not-critical new features I don't bump the minor version because such features don't deserve the whistles and blows and we'd be on v[1|2].5467676.0 version for both Quasar v1/v2.

But like I said, I empathize on this particular scenario here. **I'd like more people to weigh in** and if we receive enough backlash I'll revert it -- though I'd explicitly want these people to say that they will be taking over if/when other people jump in to complain about removing this feature.

rstoenescu:
Not saying this feature will stay in Quasar, but in the mean time, I'll throw in a third possibility. I'ts not ideal, but it's an idea. To change all validation methods to check for the second param if it's an Object or not then decide if your original behavior should be respected.

yusufkandemir:
I also do understand the concerns. Putting default messages or composing complex validation rules is a typical case. I will provide some examples below. It's also sad that your apps break. But, this is not a breaking change. You were just using the API in an unintended way. You could also experience this issue with components as @rstoenescu said, or when passing non-existing properties to functions that accept object properties, etc. The `rules` prop expects a string or a function that returns a string, boolean, or a Promise of them. There can also be some cases where another extra param is not documented and is meant to be only for internal use. If you are supplying a function as a callback to anywhere, you should comply with the documented use-case and don't pass or expect anything extra. You should keep these in mind even when using native JS APIs.

Please read these resources for a better understanding:
1. [A JavaScript Optional Argument Hazard - Allen Wirfs-Brock](https://wirfs-brock.com/allen/posts/166)
2. [Using parseInt() with map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#using_parseint_with_map)

### Explanation and solutions with code

Let's assume we are back in time when this PR didn't get released. See the examples below.

#### Exploiting the API:
```ts
const required: ValidationRule = (v, msg = 'Required field') => {
    return (Array.isArray(v) ? v.length > 0 : !!v) || msg;
};

// Short way (default message)
// :rules="[ required ]"
// Long way (custom message)
// :rules="[(v) => required(v, 'Some message')]"
```
Now, if a new parameter is introduced(_which is not a breaking change since the user shouldn't exploit the API_), our code can break, which is what you are experiencing right now.

#### A proper, API-compliant, future-proof, and more flexible solution:
```ts
function requiredRule(msg = 'Required field'): ValidationRule {
  return (v) => (Array.isArray(v) ? v.length > 0 : !!v) || msg;
}

// Short way (default message)
// :rules="[ requiredRule() ]"
// (Compared to the exploiting way, it's just two characters longer: with closing and opening parenthesis)
// Long way (custom message)
// :rules="[ requiredRule('custom message') ]"
// (Compared to the exploiting way, it's much shorter and cleaner to use)
```

Another example to prove flexibility and clean usage:
```ts
function minRule(min: number, msg = `Enter a value >= ${min}!`): ValidationRule<number> {
  return (value) => value >= min || msg;
}

// Short way (default message)
// :rules="[ minRule(50) ]"
// Long way (custom message)
// :rules="[ minRule(50, 'custom message') ]"
```
You can also create different kinds of rules by passing in more parameters, or a single object param with different optional properties. It depends on your needs.

Now, let's assume we are using the future-proof solution, and this PR gets released. Nothing will break. If we want to use the new feature(_callable builtin rules_), we can add it, otherwise, everything will continue working properly. We won't need to update `requiredRule`'s implementation or its usage.

Tofandel:
@yusufkandemir Thanks, I like this implementation, it's actually this exact one I was using for vuetify. I'm not sure why I used a different one when I started using quasar, probably for simplicity

For visibility I'd suggest adding a warning in the doc next to this rule if there is a possibility you'll revert (like: warning this feature has caused some issues and there is a chance it will be removed, and maybe example of the alternative usage)

If you want to keep it, I'd suggest a message in the release logs to warn of this breaking possibility and that people should check their rules, because its really easy to miss it and then harder to find why it happened (originally I thought it was caused by a different PR)

And in addition you could force it to break not silently, if the function returns the pattern object, throw an exception. It will be easier to notice something is not acting as expected. 

Or better, treat all objects as falsy, it will break silently but validation will still work, the only unexpected thing will be the error message (which will be even more visible)
The only valid return types being string and boolean (but this might be another breaking change if someone is not casting a truthy return correctly to boolean, its not one if we take your definition) but in return it makes it more resilient in the future

PR #14405, opened 2022-09-15
fix: Validate function not making fields dirty because of lazy-rules

kevpug:
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Any necessary documentation has been added or updated [in the docs]

**Other information:**
A small edge case / bug when using greedy form validation, reactive-rules and lazy-rules together.

When you use greedy validation and validate every input in your form, the inputs that have lazy-rules and reactives-rules but have not been touched (focus) will not react to changes in their rules.

The reason behind this is that when you use the function validate on inputs that have lazy-rules, they will not become dirty and then any changes in their rules(reactive-rules) will not trigger a new validation.

I propose to simply make the input dirty even when using lazy-rules after calling the function validate since error messages would already be shown and makes the focus / unfocus aspect of lazy-rules useless.

How to reproduce the bug / edge case:
1. Go on [https://codesandbox.io/s/cf0q11](url)
2. Press the button to active the greedy validation
3. Change the language with the q-select
4. Observe that only one of the inputs has had change (translation) in it's errors. 
5. In other words, the input that has never been interacted with (Last name) is not dirty and so it's rules will not react even if it has already been validated/error messages are already shown.

My **VERY** small change, in my eyes, has no impact on the way lazy-rules function but if I am wrong by all means correct me and close this PR :).

### DIFF
--- a/ui/src/composables/private/use-validate.js
+++ b/ui/src/composables/private/use-validate.js
@@ -119,7 +119,7 @@ export default function (focused, innerLoading) {
 
     const index = ++validateIndex
 
-    if (innerLoading.value !== true && props.lazyRules !== true) {
+    if (innerLoading.value !== true) {
       isDirtyModel.value = true
     }
 


rstoenescu:
Fixes and enhancements will be available in v2.8.4 and 1.20.2

PR #14407, opened 2022-09-16
fix(QFile): do not check for duplicates in existing list when append is not used #14406

### LINKED ISSUE #14406: QFile multiple select bug
### What happened?

1. Select only 'file1'
2. Select 'file1' and 'file2'
3. Only 'file2' is selected

'file1' and 'file2' selected

### Reproduction URL

https://quasar.dev/vue-components/file-picker#example--multiple-files

### How to reproduce?

1. Select only 'file1'
2. Select 'file1' and 'file2'
3. Only 'file2' is selected

### Flavour

### Areas

Components (quasar)

### Platforms/Browsers

Firefox, Chrome

### Quasar info output

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

pdanpdan:
close #14406

### DIFF
--- a/ui/src/mixins/file.js
+++ b/ui/src/mixins/file.js
@@ -125,7 +125,9 @@ export default {
       })
 
       // Avoid duplicate files
-      const filenameMap = currentFileList.map(entry => entry.__key)
+      const filenameMap = append === true
+        ? currentFileList.map(entry => entry.__key)
+        : []
       files = filterFiles(files, rejectedFiles, 'duplicate', file => {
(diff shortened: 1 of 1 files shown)


rstoenescu:
Perf-wise: should avoid doing any such work instead. Pushed commit for it.

pdanpdan:
~It's not good because now it does not check for duplicate files in the what is added~
Oh, it's not checking for duplicate in current list (not updating the list). Ignore above.

PR #14418, opened 2022-09-16
test(QChip): add unit tests

KareemDa:

- [x] Unit Testing

- [ ] Yes
- [X] No

**Other information:**

Workflow: 
- run `yarn test:components`
- copy output to `QChip.spec.js`
- start writing test cases and and insure that they cover all the scenarios

### DIFF
--- a/ui/src/components/chip/__tests__/QChip.spec.js
+++ b/ui/src/components/chip/__tests__/QChip.spec.js
@@ -1,125 +1,354 @@
+import { mount } from '@cypress/vue'
+import QChip, { defaultSizes } from '../QChip.js'
+
 describe('Chip API', () => {
   describe('Props', () => {
     describe('Category: content', () => {
       describe('(prop): icon', () => {
-        it.skip(' ', () => {
-          //
+        it('should render an icon', () => {
+          const icon = 'add'
+
+          mount(QChip, {
+            propsData: {
+              icon
+            }
+          })
+
+          cy.get('.q-chip')
+            .get('.q-icon')
+            .should('have.text', `${ icon }`)
         })
       })
 
       describe('(prop): icon-right', () => {
-        it.skip(' ', () => {
-          //
+        it('should render an icon on the right', () => {
+          const icon = 'add'
+
+          mount(QChip, {
+            propsData: {
+              iconRight: icon
+            }
+          })
+
+          cy.get('.q-chip')
+            .get('.q-icon.q-chip__icon--right')
+            .should('have.text', `${ icon }`)
         })
       })
 
       describe('(prop): icon-remove', () => {
-        it.skip(' ', () => {
-          //
+        it('should render a remove icon if "removable" is active', () => {
+          const icon = 'delete'
+
+          mount(QChip, {
+            propsData: {
+              removable: true,
+              iconRemove: icon
+            }
+          })
+
+          cy.get('.q-chip')
+            .get('.q-icon.q-chip__icon--remove')
+            .should('have.text', `${ icon }`)
         })
       })
 
       describe('(prop): icon-selected', () => {
-        it.skip(' ', () => {
-          //
+        it('should render a selected icon if is true', () => {
+          const icon = 'done'
+
+          mount(QChip, {
+            propsData: {
+              selected: true,
+              iconSelected: icon
+            }
+          })
+
+          cy.get('.q-chip.q-chip--selected')
+            .get('.q-icon')
+            .should('have.text', `${ icon }`)
         })
       })
 
       describe('(prop): label', () => {
-        it.skip(' ', () => {
-          //
+        it('should render the label inside the chip', () => {
+          const label = 'Chip label'
+
+          mount(QChip, {
+            propsData: {
+              label
+            }
+          })
+
+          cy.get('.q-chip')
+            .get('.q-chip__content')
+            .should('have.text', label)
         })
       })
     })
 
     describe('Category: general', () => {
       describe('(prop): tabindex', () => {
-        it.skip(' ', () => {
-          //
+        it('should set the tabindex', () => {
+          const tabindex = 1
+
+          mount(QChip, {
+            propsData: {
+              clickable: true,
+              tabindex
+            }
+          })
+
+          cy.get('.q-chip')
+            .should('have.attr', 'tabindex', `${ tabindex }`)
         })
       })
     })
 
     describe('Category: model', () => {
       describe('(prop): model-value', () => {
-        it.skip(' ', () => {
-          //
+        it('should be exist when modelValue is true', () => {
+          mount(QChip, {
+            propsData: {
+              modelValue: true
+            }
+          })
+
+          cy.get('.q-chip').should('exist')
+        })
+
+        it('should NOT be exist when modelValue is false', () => {
+          mount(QChip, {
+            propsData: {
+              modelValue: false
+            }
+          })
+
+          cy.get('.q-chip').should('not.exist')
         })
       })
 
       describe('(prop): selected', () => {
-        it.skip(' ', () => {
-          //
+        it('should be selected when selected is true', () => {
+          mount(QChip, {
+            propsData: {
+              selected: true
+            }
+          })
+
+          cy.get('.q-chip.q-chip--selected').should('exist')
+        })
+
+        it('should NOT be selected when selected is false', () => {
+          mount(QChip, {
+            propsData: {
+              selected: false
+            }
+          })
+
+          // q-ship should not have the class "q-chip--selected"
+          cy.get('.q-chip').not('.q-chip--selected').should('exist')
         })
       })
     })
 
     describe('Category: state', () => {
       describe('(prop): clickable', () => {
-        it.skip(' ', () => {
-          //
+        it('should have hover effects and emit click events when clickable is true', () => {
+          const fn = cy.stub()
+
+          mount(QChip, {
+            propsData: {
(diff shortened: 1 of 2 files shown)


ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+  return mount(QChip, options)
+}
+
 describe('Chip API', () => {
   describe('Props', () => {
     describe('Category: content', () => {
       describe('(prop): icon', () => {
-        it.skip(' ', () => {
-          //
+        it('should render an icon', () => {
```
IlCallo: This will pass even when using `icon-right`
I guess we should change this to `should render an icon on the left` and check for `q-chip__icon--left`

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+          cy.get('.q-chip')
+            .get('.q-icon.q-chip__icon--right')
+            .should('have.text', `${ icon }`)
         })
       })
 
       describe('(prop): icon-remove', () => {
-        it.skip(' ', () => {
-          //
+        it('should render a remove icon if "removable" is active', () => {
```
IlCallo: This test title should be something along `should render a custom remove icon` instead, since you're already testing `removable` prop at line 183

I'd also stick to the `it should ... [when ... [and ...]]` test naming convention instead of using `it should ... [if ... [and ...]]`, since the former is most used

Apparently we've been inconstistent with this format up until now, but `when` is the most used form right now in the codebase anyway

Do you want to fix it in this PR? Otherwise we can fix it in a later PR and add a note on the testing README for newcomers

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+          cy.get('.q-chip')
+            .get('.q-icon.q-chip__icon--remove')
+            .should('have.text', `${ icon }`)
         })
       })
 
       describe('(prop): icon-selected', () => {
-        it.skip(' ', () => {
-          //
+        it('should render a selected icon if is true', () => {
```
IlCallo: Same about the `remove-icon` test, here the focus is the ability to change the icon, not the `selectable` property
I'd suggest `should render a custom selected icon when one provided`, optionally you could add `... and selected prop is true`, but that feels be a bit too verbose

Unless you also add the un-happy path where someone uses `selected-icon` without `selected` being true, but I don't think we need that degree of coverage right now

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+          cy.get('.q-chip').should('have.attr', 'tabindex', `${ tabindex }`)
         })
       })
     })
 
     describe('Category: model', () => {
       describe('(prop): model-value', () => {
-        it.skip(' ', () => {
-          //
+        it('should be exist when modelValue is true', () => {
```
IlCallo: I guess this should be `should render when modelValue is true` or `should exist when modelValue is true`

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+          mountQChip({
+            props: {
+              modelValue: true
+            }
+          })
+
+          cy.get('.q-chip').should('exist')
+        })
+
+        it('should NOT be exist when modelValue is false', () => {
```
IlCallo: `should not render ...` or `should not exist ...`

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+          })
+
+          cy.get('.q-chip').should('not.exist')
         })
       })
 
       describe('(prop): selected', () => {
-        it.skip(' ', () => {
-          //
+        it('should be selected when selected is true', () => {
```
IlCallo: In these cases where there are clashing names I'd specify `when selected prop is true`, or be more generic and go with `when the appropriate prop is set to true`

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+        })
+
+        it('should NOT be selected when selected is false', () => {
+          mountQChip({
+            props: {
+              selected: false
+            }
+          })
+
+          // q-ship should not have the class "q-chip--selected"
```
IlCallo: I'd expect [this syntax](https://docs.cypress.io/guides/references/assertions#Class) here, as it better convey the message without the need for a comment

Same for the immediately preceding test

IlCallo: `cy.get('.q-chip').should('not.have.class', 'q-chip--selected')`

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+            .should('have.css', 'cursor', 'pointer')
+            .click()
+            .then(() => expect(fn).to.be.calledOnce)
         })
       })
 
       describe('(prop): removable', () => {
-        it.skip(' ', () => {
-          //
+        it('should displays a "remove" icon that when clicked the QChip emits "remove" event', () => {
```
IlCallo: "should display a "remove" icon emitting a "remove" event when clicked" *

IlCallo: In general, give a second round to all test names, then I'll fix the ones with can be better rephrased

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+            props: {
+              size
+            }
+          })
+
+          cy.get('.q-chip').should('have.css', 'font-size', size)
+        })
+
+        it('should change QChip size based defined values: xs, sm, md, lg, xl', () => {
+          const sizes = [ 'xs', 'sm', 'md', 'lg', 'xl' ]
```
IlCallo: Can't you obtain these as `Object.keys(defaultSizes)`? And possibly expand them as `Object.keys(defaultSizes).join(', ')` into the test title too

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+
+          sizes.forEach((size) => {
```
IlCallo: Instead of mounting a new one each time, mount it once, then use the component global reference to change the prop, so you can avoid removing the component for each cycle

Also, rely on plain `for` when possible (`for of` in this case)

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
-        it.skip(' ', () => {
-          //
+        it('should have a square style when square is true', () => {
+          mountQChip({
+            props: {
+              square: true
+            }
+          })
+
+          cy.get('.q-chip.q-chip--square').should('exist')
```
IlCallo: I'd expect you to use `have.class` here too, and in previous/following similar tests where you're checking for classes on the base `.q-chip` element

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+        cy.get('.q-chip').should('have.text', 'Default Slot Content')
       })
     })
   })
 
   describe('Events', () => {
     describe('(event): click', () => {
-      it.skip(' ', () => {
-        //
+      it('should emit click event when clicked and "clickable" is set', () => {
```
IlCallo: `and "clickable" prop is set`*

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+            'onUpdate:selected': fn
+          }
+        })
+
+        cy.get('.q-chip')
+          .click()
+          .then(() => expect(fn).to.be.calledOnce)
+      })
+
+      it('should NOT emit update:selected event selected is NOT applied', () => {
```
IlCallo: `when "selected" prop is NOT set`*

ui/src/components/chip/__tests__/QChip.spec.js
```diff
@@ -1,154 +1,467 @@
...
+        cy.get('.q-chip')
+          .click()
+          .then(() => expect(fn).to.not.be.called)
       })
     })
 
     describe('(event): remove', () => {
-      it.skip(' ', () => {
-        //
+      it('should emit remove event when clicked and removable is applied', () => {
```
IlCallo: Sometimes double quotes are used around props names, sometimes not, I'd make it consistent, use them and add "prop" where needed
I'd also document this into the README

IlCallo (changes requested):

ui/src/components/chip/__tests__/QChip.spec.js
KareemDa: I'll resolve it in this PR (but if you can update the README, this would be great !)

rstoenescu:
Thanks for contributing!

KareemDa:
> Thanks for contributing!

It's my pleasure! this is just the beginning 😃. I'll do my best to finish as much testing as I can. 

thank you

rstoenescu:
Looking forward to it!

PR #14575, opened 2022-10-07
feat(a11y): improve a11y for components

pdanpdan:
- QBadge: role status instead of alert
- QDialog: role dialog
- QBottomSheet: role list/listitem
- QItem: role listitem
- QList: role list
- QOptionGroup: role group if not radiogroup
- QPagination: role navigation; aria-label; aria-current
- QToolbar: role toolbar
- QTooltip: role tooltip instead of complementary
- QToggle: role switch


rstoenescu:
Go for v2 also so I can push both of them together.

Issue #14577, opened 2022-10-07
QPagination text color doesn't work


tomukasas:
### What happened?

As title says

### What did you expect to happen?

Text color to work

### Reproduction URL

https://stackblitz.com/edit/quasarframework-webpack-aun8hn?file=src/pages/IndexPage.vue

### How to reproduce?

1. Go to reproduction link
2. Notice the text color is blue, while it should be black based on `text-color="black"` prop

### Flavour

Quasar CLI with Webpack (@quasar/cli | @quasar/app-webpack)

### Areas

Components (quasar)

### Platforms/Browsers

Firefox, Chrome

pdanpdan:
I think you want this: https://codepen.io/pdanpdan/pen/wvjEbbw?editors=101

@rstoenescu I think `text-color` is useless (not used) for QPagination and we should remove it

tomukasas:
`color` prop is used for background color on other components, so from consistency point of view it would be wise to keep that naming pattern and set the background color, not the text color.

pdanpdan:
All buttons that are not active are flat
And for active there is activeColor and activeTextColor

tomukasas:
`active-text-color` works fine :) the bug is about `text-color`

pdanpdan:
In the implementation both do the exact same thing. If you don't set activeTextColor then textColor will work :)

tomukasas:
And what should I do if I need both? :)

I don't mind adding some custom css to override the look and feel to how I want, just trying to save couple code lines :D

pdanpdan:
That was the question. What do you want to do?

tomukasas:
To set correct colors for buttons :)

How I imagine it should work is - `text-color` should be set for all buttons, `active-text-color` should be set only on the active button. Hope that makes sense.

pdanpdan:
Isn't it what I posted first?
https://codepen.io/pdanpdan/pen/wvjEbbw?editors=101

Flat buttons only have one color (color)

tomukasas:
Right, What if I want red background color and white text? And for active button I want white background and black text?

pdanpdan:
As I said, all not active buttons are flat, so no bg

tomukasas:
Ok, got it, thanks :)

rstoenescu:
@tomukasas 

Thank you for opening this ticket. There were lots of issues with the current implementation in regards to the design.
[...]

PR #14617, opened 2022-10-13
refactor+feat+fix+perf(QBtn/QBtnDropdown/QBtnGroup/QPagination): (backport from Qv2) "square" prop; fix button design props; new props (gutter & active-design); enhance runtime speed #14577 #13912

pdanpdan:


rstoenescu:
The "gutter" prop that I wrote - not ok for Qv1.
The cache() on QPagination makes no sense as it is written now.

Fixing

pdanpdan:
ref caching of inputEvents you are right, sorry :(
but being vue2 it still needs `on: { ...this.inputEvents }`

Issue #14491, opened 2022-09-26
The dialog will be rendered 3 times.


oasis345:
### What happened?

The dialog will be rendered 3 times.
![sample](https://user-images.githubusercontent.com/51695127/192192323-ce313794-83ed-4cd6-aeef-60b785aa6da3.JPG)

### What did you expect to happen?

Dialog should be rendered only once

### Reproduction URL

https://codepen.io/pen?&editors=101

### How to reproduce?

call a function on the template like uploaded image.

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Quasar CLI Commands/Configuration (@quasar/cli | @quasar/app-webpack | @quasar/app-vite), Components (quasar)

### Platforms/Browsers

Chrome

pdanpdan:
A working reproduction: https://codepen.io/pdanpdan/pen/XWqzYgW?editors=1010

rstoenescu:
Enhancement will be available in Quasar v2.10.1 and v1.22.1

PR #14637, opened 2022-10-17
feat(transitions): do not change transition name when showing/hiding - prevent re-render #14491

pdanpdan:


rstoenescu:
Will do some tweaks and remove the extra unused classes

PR #14687, opened 2022-10-22
fix: move meta charset="utf8" to start of head (fixes #6261)

### LINKED ISSUE #6261: In ssr mode, added meta tags through meta plugin result in incorrect position of charset meta tag
**Describe the bug**
`<meta charset="utf-8">` should be placed at the beginning of the head section.
Quote from MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#attr-charset

> The <meta> element declaring the encoding must be inside the <head> element and within the first 1024 bytes of the HTML as some browsers only look at those bytes before choosing an encoding.

When adding meta tags through meta plugin in ssr mode, the added tags are placed at the top of the head section, resulting in pushing down the charset tag

https://codesandbox.io/s/busy-vaughan-7eyfx

The issue can also be seen in quasar website https://quasar.dev/
Screenshot from 2020-01-30 11-36-46

ssendev:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No?

Technically this moves the charset meta tag to the top of the template which changes the build output so might be considered a breaking change if that is the case it would be possible to add a flag to `quasar.conf` to restore the original behavior.

**Other information:**

The original issue mentions performance reasons for not fixing it but this approach just adds a slightly more complicated regex in dev and no overhead in prod

### DIFF
--- a/app-vite/lib/helpers/html-template.js
+++ b/app-vite/lib/helpers/html-template.js
@@ -47,8 +47,8 @@ function injectRuntimeInterpolation (html) {
     }
   )
   .replace(
-    /(<head[^>]*)(>)/i,
-    (_, start, end) => `${start}${end}{{ _meta.headTags }}`
+    /(<head[^>]*)(>)(?:([^]*)(<meta[^>]*charset\s*=\s*['"]utf-8['"][^>]*>)|)/i,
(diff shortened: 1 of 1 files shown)


rstoenescu:

Issue #11683, opened 2021-12-14
Sass deprecation warnings emitted during build


p3v9d5ui:
**Describe the bug**

During build (Vite 2.7.1), several deprecation warnings (Dart Sass) are emitted. The warning is:

    Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

**To Reproduce**

Start a Vue3.2.24/Quasar2.3.4/Vite@2.7.1 project using "npm run dev"

**Expected behavior**

No warnings should be emitted. However, these warnings are emitted:

```
Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($angle, 180) or calc($angle / 180)

More info and automated migrator: https://sass-lang.com/d/slash-div

   ╷
29 │   $rad: $angle / 180 * $PI
   │         ^^^^^^^^^^^^
   ╵
    node_modules/quasar/src/css/helpers/math.sass 29:9         sin()
    node_modules/quasar/src/components/time/QTime.sass 128:16  @import
    node_modules/quasar/src/css/index.sass 71:9                root stylesheet

Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div(pow(-1, $i) * pow($rad, (2 * $i + 1)), fact(2 * $i + 1)) or calc(pow(-1, $i) * pow($rad, (2 * $i + 1)) / fact(2 * $i + 1))

More info and automated migrator: https://sass-lang.com/d/slash-div

   ╷
32 │     $sin: $sin + pow(-1, $i) * pow($rad, (2 * $i + 1)) / fact(2 * $i + 1)
   │                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
    node_modules/quasar/src/css/helpers/math.sass 32:18        sin()
    node_modules/quasar/src/components/time/QTime.sass 128:16  @import
    node_modules/quasar/src/css/index.sass 71:9                root stylesheet

Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($angle, 180) or calc($angle / 180)

More info and automated migrator: https://sass-lang.com/d/slash-div

   ╷
38 │   $rad: $angle / 180 * $PI
[...]

Issue #10427, opened 2021-08-19
[sass] replace / with  math.div()


fyeeme:
╷
43 │     $sin: $sin + pow(-1, $i) * pow($rad, (2 * $i + 1)) / fact(2 * $i + 1)
   │                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
    node_modules/quasar/dist/quasar.sass 43:18    sin()
    node_modules/quasar/dist/quasar.sass 4405:16  root stylesheet

DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

pdanpdan:
duplicate of #9362

PR #12034, opened 2022-01-14
migrate to sass 1.48 fix: #9362,#10427,#11683

bompus:
**What kind of change does this PR introduce?** 

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** 

- [ ] Yes
- [x] No

**The PR fulfills these requirements:**

- [x] It's submitted to the `dev` branch (or `v[X]` branch)
- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)
- [ ] It's been tested on a Cordova (iOS, Android) app
- [ ] It's been tested on a Electron app
- [x] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs)  or explained in the PR's description.

If adding a **new feature**, the PR's description includes:
- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to [start a new feature discussion](https://github.com/quasarframework/quasar/discussions/new?category=ideas-proposals) first and wait for approval before working on it)

**Other information:**

Math divisions are now wrapped in native css `calc()`, which does not require implementing `@use`.

This resolves "Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0."

Another PR #11877 attempted to solve this, but missed `flex-addon.sass`, and used the non-desired `@use 'sass:math'` method.

fix #9362, fix #10427, fix #11683

PR #14816, opened 2022-11-04
fix: implement divide function using multiplication (fix #9362, #10427, #11683)

### LINKED ISSUE #9362: Deprecation warning with new dart-sass
**Describe the bug**
https://github.com/sass/dart-sass/releases/tag/1.33.0
After updating dart-sass I get a lot of deprecation warnings:

```
DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div(round($number * $power), $power)

More info and automated migrator: https://sass-lang.com/d/slash-div

   ╷
36 │   @return round($number * $power) / $power
   │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
    node_modules\.pnpm\quasar@1.15.15_vue@2.6.12\node_modules\quasar\dist\quasar.sass 36:11    toFixed()
    node_modules\.pnpm\quasar@1.15.15_vue@2.6.12\node_modules\quasar\dist\quasar.sass 6639:19  fg()
    node_modules\.pnpm\quasar@1.15.15_vue@2.6.12\node_modules\quasar\dist\quasar.sass 6746:3   @import
```

**Additional context**
It is not something that needs to be fixed right away as replacement function `math.div` is only introduced in latest version of `dart-sass`.

bgoscinski:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

I know that fixing this was attempted multiple times already but I didn't see any PR implementing this approach. It's quite popular to solve it this way ([bootstrap](https://github.com/twbs/bootstrap/blob/main/scss/_functions.scss#L262-L302), [bulma](https://github.com/jgthms/bulma/blob/master/sass/utilities/functions.sass#L116-L135), [foundation](https://github.com/foundation/foundation-sites/blob/develop/scss/util/_math.scss#L149-L197)) so I decided to give it a shot.

I think that with this merged the requirement of pining sass version could be dropped.

Fixes #9362, #10427, #11683

### DIFF
--- a/ui/src/css/core/flex.sass
+++ b/ui/src/css/core/flex.sass
@@ -30,13 +30,13 @@
       .row
         #{str-fe('> .col<name>-<i>', $name, $noProcNotZero, $ic)}
           height: auto
-          width: toFixed(percentage($i / $flex-cols), 10000)
+          width: toFixed(percentage(divide($i, $flex-cols)), 10000)
         @if $i != 0 or $name != ''
           #{str-fe('> .offset<name>-<i>', $name, $noProcNotZero, $ic)}
-            margin-left: toFixed(percentage($i / $flex-cols), 10000)
+            margin-left: toFixed(percentage(divide($i, $flex-cols)), 10000)
       .column
         #{str-fe('> .col<name>-<i>', $name, $noProcNotZero, $ic)}
-          height: toFixed(percentage($i / $flex-cols), 10000)
+          height: toFixed(percentage(divide($i, $flex-cols)), 10000)
           width: auto
       @if $size == 0 and $i == $flex-cols
         .row > .col-all
--- a/ui/src/css/flex-addon.sass
+++ b/ui/src/css/flex-addon.sass
@@ -194,7 +194,7 @@
 
         @for $i from 0 through $flex-cols
           $ic: quote('' + $i)
-          $internal-size: toFixed(percentage($i / $flex-cols), 10000)
+          $internal-size: toFixed(percentage(divide($i, $flex-cols)), 10000)
 
           #{str-fr('.row<name>', $name)}
             #{str-fe('> .col<name>-<i>', $name2c, $noProcNotZero2, $ic)}
--- a/ui/src/css/helpers/math.sass
+++ b/ui/src/css/helpers/math.sass
@@ -1,5 +1,38 @@
 $PI: 3.14159265359
 
+// Implements division via multiplication to avoid deprecation warnings in
+// sass@>1.32.12.
+// Source: https://github.com/twbs/bootstrap/blob/2b21094074b1bdabf9ebc4095c29398f8b2de237/scss/_functions.scss
+@function divide($dividend, $divisor, $precision: 10)
+  $sign: if($dividend > 0 and $divisor > 0 or $dividend < 0 and $divisor < 0, 1, -1)
+  $dividend: abs($dividend)
+  $divisor: abs($divisor)
+  @if $dividend == 0
+    @return 0
+  @if $divisor == 0
+    @error "Cannot divide by 0"
+  $remainder: $dividend
+  $result: 0
+  $factor: 10
+  @while ($remainder > 0 and $precision >= 0)
+    $quotient: 0
+    @while ($remainder >= $divisor)
+      $remainder: $remainder - $divisor
+      $quotient: $quotient + 1
+    $result: $result * 10 + $quotient
+    $factor: $factor * .1
+    $remainder: $remainder * 10
+    $precision: $precision - 1
+    @if ($precision < 0 and $remainder >= $divisor * 5)
+      $result: $result + 1
+  $result: $result * $factor * $sign
+  $dividend-unit: unit($dividend)
+  $divisor-unit: unit($divisor)
+  $unit-map: ("px": 1px, "rem": 1rem, "em": 1em, "%": 1%)
+  @if ($dividend-unit != $divisor-unit and map-has-key($unit-map, $dividend-unit))
+    $result: $result * map-get($unit-map, $dividend-unit)
+  @return $result
+
 @function pow($number, $exp)
   $value: 1
   @if $exp > 0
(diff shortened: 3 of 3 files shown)


bgoscinski:
I noticed that `yarn build css` perf degraded considerably after replacing native `/` with `divide` function (from ~6.2s to ~43.3s on my PC). I discovered that `sin` and `cos` functions divide huge numbers which is slow with this `divide` implementation. Fortunately I was able to optimize `sin` and `cos` by doing `$angle % 360` which "trims" irrelevant full-turns from the angle before all the divisions are executed. This is done in [this commit](https://github.com/quasarframework/quasar/pull/14816/commits/be190fd4e1c839454bf3040cb676316a66a7d6b9).

The timings I got:

Baseline - before my changes ([fd616b580](https://github.com/quasarframework/quasar/commit/fd616b58026d3a5caaef8ae43ea65bae6cac6c02))
```
Benchmark 1: yarn build css
  Time (mean ± σ):      6.221 s ±  0.064 s    [User: 8.781 s, System: 0.919 s]
  Range (min … max):    6.140 s …  6.363 s    10 runs
```

With new `divide` implementation ([09d6098e0](https://github.com/quasarframework/quasar/pull/14816/commits/09d6098e05350fa940e57bad82fc9413035a6217))
```
Benchmark 1: yarn build css
  Time (mean ± σ):     43.315 s ±  0.496 s    [User: 45.887 s, System: 0.966 s]
  Range (min … max):   42.367 s … 43.879 s    10 runs
```

With new `divide` implementation and `sin` and `cos` optimized ([be190fd4e](https://github.com/quasarframework/quasar/pull/14816/commits/be190fd4e1c839454bf3040cb676316a66a7d6b9))
```
Benchmark 1: yarn build css
  Time (mean ± σ):      6.438 s ±  0.048 s    [User: 9.037 s, System: 0.835 s]
  Range (min … max):    6.353 s …  6.525 s    10 runs
```

andrews05:
Is it possible to detect sass version and make the `divide` function just use `calc` when available?

bgoscinski:
@andrews05 Unfortunately this is not possible because `calc` is a CSS function (it evaluates at runtime). 

We need the result of the division to be available for SASS expressions (at compile time) such as here:
https://github.com/quasarframework/quasar/blob/be190fd4e1c839454bf3040cb676316a66a7d6b9/ui/src/css/helpers/math.sass#L64-L65

andrews05:
I mean the sass calc, as seen in earlier PRs such as https://github.com/quasarframework/quasar/pull/12034#issuecomment-1021129405
This is only possible in sass 1.40, but I’m not sure if it’s possible to do version/feature detection in sass.

MilosPaunovic:
Hi, thanks for your contribution!

Unfortunately, I don't see this being merged; for more details check this out: https://github.com/quasarframework/quasar/pull/14213#issuecomment-1219170007

bgoscinski:
@MilosPaunovic I saw that. This approach is different though - it doesn't use the `@use` feature and will work the same regardless of sass version/flavor so I think it deserves a fresh look 🙂

bgoscinski:
> I mean the sass calc, as seen in earlier PRs such as [#12034 (comment)](https://github.com/quasarframework/quasar/pull/12034#issuecomment-1021129405) This is only possible in sass 1.40, but I’m not sure if it’s possible to do version/feature detection in sass.

Oh I didn't know about sass trying to evaluate calc. Interesting idea. I'll experiment with this next week. Thanks! 
 
Either way I'd like to first hear what quasar maintainers think about the current approach implemented in this PR.

rstoenescu:
Hi,

First of all, thanks for contributing!

However, this has been discussed a lot of times and it currently doesn't makes sense to use any other Sass version than the recommended 1.32.12 (info is in docs). The proper way to support Sass 1.32.13+ involves making a breaking change (which is out of the question for current Quasar v2) in the way that variables are used, among other areas as well. Support for the newer Sass will come along with Quasar v3. This is just the way things can work at the moment due to the fact that the newer Sass decided to do substantial (breaking/deprecating) changes to its code in a minor version.

andrews05:
@rstoenescu Do you think there would be any harm in the user using a newer version and just setting the `quietDeps` option?

bgoscinski:
@rstoenescu Hi, thanks for looking into this. 

In this PR I didn't introduce any `@use` directive, nor any feature that the newer sass versions have and old don't so I think it's both **backward** and **forward** compatible. 

With this changes in place quasar styles build without any deprecations which made me think that now quasar uses non-deprecated features **exclusively**. I may have misinterpreted what you said but the way I read it you suggested that's not the case - that some more changes (breaking) are needed to _unblock_ newer sass versions (`<2.0.0`). I searched other closed issues and PRs and failed to find info on what you meant specifically. Do you mean `@import` syntax? 

If your concern is about `@import` there's [this sass timeline](https://github.com/sass/sass/blob/main/accepted/module-system.md#timeline) that shows that `@import` doesn't go anywhere for at least a year and is not yet deprecated (doesn't produce the warning). With this in mind I don't see a reason to keep users on `sass@1.32.12`.

Or did you mean some other problematic feature/syntax/etc?

I'm also interested in answer to what @andrews05 asked above. 

Thanks for your time and the stellar work you're doing here!

andrews05:
Personally, I do feel something needs to be done about this. The position that it will be resolved in v3 isn't particular helpful given that v2 is still relatively young and v3 isn't even on any roadmap yet. Surely we can't stay at this impasse indefinitely - there *are* solutions to the problem, and the one here seems to meet all the compatibility requirements. I'm not suggesting changing Quasar's own sass version or the recommendations on the website, but simply implementing this fix to allow users to use a newer sass version if they so choose.

That said, I am currently using the `quietDeps` option to silence the warnings. Here's how to do it with Vite:
```js
export default defineConfig({
  css: {
    preprocessorOptions: {
      scss: {
        quietDeps: true,
      },
    },
  },
})
```
Then instead of importing the quasar css directly in your `main.js` file you need to import it from within one of your own css files.

PR #14830, opened 2022-11-06
fix(VitePlugin): prevent duplicate imports in dev mode (useTreeshaking false) when a component is imported in code and auto-imported from template

pdanpdan:
```
<template>
  <q-btn />
</template

<script>
import { QBtn } from 'quasar'
</script>
```


rstoenescu:
Not ok. Not covering all use-cases.
Pushing a commit for this specifically -- and added some test-cases there.

PR #15129, opened 2022-12-16
feat(ui): Improve QInput and QFile nativeEl types (fix #15128)

### LINKED ISSUE #15128: Improve QInput.nativeEl and QFile.nativeEl types by narrowing
### Discussed in https://github.com/quasarframework/quasar/discussions/15121

<sup>Originally posted by **mv-go** December 16, 2022</sup>
According to [docs](https://quasar.dev/vue-components/input#qinput-api) the type of `nativeEl` and the return type of `getNativeElement` is
```The native input/textarea DOM Element```

However, the typings that shipped with the library suggest that the type here is a very general `Element`.

Could it be possible to narrow it down to `HTMLInputElement | HTMLTextAreaElement`? Or are there some edge cases that I'm missing?</div>

### PR DESCRIPTION
**Pull request by yusufkandemir**

- Feature

- No

**Other information:**
Resolves #15128

Ships two new types: `QFileNativeElement` and `QInputNativeElement`

`QInputNativeElement` accepts a generic parameter:
```ts
QInputNativeElement                      // HTMLInputElement | HTMLTextAreaElement
QInputNativeElement<"textarea">          // HTMLTextAreaElement
QInputNativeElement<"text">              // Omit<HTMLInputElement, "type"> & { type: "text" }
QInputNativeElement<"text" | "number">   // Omit<HTMLInputElement, "type"> & { type: "text" | "number" }
QInputNativeElement<"text" | "textarea"> // (Omit<HTMLInputElement, "type"> & { type: "text" }) | HTMLTextAreaElement
```
We can't make use of the generic param for `QInput.nativeEl` because we don't have generic component types yet(_not fully complete in Vue in a straightforward either_). But, users can use this type for advanced/flexible use cases by casting their types, typing their helper functions parameters, etc. Also, further narrowing is possible by checking against the `type` prop:
```ts
const inputRef = ref() as Ref<QInput>;
// ...
const { nativeEl } = inputRef.value
if (nativeEl.type === 'textarea') {
  nativeEl // HTMLTextAreaElement
} else if (nativeEl.type === 'files') {
  nativeEl.files // FileList
} else {
  nativeEl.files // null
}
```

### DIFF
--- a/ui/src/components/file/QFile.json
+++ b/ui/src/components/file/QFile.json
@@ -168,6 +168,7 @@
       "desc": "DEPRECATED; Access 'nativeEl' directly; Gets the native input DOM Element",
       "returns": {
         "type": "Element",
+        "tsType": "QFileNativeElement",
         "desc": "The underlying native input DOM Element"
       }
     }
@@ -176,6 +177,7 @@
   "computedProps": {
     "nativeEl": {
       "type": "Element",
+      "tsType": "QFileNativeElement",
       "desc": "The native input DOM Element",
       "addedIn": "v2.10.1"
     }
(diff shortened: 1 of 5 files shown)


rstoenescu:
Nice!

PR #15150, opened 2022-12-20
Add Additional Text Color Configuration Options

PrinceManfred:

- [ ] Bugfix
- [x] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
While implementing a dark mode theme I ran across an issue where the light mode `secondary` color we selected looked good on a `q-btn` with the default white text, but the dark mode `secondary` color isn't readable with white text. The issue can obviously be solved by providing a `text-color` value, but it's easier to ensure that my team is consitent with styles if the text color selection follows from the background color selection.

This pull request adds the following SASS variables: `$on-default`, `$on-primary`, `$on-secondary`, `$on-accent`, `$on-positive`, `$on-negative`, `$on-info`, and `$on-warning`, plus `$on-[all color variants]`. These new variables represent the preferred text color for text appearing on backgrounds colored with their existing color counterparts. By default they are set to `$on-default` which has a value on '#FFF`.

It also adds `--q-on-primary`, `--q-on-secondary`, `--q-on-accent`, `--q-on-positive`, `--q-on-negative`, `--q-on-info`, and `--q-on-warning` as css properties on the root element to expose these values and allow for changes at runtime.

Finally, it adds `text-on-[color]` classes for all defined colors. All components that accept a `color` and `text-color` prop will default the text class to `text-on-[color]` if a `color` is provided but no `text-color`. Anytime a `text-color` is supplied it is used instead.

My only uncertainty is how to update the brand configuration in quasar.config.js to optionally accept the new color options.
Default values shouldn't cause any visual difference in the vast majority of existing sites. The biggest possible break would be name collisions if users are using any of the new variable, property, or class names.


rstoenescu:

PR #15206, opened 2022-12-31
Update dom-utils.md

Daemach:
This is a documentation update, providing a tip on retrieving a dom element for use in the 'utils.dom' methods.


rstoenescu:
Hi,

Thanks for contributing!
However, the information is really implicit, so not worth it for docs. And if we supply a list of methods, we should supply them all -- otherwise we will confuse the newbies and pros alike.

PR #15325, opened 2023-01-25
fix(ui/ts): allow data-* attributes in dialog plugin prompt/options

IlCallo:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Loosen up the type a bit to allow for attributes which aren't well-known
I also noticed we let in the autocomplete some properties which shouldn't be there, like `datalist`, but I'm not sure if we're able to filter them out automatically or we would need to remove them one by one

### DIFF
--- a/ui/types/api/dialog.d.ts
+++ b/ui/types/api/dialog.d.ts
@@ -1,12 +1,7 @@
 // Error on "quasar" import shown in IDE is normal, as we only have Components/Directives/Plugins types after the build step
 // The import will work correctly at runtime
 import { QInputProps, QOptionGroupProps } from "quasar";
-
-// Taken from https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#distributive-conditional-types
-type NonFunctionPropertyNames<T> = {
-  [K in keyof T]: T[K] extends Function ? never : K;
-}[keyof T];
-type NonFunctionProperties<T> = Pick<T, NonFunctionPropertyNames<T>>;
+import { InputHTMLAttributes, HTMLAttributes } from "vue";
 
 // https://html.spec.whatwg.org/multipage/input.html#attr-input-type
 type InputType =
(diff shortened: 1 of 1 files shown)


ui/types/api/dialog.d.ts
```diff
@@ -56,9 +51,14 @@ export type QDialogInputPrompt<T = string> = {
    */
   isValid?: (value: T) => boolean;
 } & Partial<Omit<QInputProps, "modelValue" | "onUpdate:modelValue" | "type">> &
-  Partial<NonFunctionProperties<HTMLInputElement>>;
+  Omit<InputHTMLAttributes, "type"> & {
+    // Has type-checking but no autocompletion: https://github.com/microsoft/TypeScript/issues/41620
+    [dataAttributes in `data-${string}`]?: any;
+  } & {
+    "data-cy"?: string;
+  };
```
yusufkandemir: Accepting any attribute was too broad, could lead to typos and such. Aside from the element-specific attributes, common attributes like `aria-*` was already being handled. So, pretty much the other thing left is `data-*`. Because of that, I converted it to only allow `data-*`, then on top of that added `data-cy` specifically for ease of access via auto-completion.

ui/types/api/dialog.d.ts
```diff
@@ -1,12 +1,7 @@
 // Error on "quasar" import shown in IDE is normal, as we only have Components/Directives/Plugins types after the build step
 // The import will work correctly at runtime
 import { QInputProps, QOptionGroupProps } from "quasar";
-
-// Taken from https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#distributive-conditional-types
-type NonFunctionPropertyNames<T> = {
-  [K in keyof T]: T[K] extends Function ? never : K;
-}[keyof T];
-type NonFunctionProperties<T> = Pick<T, NonFunctionPropertyNames<T>>;
+import { InputHTMLAttributes, HTMLAttributes } from "vue";
```
yusufkandemir, line 4: After your comment on extra stuff like `dataset` slipping in, I dug through and decided the best way is to use Vue's JSX typings. It only includes properties that are actually fillable via JSX/h(). Not all stuff is perfect, so some stuff like the specific input types that I made is still helpful and used. We can use types like these in more components as the time goes to improve the overall experience.

PR #15379, opened 2023-02-03
ci: add types build checking and results (fix: #15372)

### LINKED ISSUE #15372: Check formatting and TS validity of the UI types build
Currently, the formatting of `ui/types` is not unified. There is no command to format or a mechanism to check if the formatting is valid. We are implicitly following the Prettier's default formatting. So, we should unify and automate this.

The JSON API and types build process can break easily. So, we either notice it after merging PRs which seemed fine at first, or have always to check out the PR and test the build, even for very simple changes.
Moreover, some changes don't break the build but produce invalid TS code. So, they might be importing non-existing types due to typos, lack of understanding of the JSON API system, etc. The bad thing is that there is no existing mechanism to detect these errors. It's completely manual now. So, we should automate this checking too.

After handling both cases, we should create a GitHub Actions workflow that runs on PRs that modify JSON API files or `.d.ts` files, then check the formatting and if the types build complete successfully by producing valid TS code.
On top of that, if we can automatically create a comment with the types diff that we already outputting via the build process, it would be great. That way we can also see if it generates the expected result, not just any valid TS code.

yusufkandemir:

- Code style update
- Build-related changes
- CI changes

- No

**Other information:**
Resolves #15372.

I improved the manual authoring `.d.ts` files experience by quite a lot by correctly mapping `quasar` to the build. So, we'll no longer have errors and zero inference when importing auto-generated types from `quasar` into manually authored types.

`yarn build js types` now checks the validity of the final TS code. So, it ensures the manually authored `.d.ts` in the `types` directory, JSON API definitions, and the related build scripts result in a valid final output. To see how it works, you can change a `tsType` value that doesn't have `autoDefineTstype` and is not an API-breaking value(_API-breaking errors were already being caught_). Example: `"tsType": "QEditorCaret"` -> `"tsType": "QEditorCaretasd"`.

We now have proper and unified formatting on the manually authored `.d.ts` files.

We have two new workflows:
- Build types: Runs on push and pull requests if types, JSON API files, types build script, etc. get modified. It ensures the files are formatted correctly and the build is passing. Then, it uploads the related artifacts. For the push event, these artifacts will be used as a reference for future PRs to diff the changes. For pull requests, the artifact also includes a file containing the PR number, to be used in the other workflow later.
- Comment build results: Gets run as soon as the "Build types" workflow completes successfully. If there is no current artifact on the dev branch to take as a reference, it ends here. This is a quite rare case, but whatever. Otherwise, it downloads the current artifact, then the artifact from the PR. Then, it diffs both artifacts, then comments the results to the PR.

As these workflows depend on a lot of different things, as far as I know, [nektos/act](https://github.com/nektos/act) is pretty much limited when testing this. So, to be perfectly honest, I am not 100% certain that this would work perfectly on the first real-life try. I know for a fact that "Build types" will work fine, but not completely sure about the other one.

We can add a similar type of check and build summary for other types of builds(_JS, CSS, icon sets, etc._) later. This should be a good start and serve as a foundation for future ones.


.github/workflows/build-types.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      # Manually authored .d.ts files
+      - 'ui/types/**'
+      # JSON API files
+      - 'ui/src/**/*.json'
+      - '!ui/src/api-file-example.json'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
IlCallo, line 40: I guess this stops a workflow when a new workflow on the same PR is run, right?
E.g. I push and push again right after, the first workflow will be stopped in favor of the second

.github/workflows/build-types.yml
```diff
@@ -0,0 +1,84 @@
...
+          # No caching because we are not checking lock files in
+
+      - name: Install dependencies
+        run: yarn
+
+      - name: Ensure types are formatted correctly
+        run: yarn format:types --write=false --check
+
+      - name: Build the types
+        run: yarn build js types
```
IlCallo, line 69: Won't this run user-defined malicious code, if an user sends a PR which updates `ui/build/build.types.js` or `ui/build/build.javascript.js`? I mean, will users be able to open multiple PRs with the same change (or with minimal differences) in which each workflow flood with requests a target endpoint DDoSing it?

If yes, we should add the same "only run after a maintainer checks for malicious code" logic we should add for tests too

ui/build/build.types.js
```diff
@@ -561,25 +566,75 @@ function writeIndexDTS (apis) {
   writeLine(contents, 'import \'./shim-icon-set\'')
   writeLine(contents, 'import \'./shim-lang\'')
 
-  writeFile(
-    resolvePath('index.d.ts'),
-    prettier.format(contents.join(''), { parser: 'typescript' })
-  )
+  return {
+    header: headerContents.join(''),
+    body: contents.join('')
```
IlCallo, line 571: Shouldn't this join using a new line or similar?
I'm probably missing something here

ui/build/build.types.js
```diff
@@ -561,25 +566,75 @@ function writeIndexDTS (apis) {
...
-    copyPredefinedTypes(typeRoot)
-    writeIndexDTS(apis)
+    await Promise.all(copyPredefinedTypes(typeRoot))
+
+    const { header, body } = getIndexDts(apis)
+    const formattedBody = prettier.format(body, { parser: 'typescript' })
+
+    // The header contains stuff that breaks TS checking.
+    // So, write only the body at first to check the validity
+    await writeFile(resolvePath('index.d.ts'), formattedBody)
```
IlCallo, line 620: What is breaking TS checking in header? References to not-existing CLI packages?
I guess we can just add a `ts-ignore` on them in this case

Can't do `ts-expect-error` since we don't know which CLI will be used

IlCallo (commented):
Overall great work!
It would greatly simplify writing custom TS interfaces to expand the Quasar API
I just left a couple questions around

.github/workflows/build-types.yml
yusufkandemir: Yes, for the same workflow on the same ref. Ref is the PR's branch(_e.g. `someuser/quasar/patch-1`_) for `pull_request` events and `dev` for `push` events. So, it also applies if we quickly merge multiple PRs one after another. If we didn't specify `cancel-in-progress`, then it would just wait for the active ones to complete before starting.

ui/build/build.types.js
yusufkandemir: We are using the `writeLine` helper to push to these arrays, which already handle the new lines. So, here, we are correctly just joining them without any extra character in between.

ui/build/build.types.js
yusufkandemir: Yes, and I believe it's happening because there is only one `@ts-ignore`. I'll try with `@ts-ignore` for each related line to see if there will be another problem.

IlCallo: Yes, `ts-ignore` only works for the following line, we need 3 of them there

.github/workflows/build-types.yml
yusufkandemir: This workflow runs on `push` and `pull_request` events.

The `push` event runs on our own commits, or when we merge a PR. So, unless we merge a malicious PR or do malicious commits ourselves, we are safe.

The `pull_request` event runs in the context of the PR branch, so it runs on the fork or wherever the PR is created from. That means if there is malicious code, they will affect their own repo. It doesn't have write access to our repo, so we can't create PR comments in the same workflow for example. Because of that limitation, we take the PR number and the build results and upload them as an artifact so we can process them in another workflow(_`./comment-build-results.yml`_) as commented below that file. So, basically, the worse a malicious user could do is to trick the build results, which would be noticeable if the user modified the build script suspiciously.
The problematic thing would be if we used `pull_request_target` and also did an explicit checkout of the user's code. In that scenario, we would check out the user's untrusted code, then give them access to our secrets and such. So, unless you do both at the same time, you are pretty much safe.

I'd suggest giving https://securitylab.github.com/research/github-actions-preventing-pwn-requests a read, which I put in the related workflow.

ui/build/build.types.js
yusufkandemir: That didn't work. It goes deep into their type definitions and says stuff like "can't find `electron-builder`". We can't use `--skipLibCheck` either because that would disable our own `.d.ts` files too. That doesn't make sense, but that's what TS offers 🤷 So, the current solution is what we have, which is fine.

.github/workflows/build-types.yml
IlCallo: Yep, I read that but still missed the part explaining that `pull_request` executes into the fork environment instead of ours, thanks for the detailed explanation :)

PR #15404, opened 2023-02-07
Fix `currentIndex` calculation

tvogel:
For some reason, `selected` and `computedRows` do not contain the same objects, so `indexOf()` always fails. Locating the correct entry by `id` however works. Is `computedRows` rebuilt under the hood with new objects?

- [x] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
I did not look for existing bug reports. I found the example not to work with Firefox 108.0.2 (64-Bit) on openSUSE Tumbleweed, so I debugged it in https://codepen.io/eijk/pen/RwBOdbm and found this change to restore keyboard navigation.

### DIFF
--- a/docs/public/examples/QTable/KeyboardNavigation.vue
+++ b/docs/public/examples/QTable/KeyboardNavigation.vue
@@ -28,7 +28,7 @@
 </template>
 
 <script>
-import { ref, computed, nextTick } from 'vue'
+import { ref, computed, nextTick, toRaw } from 'vue'
 
 const columns = [
   {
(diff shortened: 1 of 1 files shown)


pdanpdan (changes requested):
Please add `toRaw` here

pdanpdan:
Please make some changes:

- import `toRaw` from vue
- convert the line you changed to `const currentIndex = selected.value.length > 0 ? computedRows.indexOf(toRaw(selected.value[ 0 ])) : -1`
- please also add a small fix - after line 196 (inside the nextTick, as a last line) add `tableRef.value.$el.focus()` to refocus the table when changing pages

tvogel:
Thanks! Used your suggestions. Only, I don't understand the third change because it does not seem to make a difference on Firefox!?

pdanpdan:
When you are on a last/first item in a page and navigate to next/previous page (arrow down/up) then the element that had focus is removed and sometimes the table loses focus so the next keypress will not be caught

tvogel:
OK, never happened to me while I played with it.

rstoenescu:
Thanks for contributing!

PR #15607, opened 2023-03-22
feat(create-quasar): add dev and build helper scripts

Nike682631:

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No


Nike682631:
@IlCallo Feel free to review. I have implemented the serve script by adding a function that firsts checks if @quasar/cli exists locally or globally. If yes then it is added otherwise not. Lemme know if I need to cover some edge case here. I have also tried to handle them for mainly npm, yarn and pnpm package managers.

create-quasar/index.js
```diff
@@ -102,6 +102,12 @@ async function run () {
       }
     })
 
+    await utils.extendJsonFile(scope.projectFolder + '/package.json', {
+      scripts: {
+        serve: "quasar serve",
+      },
+    });
```
IlCallo: Again: you shouldn't extend the `package.json` file since you have direct access to the templates.
How you added `dev` and `build` its the correct way to add a new default command into Quasar projects

create-quasar/package.json
```diff
@@ -29,6 +29,7 @@
     "lint": "eslint --ext .js,.ts ./ --fix"
   },
   "dependencies": {
+    "@types/lodash.merge": "^4.6.7",
```
IlCallo: You can directly add `@types/lodash`, since we have the whole `lodash` library listed as a dependency
But you must absolutely move this dependency into the `devDependencies`, as you only need it while developing this package and users won't need to download this to execute the `create-quasar` script

create-quasar/utils/index.js
```diff
@@ -218,6 +220,49 @@ module.exports.ensureOutsideProject = function () {
...
+
+module.exports.extendJsonFile = async function (filePath, newData) {
+  if (Object.keys(newData).length > 0) {
+    const localNpmResponse = await getShellResponse('npm list --depth=0 | grep @quasar/cli')
+    const localPnpmResponse = await getShellResponse('pnpm list --depth=0 | grep @quasar/cli')
+    const localYarnResponse = await getShellResponse('yarn list --depth=0 | grep @quasar/cli')
+    const globalPnpmResponse = await getShellResponse('pnpm -g list --depth=0 | grep @quasar/cli')
+    const globalYarnResponse = await getShellResponse('yarn global list --depth=0 | grep @quasar/cli')
+    const globalNpmResponse = await getShellResponse('npm list -g --depth=0 | grep @quasar/cli')
+    if (globalNpmResponse || globalYarnResponse || globalPnpmResponse || localYarnResponse || localPnpmResponse || localNpmResponse) {
```
IlCallo: Checks to discover if `@quasar/cli` has been installed globally are the right path, but you can probably also check directly if a `quasar` bin exists on the global scope instead, which is a package manager-independent way to detect the global installation of that package

Local installation of `@quasar/cli` isn't officially supported for now, so we shouldn't support it either

create-quasar/index.js
```diff
@@ -102,6 +102,12 @@ async function run () {
       }
     })
 
+    await utils.extendJsonFile(scope.projectFolder + '/package.json', {
+      scripts: {
+        serve: "quasar serve",
```
IlCallo: Let aside it won't serve anything if you don't provide a path for it to serve files from (`quasar serve dist/spa` would be a good default), are you really sure `quasar serve` will work as-is for all package managers?

I can already tell you that Yarn 1 will resolve the `quasar` reference there as the local quasar bin served from `@quasar/app-*` packages and break, since `serve` command is provided by the `@quasar/cli` the user is supposed to have installed globally.
For Yarn 1 this should work: `"serve": "$(yarn global bin)/quasar serve dist/spa"`

Pnpm could possibly have a similar problem

How did you test it was working correctly and on which package managers and versions?

IlCallo: To provide different versions of the `serve` script you're supposed to check with which package manager the user is executing `create-quasar` package

You can deduce the used package manager by analyzing the process.env data

Taken from the example project I provided you access to:
```
export const runningPackageManager = (process.env.npm_config_user_agent
  ?.split(" ")[0]
  .split("/")[0]);
```

Notice that if the package manager cannot be determined due to the missing `npm_config_user_agent`, you shouldn't add the `serve` script

IlCallo (changes requested):
You're on the good path, but still a lot of refinement is needed, carry on 👍🏻

create-quasar/utils/index.js
```diff
@@ -218,6 +220,36 @@ module.exports.ensureOutsideProject = function () {
...
+  try {
+    await execComm(command)
+    return true
+  } catch (e) { 
+    return false
+  }
+}
+
+module.exports.checkServeScript = async function (filePath) {
+  const linuxResponse = await getShellResponse('which quasar')
```
IlCallo: More like `unixResponse`, since it covers for Linux and Mac
But naming a variable should be more related to semantics and intent, to convey a message

How do you think you could rename these 2 variables to better convey their purpose?

create-quasar/utils/index.js
```diff
@@ -218,6 +220,36 @@ module.exports.ensureOutsideProject = function () {
...
+  const windowsResponse = await getShellResponse('where quasar')
+  if (!(linuxResponse || windowsResponse)) {
+    try {          
+      const fileData = existsSync(filePath)
+      ? 
+      (readJsonSync(filePath))
+      : {};
+      logger.log("I'm here *****************")
```
IlCallo: This whole approach, of adding the script by default and removing it later on, isn't good enough
But let that aside, never leave dev logging into the final code

create-quasar/templates/app-extension/ae-v1/BASE/_package.json
```diff
@@ -7,7 +7,10 @@
   "license": "<%= license %>",
   "main": "src/index.js",
   "scripts": {
-    "test": "echo \"No test specified\" && exit 0"
+    "test": "echo \"No test specified\" && exit 0",
+    "dev": "quasar dev",
+    "build": "quasar build",
+    "serve": "quasar serve dist/spa"
```
IlCallo: It should look more like

```json
  "scripts": {
    <% if (preset.lint) { %>"lint": "eslint --ext .js,.vue ./",<% } %>
    "test": "echo \"No test specified\" && exit 0",
    <% if (shouldAddServeScript) { %>"serve": "<%= globalQuasarBinPath %> serve dist/spa",<% } %>
    "dev": "quasar dev",
    "build": "quasar build"
  },
```

where `shouldAddServeScript` and `globalQuasarBinPath` are injected into the template scope

IlCallo (changes requested):

Nike682631:
There can be different ways of using the global quasar binary instead of local one. I'll probably go for one of below approaches:
- Use the which or where response depending on OS to get the global bin path and use it instead of local bin
Pros: - It should be possible to do everything within the existing index files for utils and create quasar.
Cons: -  There might be a possibility to have more than 1 global bins in case package has been installed multiple times. That needs to be tackled.

- Maybe write some sort of rerouting script where if the serve script accesses it it would get rerouted to global binary if ones exists.
Pros: - If in future some script needs to be implemented that requires local quasar bin routing to it if needed shouldn't be a problem.
Cons: -  This is just a speculation. It requires further research in the code to see if it can be implemented

- Another approach can be trying to access the global bin from the package manager being used inside quasar. In this case some way needs to be found out to point to global binary if the package was installed globally using some other package manager
Pros: - This approach is probably more consistent with the codebase.
Cons: -  It might not be possible to access other global bins from different package managers.

IlCallo:
> Use the which or where response depending on OS to get the global bin path and use it instead of local bin

If you plan to hardcode the result from `which` or `where`, the command will break as soon as someone else with a slightly different environment downloads your repo, or if you migrate to a new PC, etc

Here're some solutions for this last problem, all of which require compromises:
- find an NPM package abstracting which/where, install that one via Quasar `@quasar/app-*` packages and reference it in the script. This relies on creating a transient dependency on the local CLIs, which is bad, and adding a package which may or may not ever be used.
- run `which` and `where` every time the command is executed, take its result and use that to call the global bin. I'm not sure this is doable cross package manager, it's probably the most quick&dirty solution, but would only use native commands and avoid additional packages, which is good.

> rerouting script where if the serve script accesses it it would get rerouted to global binary if ones exists

This would need to adapt the bins from `app-vite` and `app-webpack` packages and then keep in sync forever, which is a bit of a maintenance burden. Those bins are actually meant to work without knowing that a global CLI exists, and it's preferable for them to avoid knowing it, while this change would create an implicit dependency on the global CLI public API

> access the global bin from the package manager being used inside quasar

As you correctly spotted, this would be a problem in case a package manager is used in the project and another one was used to install the global CLI, which actually happens pretty often by our knowledge. Let's avoid this option.

---

Another possibility would be to abstract the `serve` command code from the global CLI into an ad-hoc package, and require it both from global and local CLIs. This would still be bad, since you open up to version mismatches between CLIs, but it's better than duplicating the code directly.

For now I'd explore the quick&dirty solution which is ugly but reduces the code changes across multiple packages: call `which`/`where` at runtime every time, and use the result to run the global bin, hoping that it's possible to have it work this way

Good luck!

Nike682631:
So finally we won't be proceeding to add `serve` script just yet. Following were the approaches we took and blockers encountered regarding the same.
- First I went through the code and just simply tried adding `quasar serve dist/spa`. But this supposedly wouldn't work because there are two quasar binaries. One is global and other is local. For a package manager like npm it might work because it probably references the global bins first but for other package managers like yarn or pnpm this approach wouldn't work as they reference the local quasar binary instead of the global quasar binary created on installation of @quasar/cli
- The next approach was trying to add it using the installation path returned from which or where and using the path to binary in the serve script. Studied a bit about [child_processes](https://nodejs.org/api/child_process.html) here.  Now this approach did work but in case the user changed machines or the project was installed in a different environment this approach was likely to cause problems then.
- We tried adding @quasar/cli as a devDependency which made it work for yarn but not for pnpm as I believe pnpm still references the other local quasar binary. Also @quasar/cli can't be added as a dev dependency because it might cause issues which some features that comes with the main local quasar binary. For example it is likely that quasar.info wouldn't work.
- We also thought of command substitution but it is not a cross platform solution while we did search for other packages that might be available that could help us achieve this but were not able to find a concrete solution just yet.

## References
Below are some links we referenced during this process
- [Searching if there is an execa shell command list](https://www.npmjs.com/package/execa)
- [Command substitution](https://superuser.com/questions/289344/is-there-something-like-command-substitution-in-windows-cli)
- [yarnpkg/shell](https://github.com/yarnpkg/berry/tree/master/packages/yarnpkg-shell)
- [piping](https://stackoverflow.com/questions/52371318/cross-platform-pipe-command-in-npm-script)
- [cross-env](https://github.com/kentcdodds/cross-env)
- [shx](https://github.com/shelljs/shx/issues/65)

IlCallo (approved):

PR #15732, opened 2023-04-19
Add keep-color prop to QTree

duffyg:

- [ ] Bugfix
- [x] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
The checkboxes in the QTree component do not keep color when using the default color. See here: https://quasar.dev/vue-components/tree#example--syncing-node-properties (the "Pleasing decor" leaf is unticked and has a grey color)

However, when selecting a different color using the control-color prop, then keep-color is automatically applied.

This is inconsistent. Also, users may wish to switch off the keep-color prop the same way they can for the QCheckbox component.

This PR adds a new keep-color Boolean prop to Qtree, defaulting to true to avoid affecting existing QTree instances. This new prop is applied to the keep-color prop of all checkboxes in the QTree.

### DIFF
--- a/ui/src/components/tree/QTree.js
+++ b/ui/src/components/tree/QTree.js
@@ -44,7 +44,10 @@ export default createComponent({
 
     color: String,
     controlColor: String,
-    keepColor: { type: Boolean, default: true },
+    keepColor: {
+      type: Boolean,
+      default: true
+    },
     textColor: String,
     selectedColor: String,
 
--- a/ui/src/components/tree/QTree.js
+++ b/ui/src/components/tree/QTree.js
@@ -44,6 +44,7 @@ export default createComponent({
 
     color: String,
     controlColor: String,
+    keepColor: { type: Boolean, default: true },
     textColor: String,
     selectedColor: String,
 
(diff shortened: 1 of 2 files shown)


ui/src/components/tree/QTree.js
```diff
@@ -44,6 +44,7 @@ export default createComponent({
 
     color: String,
     controlColor: String,
+    keepColor: { type: Boolean, default: true },
```
suyundukov: Not on the same line.

```suggestion
    keepColor: {
      type: Boolean,
      default: true
    },
```

suyundukov (changes requested):

rstoenescu:
Thank you for taking the time to contribute this and for the clear write-up.

After reviewing it, we've decided not to move forward with it. The main blocker is the Boolean prop defaulting to `true`: we avoid that pattern in Quasar's API (Boolean props should default to `false`, like QCheckbox's own `keep-color` does), and shipping it this way would make QTree inconsistent with the rest of the UI components. Flipping the default to `false` isn't an option either, since that would be a breaking change to QTree's current look.

Also worth noting there's no actual inconsistency in the current behavior: `keep-color` is always applied to the inner checkboxes; it simply has no visible effect until a `color`/`control-color` is set, because there's nothing to keep.

We appreciate the effort and hope to see more contributions from you. Closing this one.

PR #15755, opened 2023-04-26
test(vite-plugin): Setup Cypress and Vitest for e2e and unit tests

yusufkandemir:

- Tests

- No

-  It's submitted to the `dev` branch (or `v[X]` branch)

**Other information:**

@rstoenescu you can review the test cases to see if they make enough sense or add new TODO comments that you have been planning to add. I couldn't properly understand the test cases as they were named 1-1, 1-2, 1-3, 2-1, etc. I tried preserving the existing test cases, adding some new ones, and re-organizing them.

Root-level package.json and yarn.lock files are meant for [Vite Ecosystem CI](https://github.com/vitejs/vite-ecosystem-ci). For now, I used yarn as it's the generally accepted one among us, and kept other packages outside the yarn workspaces feature since it's not the main point and not needed now. I've created a basic Yarn Workspaces setup before, [during the initial Vue 3 migration work](https://github.com/yusufkandemir/quasar/blob/refactor-migrate-to-vue3/package.json). From what I remember we didn't have any problems with it. Also considering how everything evolved since then, it should now be quite easy to integrate and should work great. We should discuss, plan, and migrate to pnpm workspaces as it is proven to work great. I have been using it for apps, libraries, single/multi project repos, etc. without any problems, just like the Vue/Vite community does and recommends. @rstoenescu if you don't have anything against it, I'd like to put it into the queue and tackle it in another PR.


package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "quasar-monorepo",
+  "private": true,
+  "packageManager": "yarn@1.22.19",
```
yusufkandemir, line 4: I had to specify this because Corepack doesn't let the ecosystem CI script run `yarn` inside a project which specifies `pnpm` as its package manager. Repeating what I mentioned in the PR description, we should consider migrating to pnpm.

vite-plugin/playground/src/components/regular-script/4-Custom.vue
```diff
@@ -0,0 +1,23 @@
...
+    <custom-component />
+  </div>
+</template>
+
+<script>
+import { defineComponent } from 'vue'
+import CustomComponent from './custom-component.js';
+
+export default defineComponent({
+  name: 'AliasedImports',
```
IlCallo: I guess this should be `Custom`

vite-plugin/playground/src/components/regular-script/5-CustomWithImports.vue
```diff
@@ -0,0 +1,27 @@
...
+<script>
+import { QBtn, Notify } from 'quasar'
+import { defineComponent } from 'vue'
+import CustomComponent from './custom-component.js'
+
+console.log(QBtn)
+console.log('Notify', Notify)
+
+export default defineComponent({
+  name: 'AliasedImports',
```
IlCallo: `CustomWithImports`

vite-plugin/playground/src/components/regular-script/6-CustomWithAliasedImports.vue
```diff
@@ -0,0 +1,27 @@
...
+<script>
+import { QBtn as Mimi, Notify } from 'quasar'
+import { defineComponent } from 'vue'
+import CustomComponent from './custom-component.js'
+
+console.log(Mimi)
+console.log('Notify', Notify)
+
+export default defineComponent({
+  name: 'AliasedImports',
```
IlCallo: `CustomWithAliasedImports`

vite-plugin/playground/src/components/regular-script/1-Basic.vue
```diff
@@ -1,19 +1,17 @@
 <template>
+  <h3>Basic</h3>
+
   <div>
     <q-btn label="Click me" />
+
     <q-icon name="map" />
```
IlCallo, line 7: Could make sense to use SVG icons import for all usage of material icons in tests, to avoid downloading and bundling it every time a CI test is run

Unless that's on purpose to test if material icons loading works, but that doesn't seem like a critical thing to test

IlCallo (changes requested):
Aside reviewing, I tried to build the plugin package and launch all commands you added, they all appear to be working

vite-plugin/playground/src/components/regular-script/1-Basic.vue
yusufkandemir: First of all, it's something that's got carried over from the old example folder.

Just speculating: bundling a CSS file is probably faster than processing a much bigger JS file, detecting and dropping the unused ones.

I tested it and it doesn't make more than ~0.5s difference (_total run time ~12.5s_). One way is faster than the other sometimes. So, overall, now worth it.

As neither the CSS nor the JS version is not very special, they should not be worthy of testing from the Vite plugin perspective. So, I am leaving the code as is since it's shorter and simpler.

vite-plugin/playground/src/components/regular-script/4-Custom.vue
yusufkandemir: Yeah, nice catch. I also added the `RS` prefix for uniqueness between the script setup and the regular script.

IlCallo:
LGTM, and

@rstoenescu let us know if you want to review it too, since it only touches test files and not production code as far as I can see

Merging this in will allow Quasar to be automatically tested against newer Vite and Vue versions BEFORE they are released for GA

IlCallo (approved):

PR #15766, opened 2023-04-27
chore: Add Vue Ecosystem CI scripts

yusufkandemir:

- Testing setup adjustments and adding package scripts

- No

**Other information:**


yusufkandemir:
This PR should be rebased after merging #15755 in, only after then it could get merged.

ui/.eslintrc.js
```diff
@@ -40,7 +40,14 @@ module.exports = {
         // Add Cypress-specific lint rules, globals and Cypress plugin
         // See https://github.com/cypress-io/eslint-plugin-cypress#rules
         'plugin:cypress/recommended'
-      ]
+      ],
+
+      rules: {
+        'cypress/no-unnecessary-waiting': 'off',
+        // It doesn't even allow .then(), it is too restrictive. It's quite new, needs time to mature.
+        'cypress/unsafe-to-chain-command': 'off',
```
yusufkandemir: This is probably no longer needed as #15782 got merged. @freddy38510 please confirm if the changes you've made were about this rule.

ui/package.json
```diff
@@ -32,11 +32,12 @@
...
-    "test:e2e:ci": "yarn test:build && cd ./dev && cross-env NODE_ENV=test start-test \"node ./script.dev.js\" http-get://localhost:9000 \"cypress run --e2e\" && cd ..",
-    "test:component": "yarn test:build && cd ./dev && cross-env NODE_ENV=test cypress open --component && cd ..",
-    "test:component:run": "yarn test:build && cd ./dev && cross-env NODE_ENV=test cypress run --component && cd ..",
-    "test:component:ci": "yarn test:build && cd ./dev && cross-env NODE_ENV=test cypress run --component --record --parallel && cd .."
+    "test:e2e": "yarn test:build && cross-env NODE_ENV=test start-test \"node ./dev/script.dev.js\" http-get://localhost:9000 \"cypress open --project ./dev --e2e\"",
+    "test:e2e:run": "cross-env NODE_ENV=test start-test \"node ./dev/script.dev.js\" http-get://localhost:9000 \"cypress run --project ./dev --e2e\"",
+    "test:e2e:ci": "yarn test:build && yarn test:e2e:run",
+    "test:component": "yarn test:build && cross-env NODE_ENV=test cypress open --project ./dev --component",
+    "test:component:run": "cross-env NODE_ENV=test cypress run --project ./dev --component",
```
IlCallo, line 39: Why have you removed "test:build" when using "test:component:run" and "test:e2e:run"?
"run" version is meant to be executed locally and needs to use latest build bundles

Seems like you're implying that for "run" commands the dev is meant to manually run "test:build", but that's not a good DX from my experience while working on Quasar tests, and that's why we added it in the first place

ui/.eslintrc.js
IlCallo: I confirm that the PR sent by Freddy solves lint errors due to `cypress/unsafe-to-chain-command` and `cypress/no-unnecessary-waiting`, thus those can be removed

ui/package.json
```diff
@@ -32,11 +32,12 @@
...
     "test:create": "node ./test/cypress/helpers/create-spec.js -c",
     "test:build": "node build/script.build.js js api && node build/script.build.js js transforms",
-    "test:e2e": "yarn test:build && cd ./dev && cross-env NODE_ENV=test start-test \"node ./script.dev.js\" http-get://localhost:9000 \"cypress open --e2e\" && cd ..",
-    "test:e2e:ci": "yarn test:build && cd ./dev && cross-env NODE_ENV=test start-test \"node ./script.dev.js\" http-get://localhost:9000 \"cypress run --e2e\" && cd ..",
-    "test:component": "yarn test:build && cd ./dev && cross-env NODE_ENV=test cypress open --component && cd ..",
-    "test:component:run": "yarn test:build && cd ./dev && cross-env NODE_ENV=test cypress run --component && cd ..",
-    "test:component:ci": "yarn test:build && cd ./dev && cross-env NODE_ENV=test cypress run --component --record --parallel && cd .."
+    "test:e2e": "yarn test:build && cross-env NODE_ENV=test start-test \"node ./dev/script.dev.js\" http-get://localhost:9000 \"cypress open --project ./dev --e2e\"",
+    "test:e2e:run": "cross-env NODE_ENV=test start-test \"node ./dev/script.dev.js\" http-get://localhost:9000 \"cypress run --project ./dev --e2e\"",
```
IlCallo, line 36: `test:e2e` and `test:e2e:run` reference q-app-vite internal commands `dev` and `build`, which can only be executed inside a Quasar project, hence why there was a `cd ./dev` and related `cd ..` in test commands

I guess it works with the root level monorepo due to it being a monorepo and hoisting some dependencies, but those scripts actually break down when executing inside `ui` folder during normal development

`component` scripts work because they don't rely on Quasar own `dev` and `build` commands, but we kept them with the change directory commands to keep consistency

IlCallo (changes requested):

ui/.eslintrc.js
freddy38510: @yusufkandemir, yes that was totally about this eslint rule.

ui/package.json
yusufkandemir: Previously `ci` was useless locally, and `run` was used to achieve the same locally. Now, we need a script that doesn't use the build script to avoid having to re-build the package when running tests, primarily to be used in CI, but also has a use locally. You don't have to rebuild the package if you only change the tests. I also updated the documentation to explain the updated purposes of the scripts.

I think of the `:ci` scripts as a way of mimicking the CI-way, usable locally and in CI. If the CI has very CI-specific settings, it can set env variables, pass extra arguments to the script, or override it without changing it completely(_when the first two ways are impossible_).

ui/package.json
yusufkandemir: As there are not many e2e tests and they are not important now, it's possible that I might have overlooked this. I have adjusted the scripts in ui/dev to ensure cwd is set correctly. `cd ./dev && ... && cd ..` is both annoying, can stay stuck in a directory in case of fail, and is not possible to pass extra arguments to, so I prefer avoiding it.

ui/.eslintrc.js
yusufkandemir: @freddy38510 thanks for confirming.

I removed the rule overrides, which caused other places to fail. I also addressed those places. Everything should be fine now.

ui/package.json
IlCallo: > You don't have to rebuild the package if you only change the tests.

You usually run tests even when working on the code and sometimes update the code when working on tests.
We automated the build step to avoid people forgetting the build step and ending up catching it in the CI instead, since it happened many times to me and others who contributed while working on the feature

I'll update the docs to mention `ci` version as the one to use, and `run` as a more granular and advanced one for special cases

IlCallo:
Code is ok now, but there's a problem with the dev project due to the monorepo hoisting

image

Since some dependencies are now hoisted at root level, they're outside Vite fs allowList and thus aren't loaded

Adding a second path pointing to root into the dev project allowList fix the problem, but I'm not sure if it may have unwanted side effects

```js
viteConf.server.fs.allow = [ '..', '../..' ]
```

Check it out, if that fix is fine, it's ok to merge for me

PR #15767, opened 2023-04-27
ci: Fix and improve tests workflow (fix: #15380)

### LINKED ISSUE #15380: Fix / improve PR tests workflow
PR tests workflow is using `pull_request_target` which must be used very carefully for security: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

Currently, the tests are run on the dev branch code and not using the PR code since we are not doing an explicit checkout. This doesn't pose a security threat because we are not executing any potentially unsafe code. But, since it runs the dev branch code and not the PR code, we can only know if the PR will actually break the build after we merge it. So, at the moment, it pretty much doesn't make any sense to have this workflow running for PRs at all.

So, we should address the logical problems of this workflow without risking security. `pull_request` + `workflow_run` is a good pattern that I applied in https://github.com/quasarframework/quasar/pull/15379.
But, it doesn't seem possible to record the Dashboard-related results into a file then upload it as an artifact, then upload it to Cypress Dashboard in another workflow.

yusufkandemir:

- CI

- No

**Other information:**
Resolves #15380

We decided to drop Cypress Cloud on PR runs. It's either a security risk or doesn't make sense to have workflow at all depending on which path is used, see the linked issue for some info. Cypress doesn't offer a way to output Cypress Cloud-ready results into a file, so we can't use the `pull_request` + `workflow_run` pattern. So, the only feasible way left is using `pull_request` + `workflow_run` but dropping Cypress Cloud, using custom reporting instead. We lose other benefits of Cypress Cloud, but this is about the best we can do right now. If Cypress starts offering a way to save/upload Cypress Cloud-ready results, we can adjust the workflows to use it.

### DIFF
--- a/.github/workflows/tests-on-pr-report.yml
+++ b/.github/workflows/tests-on-pr-report.yml
@@ -1,0 +1,74 @@
+name: Test results
+
+on:
+  workflow_run:
+    workflows:
+      - UI tests
+    types:
+      - completed
+
+jobs:
+  test-results:
+    name: Test Results
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion != 'skipped' }}
+
+    permissions:
+      checks: write
+
+      # for commenting the test results
+      pull-requests: write
+
+      # for using the artifacts API
+      actions: read
+
+    steps:
+      - name: Download the tests results artifacts
+        id: results
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const { readFile, writeFile, unlink } = require('node:fs/promises');
+            const { resolve } = require('node:path');
+            const __dirname = process.env.GITHUB_WORKSPACE;
+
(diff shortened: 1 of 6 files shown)


yusufkandemir:
This PR should be rebased after merging #15766 in, only after then it could get merged.

yusufkandemir:
The workflows are failing due to a separate problem in the build step: https://github.com/quasarframework/quasar/commit/5fe8dcf4d925543d02be1db39ae3f78e1283d0b1#commitcomment-110857726
Easy to fix, but I don't want to risk a potential conflict as Raz might have fixed it but forgot to push it. This PR is not ready to merge anyways, so not a problem.

Hopefully, the updated workflows will be working fine. When they pass, they should also post a comment on the PR with the test results.

IlCallo:
Seems like Raz added the missing JSON API definitions here: https://github.com/quasarframework/quasar/commit/cea8485ec9d24b0b42db322f60e05233285c2288

Issue #15808, opened 2023-05-10
QSelect set readonly and use-input，focus to field，then press backspace and options which selected would be deleted。


xiaojun777:
### What happened?

QSelect set readonly and use-input，focus to field， then press backspace and options which selected would be deleted。

### What did you expect to happen?

in readonly state, no option can deleted by user input. 

### Reproduction URL

https://codepen.io/onlldvei-the-encoder/pen/LYgdMOE

### How to reproduce?

click on select
press backspace on keyboard

### Flavour

Vue CLI Plugin (vue-cli-plugin-quasar)

### Areas

Components (quasar)

### Platforms/Browsers

Chrome

rstoenescu:
Fix will be available in Quasar 2.12.1 and 1.22.11

PR #15819, opened 2023-05-11
fix(QSelect): ignore most key commands when QSelect is not editable #15808

pdanpdan:


ui/src/components/select/QSelect.js
```diff
@@ -724,7 +724,7 @@ export default createComponent({
         return
       }
 
-      if (e.target === void 0 || e.target.id !== state.targetUid.value) { return }
+      if (e.target === void 0 || e.target.id !== this.targetUid || state.editable.value !== true) { return }
```
yusufkandemir:
```suggestion
      if (e.target === void 0 || e.target.id !== state.targetUid.value || state.editable.value !== true) { return }
```

pdanpdan: Shit, I'll fix it

PR #15883, opened 2023-05-22
ci: Add automated project creation tests

yusufkandemir:

- CI

- No

**Other information:**

Parameters to create different combinations:
```
script-type: [js, ts]
app-engine: [vite, webpack]
node-version: [16, 18, 20]
package-manager: [yarn, npm, pnpm]
```

Using the defaults:
- SCSS
- ESLint + Prettier
- Composition API

The tests will be run:
- when triggered manually through the GitHub UI
- on PRs that update create-quasar
- any time these packages get a new release:
	- `quasar`
	- `@quasar/extras`
	- `@quasar/app-webpack`
	- `@quasar/app-vite`


yusufkandemir:
First benefits of automated tests:
- Found out project creation uses outdated dependencies when using pnpm v8, fixed it.
- Found out Vite projects don't have Node v20 in the engines field, added it. Also adjusted the engines field to fully reflect the current version support (v20, v18, v16)

create-quasar/index.d.ts
```diff
@@ -0,0 +1 @@
+export default {};
```
IlCallo: This one is only needed to make `await import('../index.js');` in `create-quasar/scripts/create-test-project.ts` to work without TS errors, right?

Can you add a comment making it explicit?

IlCallo (approved):
Is it possible to run the tests automatically **before** a release is published, with our current release workflow?
E.g. run tests when a draft release is created or updated too

If we create a "release" CI action in the future, I'd expect these tests to be run before proceeding and stopping the release process if the test fail, while allowing manual override to publish anyway

Does it have any value to run these tests together with the ones currently run by Vite and Vue ecosystem CI?

Aside from these notes, everything seems fine

IlCallo:
Bump @yusufkandemir

yusufkandemir:
@IlCallo 

> Is it possible to run the tests automatically **before** a release is published, with our current release workflow?
E.g. run tests when a draft release is created or updated too

We aren't always creating a draft release beforehand, so we can't guarantee the workflows will run before every release. But, yes, we can run them whenever a draft release is created or updated.

However, creating/editing a release only concerns the details of the releases, not the code related to them. So, one might update the release notes, but not the code, or vice versa. So, the trigger mechanism will not be accurate. That's why I made it possible to trigger it via GitHub UI on-demand. To accurately run the tests, one would eventually have to use manual runs anyway, so making it a part of the manual flow makes more sense to me. So, I am opposed to running it before a release has been made.

Also, create-quasar fetches the packages from the registry, so running the command before an npm release doesn't make much sense. In the future, we can spin up a local npm registry such as Verdaccio, to be able to make it use the unreleased versions. Simply using npm linking or a similar approach won't work as we use yarn in our monorepo, but the created projects will use npm/yarn/pnpm, so the file linking won't resolve stuff properly. For that reason, we have to use a local npm registry. And, that requires extra effort and time.

So, the actual benefit of these automated tests is to be quickly aware of any errors after making any modifications to create-quasar(_before releasing_) or after releasing the packages. Then, we can quickly publish a fix release afterward.

create-quasar/index.d.ts
yusufkandemir: After rebasing, probably due to TS v5, it was no longer needed. So, I removed the .d.ts file altogether.

IlCallo:
Ok, in this case, can you create an MD file which explains how the release process involves these tests and how/when are you supposed to run them manually?
Then, after we merge this, make sure Raz is aware of this additional step

IIRC another idea was to automate the release process itself via a GH action, but we can work on it later on

yusufkandemir:
At the moment, I think you would only manually run to test if packages into the scaffolded project, such as ESLint, Prettier, etc. were breaking, and then they got a new release and we want to check if it fixed the problem. There is not much to be aware of, it's just there in case it's useful.
The overall behavior is already documented in this PR. We can document it clearly when we improve the flow to introduce a local npm registry and when we create an automated release process. So, I don't think we need further action right now.

> IIRC another idea was to automate the release process itself via a GH action, but we can work on it later on

Yes, we have that idea, but it needs further discussion and work. So, yeah, it needs to be handled separately later.

PR #15913, opened 2023-05-28
fix(QFile): incomplete union type in QRejectedEntry['failedPropValida…

timarito:
The `processFiles` function in ui/src/composables/private/use-file.js can reject 7 different props ("accept" | "max-file-size" | "max-total-size" | "filter" | "max-files" | "duplicate"), but only 4 declared in types. This fix amends the `QRejectedEntry["failedPropValidation"]` type to match the actual behavior


rstoenescu:
Thanks for contributing!

PR #15945, opened 2023-06-05
feat: Improve quasarConfOptions, generate types for it, improve docs (fix: #14069)

### LINKED ISSUE #14069: Update `quasarConfOptions` and generate TS types for it
See this comment: https://github.com/quasarframework/quasar/commit/e3b54bc978514349df512f271d4fd0689ae8fd04#r79469945

We need to update `quasarConfOptions` to allow non-object types to be assigned. `Ripple` can take `boolean | ...`, `Dark` can take `boolean | "auto"`, but unlike some other top-level types such as `props`, `quasarConfOptions` doesn't allow `type`, making it impossible to document properly ATM. After updating the JSON API system to make this possible, we should also update the JSON API files for the related directives/plugins by adding the possible value types.

After achieving the point above, we can start generating TS types for `quasarConfOptions` like we are doing with some other top-level properties. Then, we can import these types from the `quasar` package into `app-vite` and `app-webpack` to keep the type definitions in sync.

yusufkandemir:

- Feature
- Documentation
- Refactor

- No

**Other information:**
Resolves #14069.

Previously undocumented non-object configuration example(Dark plugin):
image
Previously undocumented "plugin-like" examples:
image
image

Config-only installation cards use the title "Configuration" for improved clarity.

We are now generating and exporting a new type called `QuasarUIConfiguration` from `quasar` that holds the configurable options. That type will now be used by the `config` key of the Vue plugin installation, so Vite plugin/Vue CLI users will get config type-safety now. app-vite and app-webpack will use this type for quasar.config file types, so the types will always be in-sync without effort, and be compatible with the UI package version used in the app.

### DIFF
--- a/app-vite/types/configuration/framework-conf.d.ts
+++ b/app-vite/types/configuration/framework-conf.d.ts
@@ -2,113 +2,11 @@ import {
   QuasarIconSets,
   QuasarLanguageCodes,
   QuasarPluginOptions,
+  QuasarUIConfiguration,
 } from "quasar";
-import { Component } from "vue";
-
-interface QuasarMobileFrameworkInnerConfiguration {
-  iosStatusBarPadding: boolean;
-  backButton: boolean;
-  backButtonExit: boolean | "*" | string[];
-}
-
-interface RippleQuasarConf {
-  early?: boolean;
-  stop?: boolean;
-  center?: boolean;
-  color?: string;
-  keyCodes?: number[] | number;
-}
-
-type VueClassObjectProp = {
-  [value: string]: any
-}
-
-type VueClassProp =
-  | string
-  | Array<VueClassProp>
-  | VueClassObjectProp;
-
-type VueStyleObjectProp = Partial<CSSStyleDeclaration>;
-
-type VueStyleProp =
-  | string
-  | Array<VueStyleProp>
-  | VueStyleObjectProp;
-
-interface QuasarFrameworkInnerConfiguration {
-  brand?: {
-    primary?: string;
-    secondary?: string;
-    accent?: string;
-    dark?: string;
-    positive?: string;
-    negative?: string;
-    info?: string;
-    warning?: string;
-  };
-  capacitor?: QuasarMobileFrameworkInnerConfiguration;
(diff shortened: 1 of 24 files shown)


rstoenescu:
Great work!

dmnkrthrt:
@rstoenescu & @yusufkandemir 

What's the purpose of removing `hijackFilter()` from QLoadingBarOptions? How do I skip routes from a QLoadingBar now? Your docs still mention `hijackFilter()` in `LoadingBar.setDefaults()`.

Thank you.

yusufkandemir:
@dmnkrthrt as `hijackFilter` is a function, it can't be serialized, so it can't be accepted from quasar.config file. However, as you correctly pointed out, it's possible to set it through `LoadingBar.setDefaults()`. It's also possible to set it through the Quasar plugin options(_`app.use(Quasar, { ... })`_) when using flavors other than Quasar CLI. So, the type should only be disallowed in quasar.config file. Thanks for raising awareness, I'll address this.

dmnkrthrt:
@yusufkandemir you wrote it's possible to set it via `LoadingBar.setDefaults()` - but I can't confirm that (and that's exactly my issue). See vue-tsc output:

`error TS2345: Argument of type '{ color: string; size: string; hijackFilter(url: string): boolean; }' is not assignable to parameter of type 'QLoadingBarOptions'.
  Object literal may only specify known properties, and 'hijackFilter' does not exist in type 'QLoadingBarOptions'.`

Using `app.use(...config: { loadingBar: ... })` results in another TS error:

`error TS2769: No overload matches this call.
  Overload 1 of 2, '(plugin: Plugin<[options: Partial<QuasarPluginOptions>]>, options: Partial<QuasarPluginOptions>): App<Element>', gave the following error.
    Type '{ loadingBar: { color: string; size: string; hijackFilter(url: string): boolean; }; }' is not assignable to type 'QuasarUIConfiguration'.`

yusufkandemir:
@dmnkrthrt I meant from a functionality perspective. If you bypass the type errors, you will see that it works correctly.

I included a fix in #16049

PR #16033, opened 2023-07-03
test(ui/use-field&QField): add tests for use-field composable and QField

n05la3:

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: tests

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/components/field/__tests__/QField.cy.js
+++ b/ui/src/components/field/__tests__/QField.cy.js
@@ -1,0 +1,168 @@
+import QField from '../QField'
+import { vModelAdapter } from '@quasar/quasar-app-extension-testing-e2e-cypress'
+import { ref } from 'vue'
+import Icons from '../../../../icon-set/material-icons.mjs'
+
+function getHostElement () {
+  return cy.get('.q-field')
+}
+
+function mountQField (options) {
+  return cy.mount(QField, options)
+}
+
+describe('Field API', () => {
+  describe('Props', () => {
+    describe('Category: model', () => {
+      describe('(prop): maxlength', () => {
+        it.skip(' ', () => {
+          //
+        })
+      })
+    })
+  })
+
+  describe('Slots', () => {
+    describe('(slot): control', () => {
+      it('should use control slot as content of the field', () => {
+        const controlSlot = 'Hello there'
+        mountQField({
+          slots: {
+            control: () => controlSlot
+          }
+        })
+
+        getHostElement().get('.q-field__control-container').should('contain', controlSlot)
+      })
+    })
+
+    describe('(slot): rawControl', () => {
+      it('should use raw control slot', () => {
+        const rawControlSlot = 'Hello there'
+        mountQField({
+          slots: {
+            rawControl: () => rawControlSlot
+          }
+        })
+
+        getHostElement().get('.q-field__control-container').should('contain', rawControlSlot)
+      })
+    })
+  })
+
+  describe('Events', () => {
+    describe('(event): update:model-value', () => {
+      it('should emit onUpdate:modelValue event', () => {
+        const model = ref('text')
+        const fn = cy.stub()
+        mountQField({
+          props: {
+            ...vModelAdapter(model),
+            'onUpdate:modelValue': fn,
+            clearable: true
+          }
+        })
+
+        getHostElement().get('button[type="button"]')
+          .contains(Icons.field.clear).click()
+          .then(() => {
+            expect(fn).to.be.calledWith()
+          })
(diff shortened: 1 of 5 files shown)


ui/src/components/field/__tests__/QField.cy.js
```diff
@@ -0,0 +1,168 @@
...
+          slots: {
+            control: () => controlSlot
+          }
+        })
+
+        getHostElement().get('.q-field__control-container').should('contain', controlSlot)
+      })
+    })
+
+    describe('(slot): rawControl', () => {
```
IlCallo: Let's remove this one, it's a slot which is only used internally, dunno why it's actually listed into the API box

ui/src/components/field/__tests__/QField.cy.js
```diff
@@ -0,0 +1,168 @@
...
+
+function mountQField (options) {
+  return cy.mount(QField, options)
+}
+
+describe('Field API', () => {
+  describe('Props', () => {
+    describe('Category: model', () => {
+      describe('(prop): maxlength', () => {
+        it.skip(' ', () => {
```
IlCallo, line 18: Have you left this out on purpose? Why?

ui/src/composables/private/__tests__/use-field.cy.js
```diff
@@ -1,37 +1,113 @@
...
+            props: {
+              ...vModelAdapter(model),
+              clearable: true,
+              options: selectOptions
+            }
+          })
+
+          getHostElement().find('input').should('exist').should('have.value', model.value)
+          getHostElement().get('button[type="button"]').contains(Icons.field.clear).click()
+          getHostElement().get('input').should('not.have.value', model.value)
```
IlCallo: I'd use something like `getHostElement().get('input').should('have.value', '')` instead, or whatever works for "contains an empty value" instead of "doesn't contain the value we initially set"

ui/src/composables/private/__tests__/use-field.cy.js
```diff
@@ -133,102 +209,206 @@ describe('use-field API', () => {
...
+            },
+            slots: {
+              hint: () => bottomSlot
+            }
+          })
+
+          getHostElement().find('.q-field__bottom').should('contain.text', bottomSlot)
+        })
+
+        it('should use a bottom counter slot', () => {
```
IlCallo, line 314: Isn't this test the same as the previous one?

ui/src/composables/private/__tests__/use-field.cy.js
```diff
@@ -238,112 +418,133 @@ describe('use-field API', () => {
...
 
   describe('Methods', () => {
     describe('(method): focus', () => {
-      it.skip(' ', () => {
-        //
+      it('should focus the component', () => {
+        mountQFieldWrapper()
+
+        getHostElement()
+          .get('[tabindex="0"]')
```
IlCallo: This way of selecting the input feels really strange, any reason why you went with this one?

IlCallo (changes requested):
Couple of doubts around, but mostly good

ui/src/composables/private/__tests__/use-field.cy.js
n05la3: Fixed.

ui/src/composables/private/__tests__/use-field.cy.js
n05la3: Fixed. This one should using the `counter` prop instead of the `hint` slot as the previous one.

ui/src/composables/private/__tests__/use-field.cy.js
n05la3: I have refactored to use `.get('input')` which is also as precise and better than `[tabindex="0"]`.

ui/src/components/field/__tests__/QField.cy.js
n05la3: Unless I missed something then the `maxlength` on `QField` makes it difficult to test since we will require a control slot with a `v-model` and this is already tested with `QInput` and the rest, so I'm not exactly sure how and if we should be testing this.

IlCallo: Its ok, but leave a comment explaining this and why we're skipping it here

n05la3: Done.

PR #16068, opened 2023-07-13
feat(ui): generate types for vue directives

yusufkandemir:

- Feature

- No

**Other information:**
Resolves #15798. This was already planned way before the discussion thread, but linking it anyway.

The types we are generating for directives will no longer be empty interfaces. They will now use the `Directive<Target, Value>` type.

Volar, unfortunately, doesn't provide auto-completion for directives when doing `v-`. I am not sure if this is a known issue. But, it recognizes them and does type-checking. It's not possible to strongly type arg and modifiers, so they are documented as comments. It will be possible with https://github.com/vuejs/core/pull/3399, but without the ability to add an extra explanation for individual things from the way it is structured, it seems. So, the comments will be useful in any case.

We are now generating new types for each directive with the `Value` suffix(e.g. `TouchPanValue`) which holds the type of the value the directive expects. This type can be used to create variables to feed to the directive. Here is an example:
```html
<template>
  <div v-touch-pan.mouse="touchPanHandler">
    Touch Here
  </div>
</template>

<script setup lang="ts">
import { TouchPanValue } from 'quasar';

const touchPanHandler: TouchPanValue = (event) => {
  console.log(event);
};
</script>
```
image

Extra screenshots:

image
image
image
image

### DIFF
--- a/ui/build/build.types.js
+++ b/ui/build/build.types.js
@@ -353,10 +353,21 @@ function getIndexDts (apis) {
       const valueType = getTypeVal(content.value)
 
       const directiveValueType = `${ typeName }Value`
+      const comments = [
+        '/**',
+        ` * ${ content.value.desc }`,
+        ' *',
+        ` * @see ${ content.meta.docsUrl }`,
+        ' */'
+      ].join('\n')
+
+      write(contents, comments + '\n')
       writeLine(contents, `export type ${ directiveValueType } = ${ valueType }`)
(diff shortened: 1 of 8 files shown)


jaybo:
To use **TouchPanValue** in Quasar 2.18.6 without lint errors, Claude tells me I need to do this:
```
onPan(evt: Parameters<TouchPanValue>[0]) {
...
}
```
Which seems bizarre.

PR #16150, opened 2023-08-04
feat(QChatMessage): make color support css color

chuoke:
This PR code is inspired by the Vuetify color prop

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] Make the user's custom style convenient
- [x] Ability to use colors other than the Color Palette
- [x] Easy to use in combination with color-picker

**Other information:**

If you agree to accept, I will supplement the document

preview: 
image

### DIFF
--- a/ui/dev/src/pages/components/chat.vue
+++ b/ui/dev/src/pages/components/chat.vue
@@ -123,8 +123,10 @@ export default {
       },
       {
         name: 'Jane',
-        text: [ 'And you?' ],
+        text: [ 'And you?', 'This bg color is #497ef2ff' ],
         sent: true,
+        bgColor: '#497ef2ff',
+        textColor: '#e4f7f6',
         avatar: 'https://cdn.quasar.dev/img/linux-avatar.png',
         stamp: 'Yesterday at 13:51'
       },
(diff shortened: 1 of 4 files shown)


rstoenescu:
Thank you for the work on this and apologies that it sat unanswered for so long.

After review we've decided not to move forward with it. Across the whole framework, color props accept Quasar color palette names, and QChatMessage already covers custom colors through that contract: any color you register as a custom palette color ([docs](https://quasar.dev/style/color-palette#adding-your-own-colors)) works with `text-color` / `bg-color` as-is. For fully dynamic values (e.g. from QColorPicker) you can bind a CSS custom property:

```html
<q-chat-message class="my-msg" :style="{ '--bubble-bg': msg.bgColor, '--bubble-text': msg.textColor }" ... />
```
```css
.my-msg .q-message-text { color: var(--bubble-bg) }
.my-msg .q-message-text-content { color: var(--bubble-text) }
```

Accepting raw CSS values in this one component would make the color prop behave differently from the rest of the framework, and the `isCssColor()` detection can't cover named colors or newer functions (`oklch()`, `color-mix()`), so some valid CSS colors would silently fall through to non-existent classes. Given the workarounds above we prefer to keep the current contract. Thanks again for contributing!

PR #16211, opened 2023-08-18
fix(QScrollArea): mouse events on contents trigger on first tap in iOS #16210

### LINKED ISSUE #16210: QScrollArea requires two taps to trigger click event on elements contained within [iOS]
### What happened?

The way Safari was programmed to work, means that having a @mouseenter event that modifies DOM prevents event like @mouseup and @click from being triggered down the DOM tree. In case of QScrollArea, @mouseenter event is used to display the scroll thumb, which means that a second tap is required to interact with elements within QScrollArea.

image

Taken from [this Safari documentation page](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/HandlingEvents/HandlingEvents.html)

Now, Apple would not be Apple if it was consistent with its issues. This does not happen on some devices and happens on others, but I was able to successfully reproduce this with BrowserStack. I checked WebKit code, but I couldn't find any specific reference to the implementation of this behavior.

Quasar code that triggers this Safari issue is this one:

https://github.com/quasarframework/quasar/blob/f8b458292ec1caef77420cc5197454ee66fe4fc8/ui/src/components/scroll-area/QScrollArea.js#L355-L357

### Reproduction URL

NA

### How to reproduce?

The workaround requires delaying any DOM modifications until a short cooldown period had passed. From my testing, we don't have to wait until mouseup event had triggered, but making modification within setTimeout(0), rAF() or nextTick() leaves insufficient delay. Making it a capture event doesn't make a difference either. The safest approach that I found is to do:

```
function onMouseenter () {
  setTimeout(() => {
    hover.value = true
  }, proxy.$q.platform.is.ios ? 50 : 0)
}
```

The 50ms delay is arbitrary, but it seems to work for me every time.

### Flavour

### Areas

Components (quasar)

### Platforms/Browsers

iOS

### Quasar info output

_No response_

### Relevant log output

_No response_

### Additional context

Current behavior:

opera_5dfOrPgr3E

With the proposed change:

opera_rXPMGQUO8D

thexeos:
Fixes #16210

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/components/scroll-area/QScrollArea.js
+++ b/ui/src/components/scroll-area/QScrollArea.js
@@ -353,6 +353,7 @@ export default createComponent({
     }
 
     function onMouseenter () {
+      // setTimeout needed for iOS, see #16210
       setTimeout(() => {
         hover.value = true
       }, proxy.$q.platform.is.ios ? 50 : 0)
--- a/ui/src/components/scroll-area/QScrollArea.js
+++ b/ui/src/components/scroll-area/QScrollArea.js
@@ -353,7 +353,9 @@ export default createComponent({
     }
 
     function onMouseenter () {
-      hover.value = true
+      setTimeout(() => {
+        hover.value = true
+      }, proxy.$q.platform.is.ios ? 50 : 0)
     }
 
     function onMouseleave () {


rstoenescu:
Thank you for contributing!

PR #16213, opened 2023-08-18
fix(ui): define both UI and quasar.config file compatible types in quasarConfOptions (fix: #16046)

### LINKED ISSUE #16046: Quasar plugin config property is typed incorrectly
### Discussed in https://github.com/quasarframework/quasar/discussions/16045

<sup>Originally posted by **ghiscoding** July  5, 2023</sup>
I have global configs defined for Quasar and that used to work properly before updating (mostly because the `config` property of `QuasarPluginOptions` was set to `any` so we could provide anything we wanted really)

So considering this global config

```ts
const quasarOptions: Partial<QuasarPluginOptions> = {
  // Quasar plugins
  plugins: {
    Dialog,
    Notify,
  },
  config: {
    /* Notify defaults */
    notify: {
      position: 'top',
      timeout: 2500,
    },
  },
};

const app = createApp(App);
app.use(Quasar, quasarOptions)
```
this no longer work and tells me what `notify` is not a known property of `QuasarPluginOptions` and seem to be caused by this [PR #15945](https://github.com/quasarframework/quasar/pull/15945)

> Type '{ notify: { position: string; timeout: number; textColor: string; actions: { icon: string; color: string; }[]; }; loading: { spinner: ComponentConstructor<QSpinnerClock>; ... 4 more ...; customClass: string; }; }' is not assignable to type 'QuasarUIConfiguration'.
  Object literal may only specify known properties, and 'notify' does not exist in type 'QuasarUIConfiguration'.ts(2322)
plugin.d.ts(13, 3): The expected type comes from property 'config' which is declared here on type 'Partial<QuasarPluginOptions>'

image

I am not expecting any errors since even the documentation still mentions the same way to configure the global configs

https://quasar.dev/quasar-plugins/notify#installation

image

Am I doing something wrong or is that a Regression Bug?</div>

yusufkandemir:

- Bugfix
- Documentation

- No

**Other information:**
Fixes #16046

(_improved search will come in a separate PR, searching through quasarConfOptions like in the screenshot doesn't work in the current docs_)
image
image
image

### DIFF
--- a/docs/src/components/DocApiEntry.js
+++ b/docs/src/components/DocApiEntry.js
@@ -1,5 +1,5 @@
 import { h, ref } from 'vue'
-import { QBadge, QBtn, Notify } from 'quasar'
+import { QBadge, QBtn, Notify, QBtnToggle } from 'quasar'
 import { copyToClipboard } from 'assets/page-utils.js'
 import { mdiMinusBox, mdiPlusBox } from '@quasar/extras/mdi-v6'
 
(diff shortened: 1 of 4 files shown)


rstoenescu:
Awesome!

PR #16263, opened 2023-08-28
fix(ui): QLayout containerized on iOS, only force fixed when reveal

Evertvdw:
Adds a check to QFooter and QHeader that only forces a fixed footer/header when reveal is true.

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/components/footer/QFooter.js
+++ b/ui/src/components/footer/QFooter.js
@@ -48,7 +48,7 @@ export default createComponent({
     const fixed = computed(() =>
       props.reveal === true
       || $layout.view.value.indexOf('F') > -1
-      || ($q.platform.is.ios && $layout.isContainer.value === true)
+      || ($q.platform.is.ios && props.reveal === true && $layout.isContainer.value === true)
     )
 
     const containerHeight = computed(() => (
--- a/ui/src/components/header/QHeader.js
+++ b/ui/src/components/header/QHeader.js
@@ -45,7 +45,7 @@ export default createComponent({
     const fixed = computed(() =>
       props.reveal === true
       || $layout.view.value.indexOf('H') > -1
-      || ($q.platform.is.ios && $layout.isContainer.value === true)
+      || ($q.platform.is.ios && props.reveal === true && $layout.isContainer.value === true)
     )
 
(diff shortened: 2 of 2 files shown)


rstoenescu:
Thank you for the PR, and for your patience here (and for #11369 before it, which is what narrowed this forced-fixed behavior down to iOS in the first place). After a fresh review against current `dev` we're closing this one, for two reasons.

First, as written the change is a no-op. The `fixed` computed is `props.reveal === true || view.includes('H') || (ios && isContainer)`. Since the first clause of the chain is already `props.reveal === true`, adding `props.reveal === true` to the iOS clause means that clause can never contribute: whenever it would be true, the first clause has already short-circuited the whole expression. So the diff is exactly equivalent to deleting the iOS clause outright, which (per the description) is what was intended, but that's the change that actually needs reviewing.

Second, deleting the iOS clause on its own would not deliver scroll-away headers/footers on iOS; it would break them differently. The JS clause is coupled to a CSS workaround from #4127 in QLayout.sass: `body.platform-ios .q-layout--containerized { position: unset !important }`. Everywhere else `.q-layout` is `position: relative`, so a non-fixed (`absolute-top`) header is anchored inside the scrolled content and scrolls away naturally. On iOS that rule makes the layout static, so the header would anchor to the container's `.absolute-full` frame instead: visually still pinned, while the layout's offset bookkeeping assumes it scrolls away. QDrawer carries the same iOS + isContainer forced-fixed clause, for the same reason.

So an actual fix means reworking the whole family together (QHeader, QFooter, QDrawer and the #4127 CSS rule), verifying on devices whether modern iOS still needs the Dialog scroll workaround at all, and covering the older iOS versions still in the support window. That's well beyond this two-liner, and the branch has meanwhile drifted into conflict with `dev`.

If you'd like to pursue it, the way in is a fresh issue/PR that treats both workarounds as one change, device-tested for containerized layouts inside QDialog (the original #4127 case) and for drawers. Thanks again!

Issue #14362, opened 2022-09-11
SASS deprecation warnings


fprijate:
### What happened?

SASS has multiple deprecation warnings (see  log)

### What did you expect to happen?

No warnings

### Reproduction URL

https://codepen.io/fp/xxx

### How to reproduce?

see log

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Quasar CLI Commands/Configuration (@quasar/cli | @quasar/app-webpack | @quasar/app-vite)

### Platforms/Browsers

_No response_

### Quasar info output

_No response_

### Relevant log output

```Shell
Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sas
s 2.0.0.                                                                                                 
                                                                                                         
Recommendation: math.div($angle, 180) or calc($angle / 180)                                              
                                                                                                         
More info and automated migrator: https://sass-lang.com/d/slash-div                                      
                                                                                                         
   ╷                                                                                                     
42 │   $rad: $angle / 180 * $PI                                                                          
   │         ^^^^^^^^^^^^                                                                                
   ╵                                                                                                     
    node_modules\quasar\dist\quasar.sass 42:9     sin()                                                  
    node_modules\quasar\dist\quasar.sass 4490:16  root stylesheet                                        
                                                                                                         
Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sas
s 2.0.0.                                                                                                 
                                                                                                         
Recommendation: math.div(pow(-1, $i) * pow($rad, (2 * $i + 1)), fact(2 * $i + 1)) or calc(pow(-1, $i) * p
ow($rad, (2 * $i + 1)) / fact(2 * $i + 1))                                                               
                                                                                                         
More info and automated migrator: https://sass-lang.com/d/slash-div                                      
                                                                                                         
   ╷                                                                                                     
45 │     $sin: $sin + pow(-1, $i) * pow($rad, (2 * $i + 1)) / fact(2 * $i + 1)                           
   │                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                           
   ╵                                                                                                     
    node_modules\quasar\dist\quasar.sass 45:18    sin()                                                  
    node_modules\quasar\dist\quasar.sass 4490:16  root stylesheet                                        
                                                                                                         
[...]

PR #16393, opened 2023-09-25
fix: fix sass depration warnings

sschneider-ihre-pvs:
Long time deprecation warnings fixed

### DIFF
--- a/ui/src/css/helpers/math.sass
+++ b/ui/src/css/helpers/math.sass
@@ -21,21 +21,21 @@ $PI: 3.14159265359
 // toFixed(0.12345, 100) -> 0.12
 // toFixed(0.12345, 1000) -> 0.123
 @function toFixed($number, $power)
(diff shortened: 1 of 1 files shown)


ghiscoding:
I assume that would fix #14362 and #11683 but I wonder what the core team member will say this time since his last comment (in the first issue) was to pin the SASS version to `1.32.13` (before all these SASS warnings came in to play). It would be awesome if we could finally upgrade to latest SASS 🚀

wt0m:
Great news! Will we be able to finally upgrade to the latest SASS?

staszek998:
Is there a way to somehow "bump" this PR, so it'll be merged into production earlier than later? 😉

rstoenescu:
Hi,

Thanks for contributing! We went with another approach, but regardless, closing this as everyone can now use latest Sass.

PR #16420, opened 2023-10-01
Border property for Chat Messages

AlecBytes:
This is a feature to allow for a red border on chat messages. The red border can be used to signify an urgent message or similar semantics. If this is approved, then it can be expanded to include more options for color and size of border.

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


rstoenescu:
Hi,

Thank you for contributing!
However, this is best suited to be done in app-space, outside of Quasar, as it's one of the things that can be customized from an infinite number of options. First, we cannot force a border color (like "red"), then we need to draw a line on the number of props that can style any component -- otherwise we'll just end up with 1k styling props for every component (which in turn has very negative effect on memory consumption and runtime speed, for essentially what can easily be written as a CSS rule by the user in their own app).

PR #16513, opened 2023-10-24
fix(ui): fix generic usage of QTreeNode type (fix: #15351)

### LINKED ISSUE #15351: QTree not using given generic param.
### What happened?

Using QTreeView, an attempt was made to add an ID/key to the nodes being shown on the tree. Doing this requires augmenting the QTreeNode<unknown> with a different type, in this case, TreeNode.
Structure:

```ts
type TreeNode = { key: string } & QTreeNode<TreeNode>;
const nodes: TreeNode[] = [
  {
    key: 'foo',
    label: 'bar',
    children: []
  }
];
```
```html
<q-tree v-model:selected="selected" :nodes="nodes" dense node-key="index" />
```

The nodes prop would accept the augmented type without issue, as the type had a declaration and as such should fulfill the requirement of the `QTreeNode<unknown>`.

### Reproduction URL

No repro link, see images.

### How to reproduce?

[Error statement](https://media.discordapp.net/attachments/807654245640962059/1069937693871579176/image.png)

[Type image](https://media.discordapp.net/attachments/807654245640962059/1069937742869446656/image.png)

### Flavour

### Areas

Components (quasar)

### Platforms/Browsers

_No response_

### Quasar info output

_No response_

### Relevant log output

_No response_

### Additional context

This can be fixed by doing the following instead, however, it highlights the problem that QTreenode doesn't really use the given generic parameter.
```ts
type TreeNode = { key: string } & QTreeNode<TreeNode>;
const nodes: TreeNode[] = [
  {
    key: 'foo',
    label: 'bar',
    children: []
  }
];
```

yusufkandemir:

- Bugfix

- Yes (type-only)

**! This breaks the previous usage, so the users should migrate to the new usage. See below for more information !**

**Other information:**
Resolves #15351.

Example use case: Add `{ key: string }` to the node type.

Previous usage to achieve expected behavior (isn't an obvious solution):
```ts
type TreeNode = { key: string } & QTreeNode<TreeNode>;
const nodes: TreeNode[] = [
  {
    key: 'foo',
    label: 'bar',
    children: []
  }
];
```
New usage:
```ts
const nodes: QTreeNode<{ key: string }>[] = [
  {
    key: 'foo',
    label: 'bar',
    children: []
  }
];
```

### DIFF
--- a/ui/types/api/qtree.d.ts
+++ b/ui/types/api/qtree.d.ts
@@ -1,7 +1,7 @@
 /**
  * @template TExtra Object type to add extra properties for the node, overrides the existing ones as well
  */
-export type QTreeNode<TExtra = { [index: string]: any }> = Omit<
+export type QTreeNode<TExtra = unknown> = Omit<
   {
     label?: string;
     icon?: string;
@@ -20,9 +20,9 @@ export type QTreeNode<TExtra = { [index: string]: any }> = Omit<
     header?: string;
     body?: string;
   },
-  keyof TExtra
+  unknown extends TExtra ? "" : keyof TExtra
 > &
-  TExtra;
+  (unknown extends TExtra ? Record<string, any> : TExtra);
 
 export interface QTreeLazyLoadParams<
   Node extends QTreeNode = QTreeNode,
--- a/ui/types/api/qtree.d.ts
+++ b/ui/types/api/qtree.d.ts
@@ -1,5 +1,95 @@
 /**
+ * Node type to be used with QTree's `nodes` prop
+ *
+ * @see https://v2.quasar.dev/vue-components/tree#qtree-api
+ *
  * @template TExtra Object type to add extra properties for the node, overrides the existing ones as well
+ *
+ * @example
+ * Basic usage
+ * ```ts
+ * const nodes: QTreeNode[] = [
+ *   // ...
+ * ];
+ * // <q-tree :nodes="nodes" />
+ * ```
+ *
+ * @example
+ * Making some properties required
+ * ```ts
+ * // make label and icon required, the rest will stay optional
+ * const nodes: QTreeNode<{ label: string; icon: string }>[] = [
+ *   // ...
+ * ];
+ * ```
+ *
+ * @example
+ * Adding extra properties
+ * ```ts
+ * // on top of the existing properties, add a foo property
+ * const nodes: QTreeNode<{ foo: number }>[] = [
+ *   // ...
+ * ];
+ * ```
+ *
+ * @example
+ * Using different label/children properties
+ * ```ts
+ * type Node = QTreeNode<{ name: string; subNodes: Node[] }>;
+ * const nodes: Node[] = [
+ *   // ...
+ * ];
+ * // <q-tree :nodes="nodes" label-key="name" children-key="subNodes" />
+ * ```
+ *
+ * @example
+ * Using a different child node type
+ * ```ts
+ * type ChildNode = QTreeNode<{ foo: number }>;
+ * type ParentNode = QTreeNode<{ bar: string; children?: ChildNode[] }>;
+ *
+ * const nodes: ParentNode[] = [
+ *   // ...
+ * ];
+ * ```
+ *
+ * @example
+ * A very basic file system tree
+ * ```ts
+ * interface FileInfo {
+ *   path: string;
+ *   size: number;
+ *   lastModified: number;
+ * }
+ * type FileNode = QTreeNode<FileInfo & { type: "file", children?: never }>;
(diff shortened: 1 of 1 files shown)


ui/types/api/qtree.d.ts
```diff
@@ -1,24 +1,28 @@
...
+ * @template TExtra Object type to add extra properties for the node, overrides the existing ones as well
+ */
+export type QTreeNode<TExtra = unknown> = Omit<
+  {
+    label?: string;
+    icon?: string;
+    iconColor?: string;
+    img?: string;
+    avatar?: string;
+    children?: QTreeNode<TExtra>[];
```
IlCallo, line 101: This implementation change will break the use case where you have children which have different properties than the parent, or am I incorrect?
Maybe we can add 2 generics, one for the children Node type (which defaults to the same type of the parent) and one for extra properties on the current node

rstoenescu:
Haven't looked at the changes from this PR, but a few things to keep in mind:
1. The nodes can contain any property, even beyond what is required
2. There are props that can change Quasar's requirements. There may be scenarios where the props below are used so that the "standard" props can then contain whatever.
  
image

IlCallo:
> The nodes can contain any property, even beyond what is required

Yes, this is accomplished by this last bit: `[...] & (unknown extends TExtra ? Record<string, any> : TExtra)`
If you provide the generic parameter, you get full type safety, otherwise you get to put in any kind of property

> There are props that can change Quasar's requirements. There may be scenarios where the props below are used so that the "standard" props can then contain whatever.

I think the `Omit<..., unknown extends TExtra ? "" : keyof TExtra>`, plus the previous bit, takes care of this
And the fact that all well-known properties are optional by default
If you provide a `TExtra`, you can provide your "shape" of the additional fields for your use case
We could add more generics to account for the props you highlighted, but I don't think it's worth right now

IlCallo:
> The nodes can contain any property, even beyond what is required

Yes, this is accomplished by this last bit: `[...] & (unknown extends TExtra ? Record<string, any> : TExtra)`
If you provide the generic parameter, you get full type safety, otherwise you get to put in any kind of property

> There are props that can change Quasar's requirements. There may be scenarios where the props below are used so that the "standard" props can then contain whatever.

I think the `Omit<..., unknown extends TExtra ? "" : keyof TExtra>`, plus the previous bit, takes care of this
And the fact that all well-known properties are optional by default
If you provide a `TExtra`, you can provide your "shape" of the additional fields for your use case
We could add more generics to account for the props you highlighted, but I don't think it's worth right now

ui/types/api/qtree.d.ts
yusufkandemir: It is possible to override the `children` type like this:
```ts
type ChildNode = QTreeNode<{ foo: string }>
type ParentNode = QTreeNode<{ bar: number; children?: ChildNode[] }>
```
Sure, we can add another generic type parameter, but I think the existing solution is simpler and more flexible. You can make the children required this way, or even use a tuple type to have a known child sequence, etc.
```ts
type ChildNode = QTreeNode<{ foo: string }>
// required children
type ParentNode = QTreeNode<{ bar: number; children: ChildNode[] }>
```
```ts
type ChildNode1 = QTreeNode<{ foo: string }>
type ChildNode2 = QTreeNode<{ baz: string }>
// required children with specific children count/order/types
type ParentNode = QTreeNode<{ bar: number; children: [ChildNode1, ChildNode2] }>
```

yusufkandemir: Here is an example for representing the basics of a file system:
```ts
type File = {
  path: string;
  size: number;
  lastModified: number;
}
type FileNode = QTreeNode<File & { type: "file", children?: never }>;
type DirectoryNode = QTreeNode<File & { type: "directory", children?: (FileNode | DirectoryNode)[] }>;

const directory: DirectoryNode = {
  type: "directory",
  path: "/",
  size: 0,
  lastModified: 0,
  // allows both file and directory as children
  children: [
    {
      type: "file",
      path: "/foo.txt",
      size: 100,
      lastModified: 1000,
      // does not allow children
    },
    {
      type: "directory",
      path: "/bar",
      size: 0,
      lastModified: 0,
      // empty folder - doesn't have children
    }
  ]
}
```

IlCallo: That's actually super smart and flexible, but not so clear
Can we add some documentation about this? Or at least add these examples in the original issue

Documentation aside, LGTM at this point

yusufkandemir: Another advantage is allowing children to type when using a different children key. The `children` property can be different(e.g. `nodes`), which can be modified through `children-key`. So, this solution also allows easily adding the children types with different children property names as well.

yusufkandemir: Added some documentation with examples in the form of JSDocs

Issue #16586, opened 2023-11-14
Rounded QCircularProgress should take into account behaviour of stroke-linecap


hazzik:
### What happened?

When `QCircularProgress` with `rounded` attribute at near the max value then the circle is closed on wide lines.

![image](https://github.com/quasarframework/quasar/assets/144791/a89b5b93-4a27-48d1-9204-9045ec7134a8)

### What did you expect to happen?

I expect the component take into account that `stroke-linecap` attribute _adds_ half circles at the edges of the line, so when calculating the length of the stroke the line height (stroke-width) must be considered and subtracted from the length.

### Reproduction URL

https://codepen.io/hazzik/pen/JjxymYV?editors=1010

### How to reproduce?

1. Open codepen

### Flavour

Quasar CLI with Webpack (@quasar/cli | @quasar/app-webpack)

### Areas

Components (quasar)

hazzik:
This is the code pen that should explain what I mean: https://codepen.io/hazzik/pen/QWYMzKy


![image](https://github.com/quasarframework/quasar/assets/144791/8197d871-a8fc-453a-9d9d-7c84ec22976d)
Calculated for 98%:
```html
<svg width="125" height="125" viewBox="0 0 125 125" 
     fill="none" xmlns="http://www.w3.org/2000/svg"
     style="transform: rotate3d(0, 0, 1, -90deg);"
     >
  
  <circle 
          cx="55"
          cy="55"
          r="50" 
          stroke="#E6E8E7"
          stroke-width="10"
  />
  <circle 
          cx="55"
          cy="55"
          r="50" 
          stroke="red"
          stroke-dasharray="314.159"
          stroke-dashoffset="6.283"
          stroke-width="10"
  />
  <circle 
          cx="55"
          cy="55"
          r="50" 
          stroke="#00965F"
          stroke-dasharray="297.876 16.283"
          stroke-dashoffset="-5"
          stroke-linecap="round"
          stroke-width="10"
  />
</svg>

pdanpdan:
And how would you solve the near 100% values. Because you would need to change the cap curve or have a non-continuous 100%

hazzik:
@pdanpdan, the SVG above demonstrating what I want to achieve. The red outline is the current unrounded progress, in green proposed solution.

In essence, the stroke is offset by negative half of stroke width.
The stroke array will consist of two elements: line and a gap.
Line is `max(circumference * value - strokeWidth, 0)` 
Gap is `circumference - line`

[...]

PR #16593, opened 2023-11-15
fix: #16586 account for stroke-linecap behavior for rounded QCircularProgress

hazzik:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/components/circular-progress/QCircularProgress.js
+++ b/ui/src/components/circular-progress/QCircularProgress.js
@@ -36,13 +36,18 @@ export default createComponent({
     const { proxy: { $q } } = getCurrentInstance()
     const sizeStyle = useSize(props)
 
+    const viewBox = computed(() => diameter / (1 - props.thickness / 2))
+
+    const strokeWidth = computed(() => props.thickness / 2 * viewBox.value)
+
     const svgStyle = computed(() => {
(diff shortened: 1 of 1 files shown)


pdanpdan:
I would go with:
- drop the changes for angle in svgStyle
- starting with 75% I would progressively increase the gap until it reaches 100%

```js
const dashRatio = (props.max - normalized.value) / (props.max - props.min)
const dashGap = props.rounded === true && normalized.value < props.max && dashRatio < 0.25
  ? strokeWidth.value / 2 * (1 - dashRatio / 0.25)
  : 0
return circumference * dashRatio + dashGap
```

That way the last position where a wrong angle is observable (75%) is good, and then the spacing increases linearly to be strokeWidth/2 at 99%

hazzik:
Thanks for the feedback @pdanpdan. Yeah, if the ratio works for you, I dont mind at all. The angle change is completely optional, the problem without it is that the gap start is a little bit tilted, so I was thinking it looks nicer this way. What I was trying to achieve with untilting is when you switch between rounded and unrounded the line visible edges stay the same.

rstoenescu:
Left the strokeDashOffset computation change and reverted the svg angle one.
This will go into Quasar v2.14.1

Thank you for contributing!

PR #16615, opened 2023-11-20
(fix: QSlider): slider value should start from `innerMin` instead of `min` on drag #16614

### LINKED ISSUE #16614: [QSlider] Step starts from min instead of inner min, resulting in wrong stepping if using floats
### What happened?

The slider starts at 4$, when clicking anywhere on it, you get steps always ending with .8

Since `innerMin` is the actual minimum selectable value, the steps should be starting from that instead and be whole numbers

### Reproduction URL

https://codepen.io/Tofandel/pen/NWoymZm?editors=101

### How to reproduce?

1. Go to the reproduction link
2. Click anywhere on the slider
3. Notice the value is `x.8 $` and it doesn't match with any of the marker label steps
4. If you click on 3, the slider caps it to inner min and stays at `3 $`
5. If you use the keyboard arrows, then you get the correct behavior

### Flavour

### Areas

Components (quasar)

### Platforms/Browsers

Firefox, Chrome, Safari, Microsoft Edge, iOS, Android, Electron

### Quasar info output

_No response_

### Relevant log output

_No response_

### Additional context

The issue seems to be coming from this function which uses `min` and `max` instead of `innerMin` and `innerMax`
https://github.com/quasarframework/quasar/blob/dev/ui/src/components/slider/use-slider.js#L236-L250

Tofandel:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

I unit tested the calculation which produces the desired result but I did not test the actual behavior in a browser, as this method is used in some other places, it should definitely be tested but I'm rather confident it should work

Fixes #16614

### DIFF
--- a/ui/src/components/slider/use-slider.js
+++ b/ui/src/components/slider/use-slider.js
@@ -235,7 +235,7 @@ export default function ({ updateValue, updatePosition, getDragging, formAttrs }
 
   function convertRatioToModel (ratio) {
     const { min, max, step } = props
-    let model = innerMin.value + (ratio - innerMinRatio.value) * (max - min)
+    let model = min + ratio * (max - min)
 
     if (step > 0) {
       const modulo = (model - innerMin.value) % step
--- a/ui/src/components/slider/use-slider.js
+++ b/ui/src/components/slider/use-slider.js
@@ -235,7 +235,7 @@ export default function ({ updateValue, updatePosition, getDragging, formAttrs }
 
   function convertRatioToModel (ratio) {
     const { min, max, step } = props
-    let model = innerMin.value + ratio * (max - min)
+    let model = innerMin.value + (ratio - innerMinRatio.value) * (max - min)
 
     if (step > 0) {
       const modulo = (model - innerMin.value) % step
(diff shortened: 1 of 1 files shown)


ui/src/components/slider/use-slider.js
```diff
@@ -235,10 +235,10 @@ export default function ({ updateValue, updatePosition, getDragging, formAttrs }
 
   function convertRatioToModel (ratio) {
     const { min, max, step } = props
-    let model = min + ratio * (max - min)
+    let model = innerMin.value + (ratio - innerMinRatio.value) * (max - min)
```
pdanpdan: Why are you doing this?

Tofandel: The calculation will return the same result before and after, but this avoid floating point operation issues, if we have a float as min and a ratio also as float we end up with a float that is not precise enough eg 3.999999
So by using innerMin directly and subtracting it from the ratio which turns up exactly the same in arithmetic we can avoid the floating point precision issue

Maybe it's already countered by the modulo but it would need some thorough unit test cases to test

rstoenescu: Hi, but the same happens when having a floating point value for innermin. While it is an improvement for one case, it's the exact opposite for the other. And it requires an additional subtraction operation, so this change should be reverted.

rstoenescu:
Reversed the initial model computation as per the discussion above and merged.
Thank you for contributing!

PR #16636, opened 2023-11-27
test(QDialog): add QDialog tests

n05la3:

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/components/dialog/__tests__/DialogWrapper.vue
+++ b/ui/src/components/dialog/__tests__/DialogWrapper.vue
@@ -1,21 +1,13 @@
 <template>
-  <div data-cy="dialog-page">
-    <q-dialog ref="dialogRef" v-bind="$attrs" data-cy="dialog-form">
-      <q-card data-cy="dialog-card">
-        <q-card-section>
-          <div class="text-h6">
-            Dialog section
-          </div>
-        </q-card-section>
+<q-dialog ref="dialogRef" v-bind="$attrs">
+  <q-card>
+    <q-card-section>
+      <q-btn label="Dialog Button" data-cy="dialog-button"/>
+    </q-card-section>
+  </q-card>
+</q-dialog>
 
-        <q-card-section>
-          <q-btn label="Dialog Button" data-cy="dialog-button"/>
-        </q-card-section>
-      </q-card>
-    </q-dialog>
-
-    <q-input data-cy="input-field"/>
-  </div>
+<q-input data-cy="input-field"/>
 </template>
 
 <script>
--- a/ui/src/components/dialog/__tests__/DialogWrapper.vue
+++ b/ui/src/components/dialog/__tests__/DialogWrapper.vue
@@ -1,0 +1,43 @@
+<template>
+  <div data-cy="dialog-page">
+    <q-dialog ref="dialogRef" v-bind="$attrs" data-cy="dialog-form">
+      <q-card data-cy="dialog-card">
+        <q-card-section>
+          <div class="text-h6">
+            Dialog section
+          </div>
+        </q-card-section>
+
+        <q-card-section>
+          <q-btn label="Dialog Button" data-cy="dialog-button"/>
+        </q-card-section>
+      </q-card>
+    </q-dialog>
+
+    <q-btn data-cy="close-button" label="Open Dialog"/>
+    <q-input data-cy="input-field"/>
+  </div>
+</template>
+
+<script>
+import { defineComponent, ref } from 'vue'
+
+export default defineComponent({
+  inheritAttrs: false,
+  setup () {
+    const dialogRef = ref(null)
+
+    function focus () {
+      console.log(dialogRef.value.focus, 'Attempting to focus')
+      dialogRef.value.focus()
+    }
+
+    function shake () {
+      dialogRef.value.shake()
+    }
+
+    return { dialogRef, focus, shake }
+  }
+})
+
+</script>
--- a/ui/src/components/dialog/__tests__/DialogWrapper.vue
+++ b/ui/src/components/dialog/__tests__/DialogWrapper.vue
@@ -14,7 +14,6 @@
       </q-card>
     </q-dialog>
 
-    <q-btn data-cy="close-button" label="Open Dialog"/>
     <q-input data-cy="input-field"/>
   </div>
 </template>
@@ -28,7 +27,6 @@ export default defineComponent({
     const dialogRef = ref(null)
 
     function focus () {
-      console.log(dialogRef.value.focus, 'Attempting to focus')
       dialogRef.value.focus()
     }
 
--- a/ui/src/components/dialog/__tests__/QDialog.cy.js
+++ b/ui/src/components/dialog/__tests__/QDialog.cy.js
@@ -71,24 +71,34 @@ describe('Dialog API', () => {
           const model = ref(true)
           mountQDialogWrapper({
             props: {
-              ...vModelAdapter(model),
-              noBackDropClose: true
+              ...vModelAdapter(model)
             }
           })
 
-          getHostElement()
-            .should('exist')
-            .then(() => {
-              closeDialogViaBackdrop()
-              getHostElement()
-                .should('exist')
-                .then(() => {
-                  Cypress.vueWrapper.setProps({ noBackDropClose: false })
-                })
+          cy.dataCy('dialog-page').then(() => {
+            Cypress.vueWrapper.setProps({ noBackdropDismiss: false })
+            model.value = true
 
-              closeDialogViaBackdrop()
-              getHostElement().should('not.exist')
-            })
+            getHostElement()
+              .should('exist')
+              .then(() => {
+                closeDialogViaBackdrop()
+                getHostElement()
+                  .should('not.exist')
+              })
+          })
+          cy.dataCy('dialog-page').then(() => {
+            Cypress.vueWrapper.setProps({ noBackdropDismiss: true })
+            model.value = true
+
+            getHostElement()
+              .should('exist')
+              .then(() => {
+                closeDialogViaBackdrop()
+                getHostElement()
+                  .should('exist')
+              })
+          })
         })
       })
 
--- a/ui/src/components/dialog/__tests__/QDialog.cy.js
+++ b/ui/src/components/dialog/__tests__/QDialog.cy.js
@@ -6,10 +6,6 @@ function mountQDialogWrapper (options) {
   return cy.mount(DialogWrapper, options)
 }
 
-function getHostElement () {
-  return cy.dataCy('dialog-card')
-}
-
 function closeDialogViaBackdrop () {
   return cy.get('.q-dialog__backdrop').click({ force: true })
 }
@@ -29,7 +25,7 @@ function closeDialogViaEscKey () {
 // TODO: take into consideration that dialogs without backdrop have always 'aria-modal' false
 // Check if there's a more reliable way to check for persistent dialogs
 function assertPersistentDialogExists () {
-  cy.dataCy('dialog-form').should('not.have.attr', 'aria-modal', 'false')
+  cy.get('.q-dialog__inner').should('not.have.attr', 'aria-modal', 'false')
 }
 
 describe('Dialog API', () => {
@@ -271,25 +267,35 @@ describe('Dialog API', () => {
             }
           })
 
-          cy.dataCy('dialog-form')
-            .should('have.class', 'q-dialog--seamless')
+          cy.withinDialog(() => {
+            cy.root()
+              .closest('.q-dialog')
+              .should('have.class', 'q-dialog--seamless')
 
-          cy.dataCy('input-field')
-            .should('be.visible')
-            .then(async () => {
-              await Cypress.vueWrapper.setProps({ seamless: false })
-            })
+            cy.root()
+              .closest('body')
+              .dataCy('input-field')
+              .should('be.visible')
+              .then(async () => {
+                await Cypress.vueWrapper.setProps({ seamless: false })
+              })
 
-          cy.dataCy('dialog-form')
-            .should('not.have.class', 'q-dialog--seamless')
-          cy.dataCy('input-field')
-            .should('not.be.visible')
+            cy.root()
+              .closest('.q-dialog')
+              .should('not.have.class', 'q-dialog--seamless')
+            cy.root()
+              .closest('body')
+              .dataCy('input-field')
+              .should('not.be.visible')
+
+            closeDialogViaBackdrop()
+          })
         })
       })
 
       describe('(prop): maximized', () => {
         it('should maximize the dialog', () => {
-          const model = ref(false)
+          const model = ref(true)
           mountQDialogWrapper({
             props: {
               ...vModelAdapter(model),
@@ -297,30 +303,31 @@ describe('Dialog API', () => {
             }
           })
 
-          cy.dataCy('dialog-page').then(() => {
-            model.value = true
+          cy.withinDialog(() => {
+            cy.root()
+              .closest('.q-dialog')
+              .get('.q-dialog__inner--maximized')
+            cy.root()
+              .closest('body')
+              .dataCy('input-field')
+              .should('not.be.visible')
+              .then(async () => {
+                await Cypress.vueWrapper.setProps({ maximized: false })
+              })
 
-            getHostElement().then(() => {
-              cy.dataCy('dialog-form')
-                .get('.q-dialog__inner--maximized')
-                .should('exist')
-              cy.dataCy('input-field')
-                .should('not.be.visible')
-                .then(() => {
-                  Cypress.vueWrapper.setProps({ maximized: false })
-                })
+            cy.root()
+              .closest('.q-dialog')
+              .get('.q-dialog__inner--maximized')
+              .should('not.exist')
 
-              cy.dataCy('dialog-form')
-                .get('.q-dialog__inner--maximized')
-                .should('not.exist')
-            })
+            closeDialogViaBackdrop()
           })
         })
       })
 
       describe('(prop): full-width', () => {
         it('should use a full-width for the dialog', () => {
-          const model = ref(false)
+          const model = ref(true)
           mountQDialogWrapper({
             props: {
               ...vModelAdapter(model),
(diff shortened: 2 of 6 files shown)


ui/src/components/dialog/__tests__/DialogWrapper.vue
```diff
@@ -0,0 +1,43 @@
...
+<script>
+import { defineComponent, ref } from 'vue'
+
+export default defineComponent({
+  inheritAttrs: false,
+  setup () {
+    const dialogRef = ref(null)
+
+    function focus () {
+      console.log(dialogRef.value.focus, 'Attempting to focus')
```
IlCallo: I guess this is a leftover

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -1,21 +1,82 @@
+import FieldWrapper from './DialogWrapper.vue'
```
IlCallo: Wrong import name, should be `DialogWrapper`

ui/src/components/dialog/__tests__/DialogWrapper.vue
```diff
@@ -0,0 +1,43 @@
...
+          </div>
+        </q-card-section>
+
+        <q-card-section>
+          <q-btn label="Dialog Button" data-cy="dialog-button"/>
+        </q-card-section>
+      </q-card>
+    </q-dialog>
+
+    <q-btn data-cy="close-button" label="Open Dialog"/>
```
IlCallo: I guess you're not really using this one, since it's not referenced and the label and data-cy are opposites

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -1,21 +1,82 @@
...
+          const model = ref(true)
+          mountQDialogWrapper({
+            props: {
+              ...vModelAdapter(model),
+              persistent: false
+            }
+          })
+
+          cy.withinDialog(() => {
+            closeDialog() // Closes dialog by clicking backdrop
```
IlCallo: I'd just rename the function as `closeDialogViaBackdrop`

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -1,21 +1,82 @@
...
+          const model = ref(true)
+          mountQDialogWrapper({
+            props: {
+              ...vModelAdapter(model),
+              noEscDismiss: true
+            }
+          })
+
+          getHostElement().then(() => {
+            getHostElement().trigger('keydown', { keyCode: 27 })
```
IlCallo: Shouldn't this be `cy.get('body').type('{esc}')` too?
Why using keydown here?

ui/src/components/dialog/__tests__/QDialog.cy.js
IlCallo: This test should check that the Dialog cannot be closed either via ESC, nor backdrop, nor route navigation

I'd also avoid testing the `persistent: false` behaviour, since that will be tested in the other 3 tests about the specific props (`no-esc-dismiss`, etc)

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -1,21 +1,82 @@
...
+          const model = ref(true)
+          mountQDialogWrapper({
+            props: {
+              ...vModelAdapter(model),
+              noBackDropClose: true
+            }
+          })
+
+          cy.withinDialog(() => {
+            closeDialog()
```
IlCallo: Using it here is an even harder hint that this should be named `closeDialogViaBackdrop`

IlCallo (changes requested):
Really nice work!
Aside some dead code around and a couple readability fixes, everything seems fine

ui/src/components/dialog/__tests__/QDialog.cy.js
n05la3: Fixed.

n05la3:
I have resolved the issues with the PR. 

As concerns the `use-model-toggle`, it was a race condition that caused the issues. After digging, I noticed inconsistencies between the tests and the code when testing the `hide` event. The hide function in `QMenu.js` looks something like this:
```ts
function handleHide (evt) {
  hidePortal() // The dialog is hidden

  // Some code here. It's how fast this happens that determines if the test will pass or fail

  emit('hide', evt)
}
```
While in the test looks something like this:
```ts
cy.dataCy('method-hide')
  .click({ force: true })
cy.dataCy('menu')
  .should('not.exist')
  .then(() => {
    expect(fn).to.be.called
  })
```
- First, there's no guarantee that the click is completed before checking if the menu no longer exists, this therefore creates a race condition and depends on how fast the hide method resolves. Normally, it should check that the menu existed before running the hide method. Otherwise, it will just depend on how fast things happen to pass or fail. In addition to this problem:
- Checking if menu no longer exists is not the right thing to do because the `hidePortal` gets called the `hide` event is emitted. So, the menu may disappear before the event is emitted. This should have been passing because things were happening so fast.

With the `QDialog`, it mainly needed to ensure that the dialog existed before continuing the tests. A few tests were flaky too and only passed depending on how fast things were happening. We had tests like this:

```ts
 cy.dataCy('dialog-page').then(() => {
 1)   cy.dataCy('input-field').focus()
 2)  model.value = true
  
    cy.withinDialog(() => {
      closeDialog()
    })
    cy.dataCy('input-field').should('have.focus')
})
 ```
It has a similar problem to the `q-toggle-model`. When run (1), we are not sure whether or not it succeeds. For instance, in some cases, after running (2), the backdrop will cover the page before completing the `focus`. So (1) may silently fail and the tests after that are false positives. Instead, this has been replaced with the more stable counterpart below:
```ts
cy.dataCy('input-field')
    .type('Hello')
  cy.dataCy('input-field')
    .should('have.value', 'Hello').then(() => {
      model.value = true

      cy.withinDialog(() => {
        closeDialogViaBackdrop()
      })
      getHostElement()
        .should('not.exist').then(() => {
          cy.dataCy('input-field')
            .should('have.focus')
        })
    })
```
I have made similar changes to the rest of `QDialog` tests to improve stability.

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -1,21 +1,131 @@
...
+
+function getHostElement () {
+  return cy.dataCy('dialog-card')
+}
+
+function closeDialogViaBackdrop () {
+  return cy.get('.q-dialog__backdrop').click({ force: true })
+}
+
+// TODO: this only works in `withinDialog` context,
```
IlCallo, line 13: Check out this TODO
If you find a way to make `closeDialogViaEscKey` work both inside and outside a `within` context, that would be cool

I tried with `cy.window()` and `cy.document()` already, which I recalled worked fine, but it seems like they changed that

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -1,21 +1,131 @@
...
+// as when using it outside it `.root()` will yeld "html" tag
+// and Cypress won't be able to find anything with `.closest('body')`
+// since it only searches upwards
+function closeDialogViaEscKey () {
+  // Official way to escape `within` context
+  // https://docs.cypress.io/api/commands/within#Temporarily-escape
+  return cy.root().closest('body').type('{esc}')
+}
+
+// TODO: we'll add this check to the AE helper to automatically test persistent dialogs
```
IlCallo, line 23: This is okeish, but it doesn't account for seamless dialogs and dialogs without a backdrop
If we can find some other way to test for persistent dialogs being there, we can execute it automatically whenever a `withinDialog` command is called with `persistent: true`

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -1,21 +1,131 @@
...
+            fn: () => {
+              closeDialogViaEscKey()
+            }
+          })
+
+          assertPersistentDialogExists()
+
+          cy.wrap()
+            .then(async () => {
+              // Care as you need to wait setProps to complete its execution
```
IlCallo, line 85: `Cypress.vueWrapper.setProps()` is an async function
If you just call it, without marking the `then` callback as `async`, Cypress wouldn't await for it to complete before proceeding with other commands, possibly breaking the test or causing a race condition

Check out if VTU or Cypress already provide an helper which minimize this category of errors
If not, you can try to create your own which automate the `cy.wrap().then(async () => {})` bit, even if the user would then still need to manually await its promises

If it's generic enough, we can add it to Cypress AE later on

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -1,21 +1,131 @@
...
+              // before moving on or the tests will fail due to the race condition
+              await Cypress.vueWrapper.setProps({ noEscDismiss: false })
+            })
+
+          cy.withinDialog(() => {
```
IlCallo, line 90: Try to use `cy.withinDialog` whenever possible, as it automates some checks about Dialog existence and not existence for you

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -26,103 +136,425 @@ describe('Dialog API', () => {
...
+            }
+          })
+
+          // Test the input field regaining focus
+          cy.dataCy('input-field')
+            .focus()
+          cy.dataCy('input-field')
+            .then(async () => {
+              model.value = true
+              await nextTick()
```
IlCallo: I'm not sure `nextTick` is actually required when changing the refs, to make sure there's no race condition, check it out / test it / search for confirmation about this (I guess on VTU repo)

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -26,103 +136,425 @@ describe('Dialog API', () => {
...
+        it('should put the dialog in a seamless state', () => {
+          const model = ref(true)
+          mountQDialogWrapper({
+            props: {
+              ...vModelAdapter(model),
+              seamless: true
+            }
+          })
+
+          cy.dataCy('dialog-form')
```
IlCallo: Could this be done using `withinDialog` and `cy.root()` within it?
Not sure, check it out

In general, it's not totally clear to me how we should deal with seamless dialogs

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -26,103 +136,425 @@ describe('Dialog API', () => {
...
+            })
+
+          cy.dataCy('dialog-form')
+            .should('not.have.class', 'q-dialog--seamless')
+          cy.dataCy('input-field')
+            .should('not.be.visible')
         })
       })
 
       describe('(prop): maximized', () => {
```
IlCallo, line 294: I stopped my refactoring here, continue and re-check out again the model-toggle composable changes

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -26,103 +136,425 @@ describe('Dialog API', () => {
...
+        it('should maximize the dialog', () => {
+          const model = ref(false)
+          mountQDialogWrapper({
+            props: {
+              ...vModelAdapter(model),
+              maximized: true
+            }
+          })
+
+          cy.dataCy('dialog-page').then(() => {
```
IlCallo: Using `.get` with `dialog-page` shouldn't be needed, as you're assured for it to exist after mount is executed

ui/src/components/dialog/__tests__/QDialog.cy.js
```diff
@@ -26,103 +136,425 @@ describe('Dialog API', () => {
...
+            model.value = true
+
+            getHostElement().then(() => {
+              cy.dataCy('dialog-form')
+                .get('.q-dialog__inner--maximized')
+                .should('exist')
```
IlCallo: Avoid `should('exist')`, if not to convey an explicit message
Rely on `withinDialog` whenever possible, or on the implicit check hidden in `.get` commands

ui/src/composables/private/__tests__/use-model-toggle.cy.js
```diff
@@ -165,39 +165,53 @@ describe('use-model-toggle API', () => {
...
             expect(fn).not.to.be.called
           })
-        cy.get('body')
-          .click(499, 0)
+
         cy.dataCy('menu')
-          .should('not.exist')
-          .then(() => {
-            expect(fn).to.be.called
+          .should('exist').then(() => {
```
IlCallo: Remove `.should('exist')` where you added it, as it's implicit in `.dataCy` which is based upon `.get`

ui/src/composables/private/__tests__/use-model-toggle.cy.js
```diff
@@ -165,39 +165,53 @@ describe('use-model-toggle API', () => {
...
+            cy.get('body')
+              .click(499, 0)
```
IlCallo: If possible, add a comment explaining these numbers

ui/src/components/dialog/__tests__/DialogWrapper.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+  <div data-cy="dialog-page">
+    <q-dialog ref="dialogRef" v-bind="$attrs" data-cy="dialog-form">
```
IlCallo: `dialog-form` doesn't seem a good name for this
`dialog-container` maybe? Or just `dialog`? Or whatever

If you can avoid referencing this altogether in tests, that would be even better

IlCallo (changes requested):
I left some comments to highlight some changes I did and explain to which pitfalls you need to pay attention

I didn't complete the refactor, but I marked with a comment up until where I did it
Good luck

ui/src/components/dialog/__tests__/QDialog.cy.js
n05la3: In this particular case, `nextTick` is required. According to VTU docs, if we await `setProps`, `setValue`, etc then it is equivalent to having called the `await Vue.nextTick()` after the command, just as you noted in one of your comments.  I noticed that the `vModelAdapter` implementation does not await when calling `Cypress.vueWrapper.setProps({ [modelName]: value })`. So in this case we need `nextTick` to ensure the view has updated.

ui/src/components/dialog/__tests__/QDialog.cy.js
n05la3: I ended up using `cy.root()` within `withinDialog`

ui/src/components/dialog/__tests__/QDialog.cy.js
n05la3: What do you think about this?
```ts
function getPreviousElement (fn) {
  return cy.wrap()
    .then(async () => {
      await fn()
    })
}
```
I wrote something like this but did not find any real benefit to using it instead of just calling `cy.wrap`. Even writing it as `cy.getPrevious()` will still end up with a similar amount of typing except it is more meaningful. It may be interesting to have this in the AE, but I don't think it's worth it here. What do you think?

ui/src/components/dialog/__tests__/QDialog.cy.js
n05la3: Checked the DOM again with these props and could not find any better way to identify persistent dialogs. We could probably add `cy.get('.q-dialog').should('not.have.class', 'q-dialog--seamless')` too but it is no guarantee that the persistent prop was set. So I left it as it is.

ui/src/components/dialog/__tests__/QDialog.cy.js
n05la3: Could not find a better way to handle this either.

ui/src/components/dialog/__tests__/QDialog.cy.js
IlCallo: Shall we update `vModelAdapter` to await for `setProps` then?
I'm not sure the interaction would work as expected, due to the use of reactivity we do in tests, but we could try that

ui/src/components/dialog/__tests__/QDialog.cy.js
IlCallo: The name feels strange, why using "previous" or "previous element"?
`cy.wrap` doesn't actually touch in any way anything "previous" or an element

ui/src/components/dialog/__tests__/QDialog.cy.js
IlCallo: Ok, we can't add this automatically into the Cypress AE helper then, but we can document this pattern and its limitations into the AE README, until we find a proper solution

ui/src/components/dialog/__tests__/QDialog.cy.js
IlCallo: Ok, let's keep the TODO and see if we are able to spend some more time on this in a follow up PR

IlCallo: Documented these new helpers and their limitations here: https://github.com/quasarframework/quasar-testing/issues/355

ui/src/components/dialog/__tests__/QDialog.cy.js
IlCallo: Documented these new helpers and their limitations here: https://github.com/quasarframework/quasar-testing/issues/355

ui/src/components/dialog/__tests__/QDialog.cy.js
n05la3: Yes, we should update `vModelAdapter` and test that it does not break any existing tests. I do not foresee it having any particular issues with existing tests. I'll update and check this out.

ui/src/components/dialog/__tests__/QDialog.cy.js
n05la3: Agreed. `cy.getPrevious` is not a good name here. Perhaps something like `getFulfilled`, `wrapAndFulFill`, `wrapAndThen` or something better in this case should be a better name. However, like I said in my previous comment, I don't think it serves some real benefit adding it here. However, I think it could be good in the AE. What do you think?

IlCallo: If you didn't find it useful and we're not using it at least in our own tests, then let's avoid adding it in the AE for now

Let's try to add only what we know is actually useful

ui/src/components/dialog/__tests__/QDialog.cy.js
IlCallo: Go on and test this

n05la3: Tested the edit and it does not break any tests. I have opened a PR [here](https://github.com/quasarframework/quasar-testing/pull/364)

PR #16660, opened 2023-12-04
Better handling of notification timeouts

rotu:
parseInt is not good for coercion.
* Infinity gets implicitly converted to NaN.
* Large numbers e.g. `5e+30`, get truncated (to 5).

- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/plugins/Notify.js
+++ b/ui/src/plugins/Notify.js
@@ -116,7 +116,7 @@ function addNotification (config, $q, originalApi) {
     notif.timeout = 5000
   }
   else {
-    const t = parseInt(notif.timeout, 10)
+    const t = Number(notif.timeout)
     if (isNaN(t) || t < 0) {
       return logError('wrong timeout', config)
     }
--- a/ui/src/plugins/Notify.js
+++ b/ui/src/plugins/Notify.js
@@ -55,7 +55,7 @@ const notifTypes = {
 
   ongoing: {
     group: false,
-    timeout: 0,
+    timeout: Infinity,
     spinner: true,
     color: 'grey-8'
   }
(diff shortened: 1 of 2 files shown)


rstoenescu:
Hi,

Thank you for contributing!
The problem with the new approach is that `Number(null) === 0`. May be other problems as well. Can you go a bit deeper with the PR pls?

One thing to mention is that there is no need for Infinity as the value 0 means exactly that (there's no timeout; don't use one). But I see the value in correctly parsing numbers in `5e+3` notation.

rotu:
> Hi,
> 
> Thank you for contributing! The problem with the new approach is that `Number(null) === 0`. May be other problems as well. Can you go a bit deeper with the PR pls?

I don't think that's a problem - in JavaScript, `null` coerces to `0` in expressions like `3 + null`.

If you want it to reject `null` as a value because we're enforcing types, then I think maybe it should reject string values as well?

> One thing to mention is that there is no need for Infinity as the value 0 means exactly that (there's no timeout; don't use one). But I see the value in correctly parsing numbers in `5e+3` notation.

You are correct. This is a little underdocumented, so I will update the docs as well.

I *do* disagree with the choice in API, though, and I've updated my PR accordingly.

The issue with `0` as a sentinel value is that it's discontinuous - "never timeout" is more similar in meaning to a very large timeout than a very small timeout. If your expectations are informed by the `setTimeout` function, then a timeout of "0" means "as soon as possible".

rstoenescu:
Went with a different approach, but the result should be the same - https://github.com/quasarframework/quasar/commit/a59eb69831fae1a727e852bda71e15ebdb5c17f7

A timeout of "0" has no meaning as essentially the notification should not appear on screen at all -- which is why we went with it like this -- and there are multiple examples of the same convention in the JS world.

Issue #16097, opened 2023-07-20
QSelect with behavior="dialog" and use-input


lmarne01:
### What happened?

I used QSelect with behavior="dialog" and the attribute use-input. The menu dialog appears on top of the screen. If I remove the use-input attribute, the dialog goes to the center of the screen.

### What did you expect to happen?

I expected both situations to act the same. Perhaps a new attribute position

### Reproduction URL

https://codepen.io/pen?&editors=101

### How to reproduce?

This is taken directly from documentation. Just click through the 2 selects and see the difference

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Components (quasar)

### Platforms/Browsers

Firefox, Chrome

### Quasar info output

```Shell
Operating System - Windows_NT(10.0.22621) - win32/x64
NodeJs - 20.3.0

Global packages
  NPM - 9.1.1
  yarn - 1.22.19
  @quasar/cli - 2.2.1
  @quasar/icongenie - Not installed
  cordova - Not installed

Important local packages
  quasar - 2.12.1 -- Build high-performance VueJS user interfaces (SPA, PWA, SSR, Mobile and Desktop) in record time
  @quasar/app-vite - 1.4.3 -- Quasar Framework App CLI with Vite
  @quasar/extras - 1.16.4 -- Quasar Framework fonts, icons and animations
  eslint-plugin-quasar - Not installed
  vue - 3.3.4 -- The progressive JavaScript framework for building modern web UI.
  vue-router - 4.2.2
  pinia - 2.1.4 -- Intuitive, type safe and flexible Store for Vue
  vuex - Not installed
  vite - 2.9.16 -- Native-ESM powered web dev build tool
  eslint - 8.43.0 -- An AST-based pattern checker for JavaScript.
  electron - Not installed
  electron-packager - Not installed
  electron-builder - Not installed
  register-service-worker - 1.7.2 -- Script for registering service worker, with hooks
  @capacitor/core - Not installed
  @capacitor/cli - Not installed
  @capacitor/android - Not installed
  @capacitor/ios - Not installed

Quasar App Extensions
  @parallaxworks/quasar-app-extension-quasar-charts - 0.2.5 -- Quasar Chart Components
  @parallaxworks/quasar-app-extension-quasar-file-upload - 0.1.3 -- A Quasar App Extension
  @parallaxworks/quasar-app-extension-quasar-navigation - 0.1.3 -- Quasar Navigation Components
[...]

PR #16711, opened 2023-12-17
fix(QSelect): dialog positioning in select behavior #16097

IceIsFrozen:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

Currently the dialog that appears from QSelect's behavior defaults to the middle of the screen (which is the dialog standard). However if use-input was set along with the behavior dialog setting it was hard-coded to the top of the screen. With this change just doing a general control of the dialog and defaulting to the standard (middle) position anyone who has a dialog with use-input will have an unexpected change. Now could code in support for both possibilities of dialog position determined default if use-input or not but seems unnecessary.

The fix though is simple in that adding dialog-position to top in the qselect to put it back.

**Other information:**

### DIFF
--- a/ui/src/components/select/QSelect.js
+++ b/ui/src/components/select/QSelect.js
@@ -116,6 +116,12 @@ export default createComponent({
       default: 'default'
     },
 
+    dialogPosition: {
+      type: String,
+      validator: v => [ 'standard', 'top', 'bottom', 'right', 'left' ].includes(v),
+      default: 'standard'
+    },
+
     virtualScrollItemSize: {
       type: [ Number, String ],
       default: void 0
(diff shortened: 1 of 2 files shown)


KaisNeffati:
Same here

rstoenescu:
Thank you for the effort here, and apologies for how long it sat. After review we won't be merging this one.

The core problem is the default: setting `dialog-position` to `standard` silently moves every existing `use-input` dialog from top to center. The top placement with `use-input` is deliberate mobile-keyboard UX: `behavior="dialog"` mostly kicks in on mobile, where the on-screen keyboard covers the lower half of the screen, so a centered dialog would end up hidden behind it. Two other pieces of QSelect assume that placement too (the iOS-specific `fade` transition for the dialog + `use-input` combo, and the scroll-to-top handler when the dialog's input gains focus), so making the position freely configurable needs those addressed as well, otherwise e.g. `bottom` with the keyboard open would be broken.

If we ever expose this, the prop would need to default to `undefined` (keep the current automatic behavior) with explicit values as opt-in, plus docs and tests. Given the branch is also conflicting with dev after two years, closing this rather than reworking it. See the note on #16097 for the rationale on the underlying report.

Issue #16265, opened 2023-08-28
feat: add typescript type definitions in the documentation


dongwa:
### Description

Thanks to @yusufkandemir, we now have TypeScript support for Quasar. However, we are unable to locate the type definitions in the documentation. Therefore, it's possible that we need to incorporate them into our documentation. For instance, we can observe that Quasar has exported a `QTableColumn` type for use with `QTable`, which assists in defining columns.

### Flavour

None

### Areas

TypeScript Support

PR #16779, opened 2024-01-17
feat(docs): add ts types  in the docs #16265

dongwa:

- [ ] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

image


dongwa:
I'm sorry for breaking the code format, but my eslint cannot be restored to the correct format. Can you help me correct it 
@rstoenescu  @yusufkandemir

rstoenescu:
Hi @dongwa, thank you for taking the time to work on this, and apologies that it sat for so long. The goal behind it (making exported TS types like `QTableColumn` discoverable in the docs, #16265) is absolutely valid and we want to solve it. After a close review though, we won't be merging this particular approach:

* **Page-to-types matching is fragile.** Matching a docs page to a `.d.ts` file by filename substring (`typeFiles.find(f => f.includes(fileName))`) takes the first hit and already needs hardcoded exceptions for `color-palette`, `touch-swipe` and `field`. Across hundreds of docs pages this would produce false positives and silent misses that nobody would notice until users report wrong types.
* **The regex cleanup corrupts the declarations.** Stripping `//` comments also eats URLs inside doc comments, and removing `import(...)` prefixes changes what the types actually say. The docs would present signatures that don't match the real ones.
* **Raw `.d.ts` dumps aren't the right shape for docs.** They duplicate the curated API cards and are large enough that the PR needs a collapse workaround. The supplementary types deserve curated placement on the relevant pages instead.
* **Build coupling.** Reading `node_modules/quasar/dist/types/api` at module scope makes the docs build crash whenever the ui dist types aren't built yet, which is a common state in the monorepo.
* The docs markdown pipeline has also been fully rewritten since January 2024, so the patch no longer applies to the current code.

We're keeping #16265 open and plan to address it deliberately, most likely by sourcing per-component "exposed types" sections from the same pipeline that feeds the API cards. Thanks again for pushing on this, the issue is better understood because of your work here.

PR #16783, opened 2024-01-18
feat(docs): Add convenient navigation like vscode

dongwa:

- [ ] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Add convenient navigation like vscode `command + p` hot key

https://github.com/quasarframework/quasar/assets/38457491/af4f2598-927d-49b6-87f8-af4d137d45ce

### DIFF
--- a/docs/src/layouts/doc-layout/DocHeader.vue
+++ b/docs/src/layouts/doc-layout/DocHeader.vue
@@ -27,6 +27,10 @@
         nav-class="text-uppercase text-size-16 letter-spacing-300"
       />
 
+      <div class="flex justify-center items-center">
+        <kbd class="quickly-navigation-kbd q-ma-sm text-grey-6" @click="docStore.openQuicklyNavigation"> {{ keysLabel }}</kbd>
+      </div>
+
       <doc-search />
 
       <div v-if="showThemeChanger" class="doc-header-icon-links q-ml-sm row no-wrap items-center">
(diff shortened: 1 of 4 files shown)


dongwa:
The final effect is as follows

https://github.com/quasarframework/quasar/assets/38457491/deccfd44-8f1c-470b-8b5a-4286ee199f07

rstoenescu:
Hi @dongwa, thank you for the PR and for the neat idea, and apologies that it sat for so long.

We took a fresh look at it and decided not to move forward. The reasoning:

The docs search (Ctrl/Cmd+K) already covers the "jump to a page by name" use-case. The search index emits a dedicated `page-link` entry for every page (carrying its menu path, description and extra keywords), so typing a page name already surfaces the page itself as a top hit. We verified this against the live search service with queries like "qbtn", "button", "table pagination" and "dark mode", and the target page ranks first or second in each case. Matching on page names also works better than matching on URL path fragments, since that is how readers think of pages.

The remaining benefits of a client-side route palette (instant results before the network round trip, working while the search service is down) are too marginal to justify a second, competing search UI in the header. There are also a couple of practical issues with the approach itself: Ctrl/Cmd+P shadows the browser's print shortcut, and raw route paths are not great option labels.

If anyone runs into a page that ranks poorly in search, the right fix is adding `keys:` to that page's frontmatter, which feeds the same index.

Thanks again for contributing, and we hope to see you around!

PR #16931, opened 2024-02-26
Add missing "persistent" prop for QTooltip

lyohaplotinka:

- [X] Bugfix
- [ ] Feature
- [X] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Description**
I discovered that the QTooltip component lacks the declaration of the "persistent" prop, while it is used within the component's code: 

```javascript
/* QTooltip.js */

// line 95:
const hideOnRouteChange = computed(() => props.persistent !== true)

// line 132:
const hasClickOutside = computed(() =>
  // it doesn't has external model
  // (null is the default value)
  props.modelValue === null
  // and it's not persistent
  && props.persistent !== true
  && showing.value === true
)
```

Because of this, there's no way to create a tooltip that wouldn't disappear when changing routes.

I encountered this need during my work. In this PR, I have added the missing prop and its description to the API.

### DIFF
--- a/ui/src/components/tooltip/QTooltip.js
+++ b/ui/src/components/tooltip/QTooltip.js
@@ -72,7 +72,9 @@ export default createComponent({
     hideDelay: {
       type: Number,
       default: 0
-    }
+    },
+
+    persistent: Boolean
   },
 
   emits: [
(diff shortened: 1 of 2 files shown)


rstoenescu:
Good catch! Thanks for contributing!

rstoenescu:
This will go into Quasar v2.14.6

PR #16932, opened 2024-02-26
feat(QImg): add `spinner-delay` prop

thexeos:

- [X] Feature

- [X] No

- [X] It's been tested on a Cordova (iOS, Android) app
- [X] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs) or explained in the PR's description.

- [X] A convincing reason for adding this feature

**Other information:**

We use QImg for displaying user avatars and country flags. We keep the spinner enabled, as it helps display the loading state to the user. After the first visit, the user will likely have some or all images cached, causing the spinner to quickly "flash" over the almost-instantly loaded image.

This pull request adds a `spinner-delay` prop which allows delaying the display of the spinner until after a short (user-defined) period of time has passed since the image begun loading.

The value is customizable, as it is context-dependent and only the developer knows how long it should be. For example, setting the value to 50 (milliseconds) is enough for loading isolated images within a single viewport, however loading dozens or hundreds of images in a tight grid will hog the browser's cache and network modules and would require a much larger delay (e.g., 1000 milliseconds).

The default value is 0 - consistent with current behavior.

We have been using this code for almost 2 years and it is stable.

chrome_QhRsBy6FCx

### DIFF
--- a/ui/src/components/img/QImg.js
+++ b/ui/src/components/img/QImg.js
@@ -60,7 +60,11 @@ export default createComponent({
     noTransition: Boolean,
 
     spinnerColor: String,
-    spinnerSize: String
+    spinnerSize: String,
+    spinnerDelay: {
+      type: Number,
+      default: 0
+    },
   },
 
   emits: [ 'load', 'error' ],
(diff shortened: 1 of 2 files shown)


rstoenescu:
Hi,

First of all, thank you for taking the time to contribute! Even though you have this working, you still wanted to improve Quasar with your feature, which is something that I can only applaud.

Went with a slightly different approach (delaying isLoading directly) and did some other refactoring.

Will be available in Quasar v2.14.6.

PR #16978, opened 2024-03-08
fix: remove regex double-escape in custom service worker denylist

thexeos:

- [X] Bugfix

Current regex has a backslash erroneously escaped: `/workbox-(.)*\\.js$/`, while it should actually be `/workbox-(.)*\.js$/`. The current version would match a pathname such as `/workbox-bar\.js` and even `/foo/workbox-bar/hello\ajs`, but not `/workbox-foo.js`. 

- [X] Yes

I've never personally seen any `workbox-*.js` files being emitted, but if they were then they are currently being served by Service Worker and with this change they will start to get loaded from the server instead.

- [X] It's submitted to the `dev` branch
- [ ] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs) or explained in the PR's description.

**Other information:**

It is also entirely possible that this regex is no longer relevant if Workbox is no longer emitting files with those names.

On the other hand, if we know these files are being emitted, then a more appropriate regex is `/workbox-[^/]+\.js$/`.

### DIFF
--- a/app-vite/templates/pwa/default/custom-service-worker.js
+++ b/app-vite/templates/pwa/default/custom-service-worker.js
@@ -24,7 +24,7 @@ if (process.env.MODE !== 'ssr' || process.env.PROD) {
   registerRoute(
     new NavigationRoute(
       createHandlerBoundToURL(process.env.PWA_FALLBACK_HTML),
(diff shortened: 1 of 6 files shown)


rstoenescu:
Thanks for catching this and contributing!

PR #17061, opened 2024-03-30
fix(ui): use correct list type for QSliderSlots (fix: #17060)

### LINKED ISSUE #17060: Incorrect TS type for `markerList` of QSliderSlots
The type definition for when defining custom QSlider marker labels, defined here:

https://github.com/quasarframework/quasar/blob/881803b8bc91552ce67a10124d411df6b432a176/ui/src/components/slider/use-slider.json#L315-L316

https://github.com/quasarframework/quasar/blob/881803b8bc91552ce67a10124d411df6b432a176/ui/src/components/slider/use-slider.json#L408-L409

Results in an incorrect generated types.

<details>

<summary>Show generated TS interface</summary>

```ts
export interface QSliderSlots {
  /**
   * What should the menu display after filtering options and none are left to be displayed; Suggestion: <div>
   * @param scope
   */
  "marker-label": (scope: {
    /**
     * Config for current marker label
     */
    marker: SliderMarkerLabelConfig;
    /**
     * Array of marker label configs
     */
    markerList: SliderMarkerLabelArrayConfig[];
    /**
     * Object with key-value where key is the model and the value is the marker label config
     */
    markerMap: SliderMarkerLabelObjectConfig;
    /**
     * Required CSS classes to be applied to the marker element
     */
    classes: string;
    /**
     * Get CSS style Object to apply to a marker element at respective model value; For perf reasons, use only if requested model value is not already part of markerMap
     * @param value The marker label equivalent model value
     * @returns CSS style Object to apply to a marker element at respective model value
     */
    getStyle: (value: number) => any;
  }) => VNode[];
  /**
   * What should the menu display after filtering options and none are left to be displayed; Suggestion: <div>
   * @param scope
   */
  "marker-label-group": (scope: {
    /**
     * Array of marker label configs
     */
    markerList: SliderMarkerLabelArrayConfig[];
    /**
     * Object with key-value where key is the model and the value is the marker label config
     */
    markerMap: SliderMarkerLabelObjectConfig;
    /**
     * Required CSS classes to be applied to the marker element
     */
    classes: string;
    /**
     * Get CSS style Object to apply to a marker element at respective model value; For perf reasons, use only if requested model value is not already part of markerMap
     * @param value The marker label equivalent model value
     * @returns CSS style Object to apply to a marker element at respective model value
     */
    getStyle: (value: number) => any;
  }) => VNode[];
}
```

</details>

Note how `markerList` in generated types for both slots is `SliderMarkerLabelArrayConfig[]`. However, `SliderMarkerLabelArrayConfig` is already an array of `SliderMarkerLabelConfig`:

https://github.com/quasarframework/quasar/blob/881803b8bc91552ce67a10124d411df6b432a176/ui/types/api/slider.d.ts#L43

So we end up with `typeof markerList === SliderMarkerLabelConfig[][]`.

thexeos:

- [X] Bugfix

- [X] No

This fixes #17060

### DIFF
--- a/ui/src/components/slider/use-slider.json
+++ b/ui/src/components/slider/use-slider.json
@@ -317,7 +317,7 @@
 
         "markerList": {
           "type": "Array",
-          "tsType": "SliderMarkerLabelArrayConfig",
+          "tsType": "SliderMarkerLabelConfig",
           "desc": "Array of marker label configs",
           "definition": {
             "index": {
@@ -410,7 +410,7 @@
       "scope": {
         "markerList": {
           "type": "Array",
-          "tsType": "SliderMarkerLabelArrayConfig",
+          "tsType": "SliderMarkerLabelConfig",
           "desc": "Array of marker label configs",
           "definition": {
             "index": {
--- a/ui/types/api/slider.d.ts
+++ b/ui/types/api/slider.d.ts
@@ -40,8 +40,6 @@ export type SliderMarkerLabelConfig = {
   style: VueStyleObjectProp;
 };
 
-export type SliderMarkerLabelArrayConfig = SliderMarkerLabelConfig[];
-
 export interface SliderMarkerLabelObjectConfig {
   [value: number]: SliderMarkerLabelConfig;
 }


yusufkandemir (approved):
Thanks for contributing! I also removed the type itself as it's not needed anymore.

PR #17130, opened 2024-04-22
fix: ios always undefined

bartduisters:

- [x] Bugfix

This reverts an `else` that was added two months ago, perhaps by mistake? With the else, it is impossible to detect 'iOS' when using either chrome, edge or firefox on an iOS device.

- [ ] Yes
- [x] No

**Other information:**
This came from feedback in Discord.

Screenshot 2024-04-22 at 21 39 29


rstoenescu:
Thanks for contributing! This will go into Quasar v2.15.4

PR #17168, opened 2024-05-06
feat(create-quasar): add TS AE template

yusufkandemir:

- Feature

- No

**Other information:**
Contains fixes for app-vite&app-webpack types of AE API files, e.g. IndexAPI

Adds a TS AE template, highlights:
- TypeScript setup, of course
- ESLint v9 using flat config format, with a good configuration (_ESLint v9 will be backported to all other relevant templates later_)
- Vue SFC (.vue files) support (_will be backported to the JS template later_) (thanks @danielroe for tips)
- Monorepo structure with app-vite and app-webpack playground apps

There is a limitation with the SFC support. When using `<script setup lang="ts">`, it won't work in JS app-webpack apps. This is because mkdist skips processing it and there is no TypeScript support in JS app-webpack, so it fails. If https://github.com/unjs/mkdist/pull/210 gets released, then this limitation will become "when using `<script setup lang="ts">` AND type-only macros, e.g. `defineProps<QBtnProps>()`.
Alternatively, to completely eliminate the limitation, we could add TS processing support to app-webpack when using JS. We may do this in the core, or via a snippet in AE index API. app-vite JS works fine due to how Vite is configured to handle TS automatically.
For now, I just left some comments explaining the limitations.

**What to do next**:

I haven't added the UI kit due to complexity, limitations, time constraints, etc. It can be plugged into the template in the form of another workspace entry in the monorepo later. As the AE can now also ship SFC files and act as a library that can expose methods and such, the importance of the UI kit is reduced. The only purpose of the UI kit is to support other flavors, e.g. Vite plugin, Vue CLI plugin, and UMD. UMD will be problematic regarding SFC support as AEs will ship transpiled .vue files and there is no build system to integrate those. We may need to build SFCs into .js and ship it that way for UMD, but we would need to figure out the limitations to do that.

We may also want to offer the non-monorepo version of the TS AE kit as people might want to create AEs into their existing monorepos.


create-quasar/templates/app-extension/ae-ts/BASE/_eslint.config.js
```diff
@@ -0,0 +1,106 @@
+import eslint from '@eslint/js';
+import tseslint from 'typescript-eslint';
+import pluginVue from 'eslint-plugin-vue';
+import configPrettier from 'eslint-config-prettier';
+import globals from 'globals';
```
IlCallo, line 5: `globals` package hasn't been installed, causing a TS error at this import
Tested using `pnpm`
Adding it as a dev dependecy solves the problem

Apparently it has been removed from ESLint since v9 and is meant to be installed manually when needed: [eslint.org/docs/latest/use/configure/language-options#predefined-global-variables](https://eslint.org/docs/latest/use/configure/language-options#predefined-global-variables)

create-quasar/templates/app-extension/ae-ts/BASE/_eslint.config.js
```diff
@@ -0,0 +1,106 @@
...
+import { version as vueVersion } from 'vue';
```
IlCallo, line 6: `vue` package hasn't been installed at root level, causing a TS error at this import

This can be solved by
- using `shamefully-hoist` option into a `.npmrc` file at root level
- removing vue dependency into the 3 packages of the monorepo and add it at root level

I tested both solutions with `pnpm` and `npm`
Unluckily the second one doesn't work for `npm` (of course), so I'd go with the first one
I haven't tested it with `yarn` and `bun`

create-quasar/templates/app-extension/ae-ts/BASE/_package.json
```diff
@@ -0,0 +1,31 @@
...
+    "@eslint/js": "^9.1.1",
+    "@types/eslint": "^8.56.10",
+    "@types/eslint__js": "^8.42.3",
+    "@types/node": "^20.12.8",
+    "eslint": "^9.1.1",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-plugin-vue": "^9.25.0",
+    "prettier": "^3.2.5",
+    "typescript": "^5.4.5",
+    "typescript-eslint": "^7.8.0"
```
IlCallo: A new release of `typescript-eslint` is out which supports ESLint 9
Using v7 triggers a peerDeps warning when installing

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-vite/README.md
```diff
@@ -0,0 +1,37 @@
+# Quasar AE <%= name %> Playground (quasar-cli-vite)
+
+Playground for Quasar AE <%= name %>
+
+## Install the dependencies
+
+```bash
+pnpm install
```
IlCallo, line 13: Shouldn't the package manager here be dynamically filled in instead of hardcoded to `pnpm`?

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-vite/_package.json
```diff
@@ -0,0 +1,29 @@
...
+  "scripts": {
+    "format": "prettier --write \"**/*.{js,ts,vue,scss,html,md,json}\" --ignore-path .gitignore",
+    "test": "echo \"No test specified\" && exit 0",
+    "dev": "quasar dev",
+    "build": "quasar build"
+  },
+  "dependencies": {
+    "@quasar/extras": "^1.16.11",
+    "quasar": "^2.15.2",
+    "<%= pkgName %>": "workspace:*",
```
IlCallo: Does this work with `npm` and `yarn` too?
Isn't `workspace:*` a `pnpm` thing?

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-vite/_tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
+  "extends": "../../tsconfig.json",
```
IlCallo: Any reason for not using `@quasar/app-vite/tsconfig-preset`?
I tried it out and it worked like a charm, slimming the file down considerably

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-vite/src/quasar.d.ts
```diff
@@ -0,0 +1,9 @@
+// Forces TS to apply `@quasar/app-vite` augmentations of `quasar` package
+// Removing this would break `quasar/wrappers` imports as those typings are declared
+//  into `@quasar/app-vite`
+// As a side effect, since `@quasar/app-vite` reference `quasar` to augment it,
+//  this declaration also apply `quasar` own
+//  augmentations (eg. adds `$q` into Vue component context)
+/// <reference types="@quasar/app-vite" />
+
+/// <reference types="<%= pkgName %>" />
```
IlCallo, line 10: I'd add a comment about this line too, it may not be that clear to newcomers

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-webpack/README.md
```diff
@@ -0,0 +1,37 @@
+# Quasar AE <%= name %> Playground (quasar-cli-webpack)
+
+Playground for Quasar AE <%= name %>
+
+## Install the dependencies
+
+```bash
+pnpm install
```
IlCallo, line 13: Shouldn't this use the package manager name chosen by the user?

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-webpack/_package.json
```diff
@@ -0,0 +1,40 @@
...
+    "format": "prettier --write \"**/*.{js,ts,vue,scss,html,md,json}\" --ignore-path .gitignore",
+    "test": "echo \"No test specified\" && exit 0",
+    "dev": "quasar dev",
+    "build": "quasar build"
+  },
+  "dependencies": {
+    "@quasar/extras": "^1.16.11",
+    "core-js": "^3.6.5",
+    "quasar": "^2.15.2",
+    "<%= pkgName %>": "workspace:*",
```
IlCallo: Does this syntax work with `npm` and `yarn` too?

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-webpack/_tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
+  "extends": "../../tsconfig.json",
```
IlCallo: Any reason for not using `@quasar/app-webpack/tsconfig-preset`?
I tried it out and it worked like a charm, slimming the file down considerably

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-webpack/postcss.config.cjs
```diff
@@ -0,0 +1,8 @@
+// https://github.com/michael-ciniawsky/postcss-load-config
```
IlCallo, line 1: This requires a `/* eslint-disable */` as other config files, otherwise linting will fail due to `require` usage

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-webpack/src/quasar.d.ts
```diff
@@ -0,0 +1,9 @@
+// Forces TS to apply `@quasar/app-webpack` augmentations of `quasar` package
+// Removing this would break `quasar/wrappers` imports as those typings are declared
+//  into `@quasar/app-webpack`
+// As a side effect, since `@quasar/app-webpack` reference `quasar` to augment it,
+//  this declaration also apply `quasar` own
+//  augmentations (eg. adds `$q` into Vue component context)
+/// <reference types="@quasar/app-webpack" />
+
+/// <reference types="<%= pkgName %>" />
```
IlCallo, line 10: I'd add a comment on this one too, to explain it to TS newbies

IlCallo (changes requested):
`.gitignore` file is missing into `templates/app-extension/ae-ts/BASE`

Example content:
```ini
.DS_Store
.thumbs.db
node_modules
dist

# Log files
npm-debug.log*
yarn-debug.log*
yarn-error.log*

# Editor directories and files
.idea
*.suo
*.ntvs*
*.njsproj
*.sln
```
This is important since we now auto-initialize a Git repo when the scaffolding completes

---

Latest commits and in general some `package.json` lines assume the dev will be using `pnpm`

E.g. `workspace:*` would break installation with `npm` and `yarn`
For `npm` it seems that using `^0.0.1` would be fine to get past deps installation, I couldn't find a way to say "just take the latest available in the repository"

I haven't tested this with Bun

**This needs to be fixed before merging**

Btw, I simulated the usage of different package managers with this command, to get `runningPackageManager` filled in with some actual values

`npm_config_user_agent=npm/10.5.0 node <path-to-quasar-folder>/quasar/create-quasar/index.js`

---

I haven't explored how hard it would be to implement this, but we should explore at some point running the root level `format` command after deps installation but before linting
This should probably be done automatically and for all `create-coralloy` templates, if a `format` script actually exists, without an option to let the user choose

---

I went ahead, rebased the PR and fixed a couple of my own comments in my local copy
If you're ok with it, I can force-push the rebase and my commits for you

---

If [this](https://github.com/unjs/mkdist/pull/210) is fixed, I can ping Daniel about it and see if we can get it in

---

When running the AE `build` command, I got a warning from `mkdist` about installing `vue-tsc` to get proper typings for components, or something similar
Are you aware of it?

IlCallo:
Check out latest commits in [this branch](https://github.com/IlCallo/quasar/tree/ae-ts-fixes), where I pushed some fixes for the problems I highlighted

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-webpack/postcss.config.cjs
yusufkandemir: The ignore patterns were incorrect. I've adjusted it to address this.

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-vite/_package.json
yusufkandemir: I don't think so. The template was pnpm only at first. It seems I've overlooked this part. I will think about it and do some research. yarn v1 should auto-link them. Not sure how npm does it. Should be solvable with some if-else rendering.

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-vite/README.md
yusufkandemir: Yes, will do

create-quasar/templates/app-extension/ae-ts/BASE/_package.json
yusufkandemir: It was updated, but not have been pushed yet at the time of your comment. All the packages have been updated.

create-quasar/templates/app-extension/ae-ts/BASE/_eslint.config.js
yusufkandemir: I've gone with the first way. But it wasn't pushed at the time of your comment.

create-quasar/templates/app-extension/ae-ts/BASE/_eslint.config.js
yusufkandemir: Already installed, but wasn't pushed at the time of your comment.

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-vite/_tsconfig.json
yusufkandemir: It's for re-using the root config. There is no way to extend both. The root config shouldn't be special now, so we could use the app-vite preset. If the users greatly customize their tsconfigs, they could update it back to point to the root config, if needed. I'll see what the differences would be and decide.

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-vite/src/quasar.d.ts
yusufkandemir: That's a good idea. I should also add it to the AE installation instructions. That way, the users who will install the AE can also benefit from the global component types.

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-webpack/README.md
yusufkandemir: Yes, will do

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-webpack/_package.json
yusufkandemir: Resolving in favor of the app-vite one

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-webpack/_tsconfig.json
yusufkandemir: Resolving in favor of the app-vite one

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-webpack/src/quasar.d.ts
yusufkandemir: Resolving in favor of the app-vite one

yusufkandemir:
> I went ahead, rebased the PR and fixed a couple of my own comments in my local copy
If you're ok with it, I can force-push the rebase and my commits for you

I already had rebased it and made some changes. They were waiting locally. So, I force-pushed them instead. Thanks. I will check the changes for other bits that I haven't addressed.

> If https://github.com/unjs/mkdist/pull/210 is fixed, I can ping Daniel about it and see if we can get it in

I've done it myself, thanks.

> Btw, I simulated the usage of different package managers with this command, to get runningPackageManager filled in with some actual values
> npm_config_user_agent=npm/10.5.0 node <path-to-quasar-folder>/quasar/create-quasar/index.js

Just FYI, `pnpm node .../index.js`, `yarn node .../index.js`, etc. also work, while inside a directory with package.json.

> When running the AE build command, I got a warning from mkdist about installing vue-tsc to get proper typings for components, or something similar
Are you aware of it?

It's a recent change. I've noticed it recently, after returning for the final touches. Installing it didn't make any difference for the test component. But, I'll add vue-tsc to the template, to avoid the warning and just in case.

> I haven't explored how hard it would be to implement this, but we should explore at some point running the root level format command after deps installation but before linting

We used to run the format script, but probably after listing. It was probably a very long time ago. Maybe it was lost while migrating from starter-kit to create-quasar. It wasn't really hard to do so. Running it after linting is better, when ESLint auto-fixes some stuff. For example, if there is an unused eslint-disable comment, ESLint removes the comment, leaving the line empty. Running the format script afterward, clean up those empty lines, and adjust other styles if needed. Do you have any reason why do you want to run linting before formatting?

IlCallo:
> Do you have any reason why do you want to run linting before formatting?

Honestly, your reasoning makes much more sense than my own thoughts
I probably remember wrong and linting > formatting is the correct order

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-webpack/README.md
yusufkandemir: done

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-vite/_tsconfig.json
yusufkandemir: done

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-vite/_package.json
yusufkandemir: done

create-quasar/templates/app-extension/ae-ts/BASE/playground/quasar-cli-vite/README.md
yusufkandemir: done

IlCallo (approved):

rstoenescu:
AE, TS, PNPM:
image

rstoenescu:
AE, TS, Yarn:

1. "To get started" section is wrong. We'll need some custom code.
2. Picked yarn and still having pnpm stuff (pnpm-workspace, package.json > scripts > pnpm commands)

image

rstoenescu (changes requested):
Posted two comments earlier

yusufkandemir:
@rstoenescu

Regarding the instructions for getting started in the CLI message, I'll check it out.

The package manager stuff is like that because it uses the detected package manager to scaffold stuff accordingly. The selection only happens after the scaffolding is complete, so we can't use it with how the code is structured now. `pnpm create`, `npm init`, etc. should all detect the package manager fine. The only case I can find is running the script directly via Node. So, in practice, I think it's fine.

So, do you think we should leave it as is? Should I rework how package manager selection happens to scaffold stuff accordingly? Should I do something else?

IlCallo:
Quoting from my first comment

> Btw, I simulated the usage of different package managers with this command, to get runningPackageManager filled in with some actual values
> `npm_config_user_agent=npm/10.5.0 node <path-to-quasar-folder>/quasar/create-quasar/index.js`

Use this to test out how users will actually experience this

Maybe we can add a couple scripts on `create-quasar` to allow for an easier testing from contributors and ourselves?
With a comment linking this PR about why it's needed

PR #17193, opened 2024-05-14
feat[QSlideItem] new composable incapsulating swipe reset

RussianYankee:

- [ ] Bugfix
- [x] Feature
- [x] Documentation
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/docs/src/assets/menu.js
+++ b/docs/src/assets/menu.js
@@ -696,6 +696,11 @@ export default [
         name: 'useSplitAttrs',
         badge: 'new',
         path: 'use-split-attrs'
+      },
+      {
+        name: 'useSlideItemSwipe',
+        badge: 'new',
+        path: 'use-slide-item-swipe'
       }
     ]
   },
--- a/docs/src/pages/vue-composables/use-slide-item-swipe.md
+++ b/docs/src/pages/vue-composables/use-slide-item-swipe.md
@@ -1,0 +1,47 @@
+---
+title: useSlideItemSwipe composable
+desc: What is useSlideItemSwipe() composable and how you can use it
+keys: useSlideItemSwipe
+badge: Quasar v2.15+
+---
+
+The `useSlideItemSwipe()` is designed to simplify usage of the [Slide Item](https://quasar.dev/vue-components/slide-item) component. This composable encapsulates invoking user defined method on swipe,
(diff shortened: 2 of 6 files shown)


rstoenescu:
Thank you for taking the time to build and document this, and apologies that it sat unreviewed for so long. After going through it carefully, we've decided not to move this forward. I want to lay out the actual reasoning rather than just close it:

1. **It overlaps with the existing `useTimeout()` composable.** The core value here (a timeout that auto-clears on unmount) is already covered by `useTimeout()`, so the same result is a few lines of app code while keeping QSlideItem's natural event signature:

```js
import { useTimeout } from 'quasar'
const { registerTimeout } = useTimeout()

function onLeft ({ reset }) {
  doSomething()
  registerTimeout(reset, 300)
}
```

```html
<q-slide-item @left="onLeft">...</q-slide-item>
```

Compare that with the proposed usage, which requires wiring the payload manually in the template (`@left="onLeft({ resetFn: $event.reset, param: 1234 })"`) and restricts you to a single extra parameter. The wrapper ends up with more API surface than the pattern it replaces.

2. **Auto-reset after a fixed delay is an app-level UX decision, not a framework primitive.** In many real usages the right moment to call `reset()` is when an async action settles (e.g. a delete request succeeds or fails), not after a fixed number of milliseconds. Quasar's composables (`useTimeout`, `useInterval`, `useTick`, ...) are deliberately generic building blocks; blessing one component-specific usage pattern as an official composable would steer users toward the narrower pattern and commit us to documenting and maintaining it indefinitely.

3. There's also a subtle correctness issue that illustrates why this state wants to live per-item in app code: the returned handler is naturally shared across all items of a list, but the composable holds a single `timer`, and `finalize()` overwrites it without clearing the previous one. Swipe two items in quick succession and the first pending reset's timer id is lost, so it can no longer be cancelled on unmount. Fixable, of course, but it shows the abstraction boundary is off.

So this fits better as an app-land recipe built on `useTimeout()` than as public API. As a side note for future proposals: the feature-discussion step in the PR template exists precisely so that scope fit gets settled before you invest this kind of effort.

Thanks again for the contribution, and we'd be glad to see more from you.

PR #17207, opened 2024-05-18
fix(QScrollArea): correctly compute snap position for thumb #17206

### LINKED ISSUE #17206: QScrollArea: thumb doesn't center on mouse point for small thumbs (long content)
### What happened?

When pressing on the bar (not the thumb) of QScrollArea, the content scrolls such that middle of the thumb is positioned underneath the mouse cursor. This behavior is consistent with native scrolling (although inline [non-<body>] scroll can behave differently).

It works well when there is little content inside scroll area and the thumb size is directly proportional to the size of the content. But when there is a lot of content the snapping breaks.

### Reproduction URL

https://codepen.io/thexeos/pen/abrdvVL

### How to reproduce?

1. Go to reproduction link
2. Press on scroll bar area underneath the thumb of the yellow container
3. See how the thumb snaps to be centered on mouse pointer
4. Now do the same with amber container
5. Notice how the further down you click the more off-center the snapping becomes

I'll be submitting a PR shortly.

thexeos:
Resolves #17206

- [X] Bugfix

- [X] No

### DIFF
--- a/ui/src/components/scroll-area/QScrollArea.js
+++ b/ui/src/components/scroll-area/QScrollArea.js
@@ -317,8 +317,9 @@ export default createComponent({
       if (data.thumbHidden.value !== true) {
         const offset = evt[ dirProps[ axis ].offset ]
         if (offset < data.thumbStart.value || offset > data.thumbStart.value + data.thumbSize.value) {
-          const pos = offset - data.thumbSize.value / 2
-          setScroll(pos / container[ axis ].value * data.size.value, axis)
+          const targetThumbStart = offset - data.thumbSize.value / 2
+          const percentage = between(targetThumbStart / (props.container[ axis ].value - data.thumbSize.value), 0, 1)
+          setScroll(percentage * Math.max(0, data.size.value - props.container[ axis ].value), axis)
         }
 
         // activate thumb pan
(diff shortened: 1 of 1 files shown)


ui/src/components/scroll-area/QScrollArea.js
```diff
@@ -317,8 +317,9 @@ export default createComponent({
       if (data.thumbHidden.value !== true) {
         const offset = evt[ dirProps[ axis ].offset ]
         if (offset < data.thumbStart.value || offset > data.thumbStart.value + data.thumbSize.value) {
-          const pos = offset - data.thumbSize.value / 2
-          setScroll(pos / container[ axis ].value * data.size.value, axis)
+          const targetThumbStart = offset - data.thumbSize.value / 2
+          const percentage = between(targetThumbStart / (props.container[ axis ].value - data.thumbSize.value), 0, 1)
```
rstoenescu: Hi,

There is no `container` prop. Did you mean `container` instead of `props.container`?

rstoenescu (changes requested):

ui/src/components/scroll-area/QScrollArea.js
thexeos: Hi. You are correct! I have taken this from another branch that has `props.container`. New commit pushed.

rstoenescu:
Thanks for contributing!
This will go into Quasar v2.16.12

PR #17301, opened 2024-06-20
feat(app-vite&app-webpack): upgrade tsconfig preset

IlCallo:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

Minumum accepted TS version would be 5, since we'll use `bundler` option for `moduleResolution`

**Other information:**

I tested Vite new preset on a couple of projects of ours and aside the problems I reported down there, which should be used as base for a migration guide, I didn't have other problems

---

TS reference suggests to avoid using `baseUrl` and rely on relative values for `path` option
This isn't possible in our case, since we ship default "paths" in the presets instead of the devland tsconfig
Using relative paths in the preset would cause the mapping to search types relatively to the preset tsconfig location and not the devland project folder
There doesn't seem to be another way to achieve this feature, and putting the paths back into the devland tsconfig would remove a good chunk of our flexibility to update those paths on the user behalf
Instead, I added comments to explain this where needed
We could remove "baseUrl" from jsconfig or older tsconfig from Qv1 which weren't using a tsconfig preset, but it doesn't really make sense changing something that's not broken

---

Switching to `bundler` option for `moduleResolution` breaks any kind of deep import, since it nows honors `exports` field in `package.json` when defined
Deep imports are still possible, but the dev needs to define an appropriate entry in `paths` option mapping to the deep import location into `node_modules` folder

E.g.

```json
{
  "compilerOptions": {
    "paths": {
      // ... other Quasar paths definition
      "cypress/types/net-stubbing": ["node_modules/cypress/types/net-stubbing"],
      "@vue/apollo-composable/dist/*": ["node_modules/@vue/apollo-composable/dist/*"]
    }
  }
}
```  

Additionally, `bundler` mode won't recognize CJS imports, thus any kind of CJS file, especially `quasar.conf.js` or `quasar.conf.cjs` in our case, isn't correctly interpreted by TS
I haven't tried, but I expect `quasar.conf.ts` or `quasar.conf.mjs` to work with this setup

---

I added a new version of the presets which is even stricter, and thus could not fit everyone

`useDefineForClassFields` in Vite tsconfig preset isn't needed, since it's true by default when `target` is `esnext` as it is.
This is true since [Vite 2.5 apparently](https://vitejs.dev/guide/features#usedefineforclassfields)

Regarding `target` option, it's actually ignored by Vite, so setting it to `esnext` is a good way to force `useDefineForClassFields` to true and accept any kind of new syntax and feature

For webpack, `target` is apparently used to [determine what to transpile](https://webpack.js.org/guides/typescript/#basic-setup), so I guess we should still keep it set to `es6`

From TS documentation:
> Setting this value to the lowest ECMAScript version that you intend to support ensures the emitted code will not use language features introduced in a later version.

### DIFF
--- a/app-vite/stricter-tsconfig-preset.json
+++ b/app-vite/stricter-tsconfig-preset.json
@@ -2,6 +2,7 @@
   "extends": "./tsconfig-preset.json",
   "compilerOptions": {
     "exactOptionalPropertyTypes": true,
+    "noImplicitReturns": true,
     "noPropertyAccessFromIndexSignature": true,
     "noUncheckedIndexedAccess": true
   }
--- a/app-vite/stricter-tsconfig-preset.json
+++ b/app-vite/stricter-tsconfig-preset.json
@@ -1,0 +1,8 @@
+{
+  "extends": "./tsconfig-preset.json",
+  "compilerOptions": {
+    "exactOptionalPropertyTypes": true,
+    "noPropertyAccessFromIndexSignature": true,
+    "noUncheckedIndexedAccess": true
+  }
+}
--- a/app-vite/tsconfig-preset.json
+++ b/app-vite/tsconfig-preset.json
@@ -1,23 +1,20 @@
 {
   "compilerOptions": {
+    "allowImportingTsExtensions": true,
     "allowJs": true,
     // `baseUrl` must be placed on the extending configuration in devland, or paths won't be recognized
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
     // Needed to address https://github.com/quasarframework/app-extension-typescript/issues/36
     "noEmit": true,
     "resolveJsonModule": true,
-    // Avoid cross-os errors due to inconsistent file casing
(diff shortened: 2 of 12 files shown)


MartinX3:
@rstoenescu I don't have the `stricter-tsconfig-preset.json` in Beta 16.
I also don't see it here
https://www.npmjs.com/package/@quasar/app-vite/v/2.0.0-beta.16?activeTab=code

Maybe it didn't get released?

yusufkandemir:
@MartinX3 thanks for noticing it. It's due to not adding the file to `package.json > files`.
Fixed in https://github.com/quasarframework/quasar/commit/151d0444e9b7ece67f88be80c84f7839a33f3daa and https://github.com/quasarframework/quasar/commit/6b4e280806e3a82a09645334a2f26706569e8035.

MartinX3:
@yusufkandemir thank you!
Now we just need another release :D

rstoenescu:
Released both q/app-vite betas

PR #17429, opened 2024-08-09
fix(app-vite&app-webpack): fix types feature flags

yusufkandemir:

- Bugfix
- Feature

- No

**Other information:**
#17301 changes `moduleResolution` from `node` to `bundler`. That broke the feature flags system. Here is an example common error:
> Property `store` does not exist on type `BootFileParams<any>`

It's because even though TS normally doesn't enforce the file extension inside the app code and imports from node_modules when using `bundler`, it does enforce it for wildcard `package.json > exports` entries. So, `quasar/dist/types/feature-flag` no longer works and must be changed to `quasar/dist/types/feature-flag.d.ts` instead.

Does not have to be backported to current stable releases of app-vite and app-webpack, as long as #17301 is not backported either.

This change will only apply to feature flags generated after the release of this. So, existing projects will only benefit from it when they add Pinia or build modes, OR delete the feature flag files and run `quasar dev`/`quasar build` to re-generate them. New projects will work fine. So, we should add a notice to [`@quasar/app-vite-v2.0.0-beta.16`](https://github.com/quasarframework/quasar/releases/tag/%40quasar%2Fapp-vite-v2.0.0-beta.16) and [`@quasar/app-webpack-v4.0.0-beta.17`](https://github.com/quasarframework/quasar/releases/tag/%40quasar%2Fapp-webpack-v4.0.0-beta.17) to notify them. We better also post an announcement about it. I recommend informing the users to delete the files, and then regenerate them by running `quasar dev`/`quasar build`, instead of letting them update them individually. This will be easier to tell and less error-prone.

Using `rimraf` through `npx`(or `pnpx`, `yarn dlx`, `bunx`)
```sh
$ npx rimraf -g ./**/*-flag.d.ts
$ quasar build # or dev
```

Unix-like (Linux, macOS):
```sh
$ rm ./**/*-flag.d.ts
$ quasar build # or dev
```

Windows (CMD):
```cmd
> del /s *-flag.d.ts
> quasar build &:: or dev
```

Windows (PowerShell):
```ps
> Remove-Item -Recurse -Filter *-flag.d.ts
> quasar build # or dev
```


rstoenescu:
Thanks for the detailed explanation!

PR #17476, opened 2024-08-31
fix(ui): IME composition detection not working

NitroRCr:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Fix the issue where QInput & QSelect updates model-value during the IME composition process.

Previously, the code detected `e.data` and set qComposing to true only when a CJK character was detected. This might work in some IMEs, but it at least doesn't work in my environment (Windows' default Microsoft Pinyin IME), as during composition, `e.data` contains the original key characters instead of CJK characters:
image

To fix this, simply use `compositionstart` and `compositionend`, just like the [official Vue v-model implementation](https://github.com/vuejs/core/blob/d298c431cc422b53cf4e9c69bf1daf926c33b6e0/packages/runtime-dom/src/directives/vModel.ts#L28).

After making the modification, I tested it, and it works well.

### DIFF
--- a/ui/src/composables/private.use-key-composition/use-key-composition.js
+++ b/ui/src/composables/private.use-key-composition/use-key-composition.js
@@ -1,10 +1,3 @@
-import { client } from '../../plugins/platform/Platform.js'
-
-const isJapanese = /[\u3000-\u303f\u3040-\u309f\u30a0-\u30ff\uff00-\uff9f\u4e00-\u9faf\u3400-\u4dbf]/
-const isChinese = /[\u4e00-\u9fff\u3400-\u4dbf\u{20000}-\u{2a6df}\u{2a700}-\u{2b73f}\u{2b740}-\u{2b81f}\u{2b820}-\u{2ceaf}\uf900-\ufaff\u3300-\u33ff\ufe30-\ufe4f\uf900-\ufaff\u{2f800}-\u{2fa1f}]/u
-const isKorean = /[\u3131-\u314e\u314f-\u3163\uac00-\ud7a3]/
-const isPlainText = /[a-z0-9_ -]$/i
-
 export default function (onInput) {
   return function onComposition (e) {
     if (e.type === 'compositionend' || e.type === 'change') {
@@ -12,18 +5,8 @@ export default function (onInput) {
       e.target.qComposing = false
       onInput(e)
     }
-    else if (
-      e.type === 'compositionupdate'
-      && e.target.qComposing !== true
-      && typeof e.data === 'string'
-    ) {
-      const isComposing = client.is.firefox === true
-        ? isPlainText.test(e.data) === false
-        : isJapanese.test(e.data) === true || isChinese.test(e.data) === true || isKorean.test(e.data) === true
-
-      if (isComposing === true) {
-        e.target.qComposing = true
-      }
+    else if (e.type === 'compositionstart') {
+      e.target.qComposing = true
     }
   }
 }


rstoenescu:
Thank you for your awesome contribution!
This will go into Quasar v2.16.10

jtilford:
I'm having a problem with v2.16.10 that I'm not having with v2.16.9, related to a QInput field.

On Android (8.1.0 tested) with the software on-screen keyboard, the v-model for a QInput doesn't update when characters (English) are entered, but DOES seem to update on enter, space, or backspace. After the first update, "enter" seems to not update the model anymore.

The characters all show in the input. It's the code watching the v-model that doesn't get the new characters until those noted keys (at least) are pressed.

Use-case: We have search on our app where users type the characters of a product they want to see, and until v2.16.10, every character input would update the search.

On web builds, this problem is not present -- we can search on web and every character input to the QInput updates the v-model value live.

NitroRCr:
It seems that this issue only occurs on a few on-screen keyboards. I was able to reproduce the problem on an AOSP on-screen keyboard on Android 7, but the same situation also appears in the official Vue v-model implementation ([tested here](https://vuejs.org/guide/essentials/forms)). 

I observed that although it's English characters that are being typed, the characters in the input field have an underline, indicating that it is in IME composition mode. At this point, the `model-value` does not update, and I don't think this is a problem. 

Perhaps it should be considered to add `input` event to QInput in order to handle input events during the composition process. As long as there is composition detection, `input` and `update:model-value` are different events, and I believe both are necessary.

jtilford:
I'm a little unclear on your suggestion's impact. Would I need to change my inputs across my app, after Quasar 2.16.9, to use something else instead of v-model? Or can we restore the previous functionality of the inputs, and make IME composition detection some kind of boolean parameter or a separate input type?

NitroRCr:
> I'm a little unclear on your suggestion's impact. Would I need to change my inputs across my app, after Quasar 2.16.9, to use something else instead of v-model? Or can we restore the previous functionality of the inputs, and make IME composition detection some kind of boolean parameter or a separate input type?

I was suggesting the developers of Quasar to add `input` event to QInput. This way, it would be possible to handle input during the IME composition process using `:model-value="model"` and `@input="model = $event"`. Later, I made these modifications myself and created PR #17543. Unfortunately, that PR has not received a response.

jtilford:
> I was suggesting the developers of Quasar to add `input` event to QInput. This way, it would be possible to handle input during the IME composition process using `:model-value="model"` and `@input="model = $event"`. Later, I made these modifications myself and created PR #17543. Unfortunately, that PR has not received a response.

I appreciate your quick reply. Looking at your PR, I think I can answer my own question.

The PR would restore the prior (<=2.16.9) functionality to `@update:model-value`, while adding `@input` to allow IME composition detection (like `@update:model-value` has been working in 2.16.10 and up).

Is that correct?

NitroRCr:
> I appreciate your quick reply. Looking at your PR, I think I can answer my own question.
> 
> The PR would restore the prior (<=2.16.9) functionality to `@update:model-value`, while adding `@input` to allow IME composition detection (like `@update:model-value` has been working in 2.16.10 and up).
> 
> Is that correct?

I'm afraid not. IME composition detection means `model-value` won't update during IME composition process until composition is complete. The PR wouldn't restore the functionality to `@update:model-value`, but would add `@input`, which is also emitted during the IME composition process. This behavior is consistent with using `@update:model-value` or `@input` dirctly on native input element. In your use case you need to use `@input` instead of `v-model` if that PR is merged.

rstoenescu:
Hi All,

I am temporarily reverting this PR in v2.17.1 because it causes even more trouble (see https://github.com/quasarframework/quasar/issues/17536). I am open to suggestions and a new PR.

jtilford:
> Hi All,
> 
> I am temporarily reverting this PR in v2.17.1 because it causes even more trouble (see #17536). I am open to suggestions and a new PR.

Thanks, @rstoenescu.

For the sake of exploring all the options, what would you think about restoring the pre-2.16.10 functionality, considering 2.16.10 as having introduced a bug? Then, adding some kind of opt-in configuration for anyone who needs the IME composition detection.

My alternative solution, to update beyond 2.16.9, is probably to create an internal copy of the 2.16.9 QInput component, and update all of my inputs to reference that copy instead of the up-to-date Quasar-provided input.

From my perspective, this PR #17476 introduces a bug. I don't have experience collaborating in FOSS, so pardon any lack of understanding the process from my end. But I would imagine the ideal path is to roll back the bug and find a different solution to it.

Thanks for taking any time at all to help consider this case. Huge respect for everyone contributing here, and I hope to be able to contribute code to a project like this in the future (or as soon as I figure out the proper ways to help, and find something I can actually help with).

jtilford:
> Hi All,
> 
> I am temporarily reverting this PR in v2.17.1 because it causes even more trouble (see #17536). I am open to suggestions and a new PR.

Thanks so much, @rstoenescu! 

I super appreciate your time on this. I wrote out a whole reply and realized I may have misinterpreted your message. It looks like you're reverting #17476, but for a moment I thought you were discarding #17543, when referencing "this PR" in your message.

Do you think it makes sense to add some kind of parameter, like `:ime-composition-detection="true"`, to switch on the change? At least, in that case, the feature would work, and we could document the known cases where it causes problems with the inputs, such as on Android 7 and 8.

Again, thank you for looking at this. 

Edit to add: I'm still pretty green when it comes to discussions on GitHub and using PRs in general, so I appreciate your understanding. I work on a team that pretty much develops on trunk without PRs, and I haven't had many opportunities to join conversations on FOSS projects.

NitroRCr:
> Hi All,
> 
> I am temporarily reverting this PR in v2.17.1 because it causes even more trouble (see #17536). I am open to suggestions and a new PR.

See discussion #17684 please

PR #17499, opened 2024-09-09
feat(ui) Add new prop 'option-data-id' to QSelect

lucashenkes:

- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [X] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs)  or explained in the PR's description.

- [X] A convincing reason for adding this feature (to avoid wasting your time, it's best to [start a new feature discussion](https://github.com/quasarframework/quasar/discussions/new?category=ideas-proposals) first and wait for approval before working on it)

**Other information:**

Currently, there is no easy way to find a specific option (e.g. using querySelector / in automated tests).
This PR adds the option to set a custom data-id attribute to the option-elements, making them easy to query.

### DIFF
--- a/ui/src/components/select/QSelect.js
+++ b/ui/src/components/select/QSelect.js
@@ -124,7 +124,8 @@ export default createComponent({
     virtualScrollItemSize: useVirtualScrollProps.virtualScrollItemSize.type,
 
     onNewValue: Function,
-    onFilter: Function
+    onFilter: Function,
+    optionDataId: [ String, Function ]
   },
 
   emits: [
@@ -341,6 +342,9 @@ export default createComponent({
           onClick: () => { toggleOption(opt) }
         }
 
+        if (props.optionDataId) {
+          itemProps[ 'data-id' ] = typeof props.optionDataId === 'string' ? `${ props.optionDataId }-${ i }` : props.optionDataId(opt)
+        }
         if (disable !== true) {
           optionIndex.value === index && (itemProps.focused = true)
 
--- a/ui/src/components/select/QSelect.json
+++ b/ui/src/components/select/QSelect.json
@@ -186,6 +186,30 @@
       "category": "options"
     },
 
+    "option-data-id": {
+      "type": [ "String", "Function" ],
+      "desc": "Add a 'data-id'-attribute to the option elements",
+      "params": {
+        "option": {
+          "type": [ "String", "Object" ],
+          "desc": "The current option being processed",
+          "examples": [
+            "'Tesla'",
+            "'iPhone'",
+            "{ label: 'Tesla', value: 'car' }"
+          ]
+        }
+      },
+      "returns": {
+        "type": "String",
+        "desc": "Use for custom data-id values"
+      },
+      "examples": [
+        "opt => (opt.value || opt).toLowerCase()"
+      ],
+      "category": "options"
+    },
+
     "menu-anchor": {
       "type": "String",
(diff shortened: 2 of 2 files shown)


peter-schunk-ion2s:
@rstoenescu would be nice if we could get feedback on this small addition. It's a blocker for test automation and would really help us and probably others. 
Thanks and regards

rstoenescu:
Would something like this suffice for you instead? It's a Vitest example, but I guess there are alternatives for other testing environments. It points to the actual content of the QDialog/QMenu/QTooltip.

```
// Quasar supplies this for some time now already
wrapper.findComponent({ name: 'QPortal' })
            .get('.q-dialog__inner')
```

peter-schunk-ion2s:
Thanks for the answer, @rstoenescu . This sure would work but it has some drawbacks.
Using the right option in the test framework (Cypress in our case) needs the selectors to go deep down all the way to the label to have its content (string-) compared to make sure the correct option will be selected. This is due to the fact that we cannot guarantee that the order and amount of available selections is always the same. Otherwise we could select it by index, sure.
Also it should be somewhat future proof and if the (HTML) structure of Quasar differs those tests would fail.
All in all we feel like using an optional option-data-id attribute would be less flaky and more consistent.

yusufkandemir:
It is possible to achieve this through the `#option` slot:
https://quasar.dev/vue-components/select#example--options-slot

Example code:
```html
<q-select ...>
  <template #option="scope">
    <q-item :data-id="scope.opt.value.toLowerCase()" v-bind="scope.itemProps">
      <q-item-section>
        <q-item-label>{{ scope.opt.label }}</q-item-label>
      </q-item-section>
    </q-item>
  </template>
</q-select>
```

This is not to say this PR doesn't provide any value. With the addition of a new prop, you can just use a simple prop instead of overriding the whole slot.

peter-schunk-ion2s:
Yeah the idea here is to have simple and straight forward solution instead of workarounds which might cause incompatibilities on the long run.

yusufkandemir:
To clarify things: What I've sent earlier is not a workaround and will not cause any problems in the future. It is the solution. What this PR adds is just a convenience feature not a fix.

Smenger:
I think this is exactly what it aimed for to be: A convenience feature. I don't see how this would affect anything or anyone negatively?

yusufkandemir:
First of all, we didn't discuss this together and decided to close the PR, but it was going to go that way naturally, and it did. It opens a can of worms, someone can say we want `data-cy`, `data-test-id`, etc. We recommend opening a feature request in GitHub Discussions to avoid situations like this, as written in the PR template.

I think a more generalized prop like `option-attributes`/`option-attrs` would work better. You'd return any attribute you like. Example: `:option-attrs="(opt) => ({ 'data-cy': 'option-item', 'data-cy-option-id': opt.id, class: 'my-option', style='foo: bar' })"`. @rstoenescu what do you think?

rstoenescu:
Sorry, but no. We start to drift into the bloatware category.

1. Slots exist exactly for customization purposes. It's not a "workaround" nor does it mean breaking changes further down the line.
2. If we are to add this prop then all other relevant components should have similar ones for consistency. But first, it must add real tangible value that cannot be achieved otherwise. It's not the case here.

PR #17522, opened 2024-09-19
feat(app-vite): use sass-embedded

yusufkandemir:

- Feature

- Maybe: if the system doesn't use Dart. There shouldn't be many systems out there that don't support Dart. See the system requirements [here](https://dart.dev/get-dart).

**Other information:**
Vite was defaulting to the `legacy` API which now produces warnings, see #17518. We decided to skip `modern` and directly go to `modern-compiler` which uses `sass-embedded`. It should have better performance.

### DIFF
--- a/app-vite/lib/config-tools.js
+++ b/app-vite/lib/config-tools.js
@@ -121,6 +121,9 @@ export async function createViteConfig (quasarConf, { compileId }) {
     build.viteVuePluginOptions
   )
 
+  /**
+   * @type {import('vite').UserConfig}
+   */
   const viteConf = {
     configFile: false,
     root: appPaths.appDir,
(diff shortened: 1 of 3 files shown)


jeremyVignelles:
> Does this PR introduce a breaking change?
>    Maybe: if the system doesn't use Dart. There shouldn't be many systems out there that don't support Dart. See the system requirements [here](https://dart.dev/get-dart).

stackblitz uses webcontainers which doesn't seem to support dart.

yusufkandemir:
@jeremyVignelles yes, that's why we have these workarounds here:
https://github.com/quasarframework/stackblitz-templates/blob/8afec4b1b339f79dc9905535ffd7a35f32ee32a9/app-vite-v2/package.json#L36-L38
https://github.com/quasarframework/stackblitz-templates/blob/8afec4b1b339f79dc9905535ffd7a35f32ee32a9/app-vite-v2/quasar.config.js#L59-L71

PR #17527, opened 2024-09-20
fix(ui): prevent disabled styling on disabled=false

Evertvdw:
Currently the default Quasar styling will also style elements that have `disabled=false` as attribute value due to the generic `[disabled]` attribute selector that is used. This is not correct behavior, this PR fixes that.

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/ui/src/css/core/visibility.sass
+++ b/ui/src/css/core/visibility.sass
@@ -27,12 +27,12 @@
 
 .readonly
   cursor: default !important
-.disabled, [disabled]
+.disabled, [disabled]:not([disabled=false])
   &,
   *
     outline: 0 !important
     cursor: not-allowed !important
-.disabled, [disabled]
+.disabled, [disabled]:not([disabled=false])
   opacity: .6 !important
 
 .hidden


rstoenescu:
Thanks!

rstoenescu:
This will go into Quasar 2.17.1

rstoenescu:
Reverting because:
1. The documentation for "disabled" attribute does not say it is a boolean. It just mentions that it has effect when it is present. https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled
2. The selector in visibility.sass becomes too specific and breaks things all over the UI.

Evertvdw:
> Reverting because:
> 
> 1. The documentation for "disabled" attribute does not say it is a boolean. It just mentions that it has effect when it is present. https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled
> 2. The selector in visibility.sass becomes too specific and breaks things all over the UI.

Reading the Html spec it seems that the disabled=false pattern is just bad Html. I will contact the external tool which has this issue to get them to just omit the disabled attribute when it should not be disabled.

Thanks!

rstoenescu:
Exactly. Yes

ReneBrandt:
I was on the verge of creating a new issue for this (let me know if I should!), but I think the current state (not being able to bind a boolean directly to `disabled`) is breaking with Vue's Boolean Casting and how this normally works in vanilla Vue. Note Vue specifically uses this prop `disabled` as an example in the docs, stating `:disabled="false"` is equivalent to not providing it at all. That is no longer the case now when working in Quasar.

I also think the styling oversteps its boundaries, applying the disabled visualization also to plain <div> components. See https://jsfiddle.net/35jz9am7/31/, which makes it even more tricky.

I will move away from using disabled for now as much as possible, but 1) it should be considered a breaking change, and 2) I would request this to be reconsidered in the future, due to the breakage with very typical Vue convention (despite this not being in-line with the HTML spec) and thus possibly breaking with lots of libraries applying this typical Vue use-case

PR #17561, opened 2024-10-04
Add tip on `cypress.config.ts` to testing-and-auditing documentation

joshbuker:
If you use both Cypress and another testing framework that implements `expect` and other common keywords, you may need to exclude `cypress.config.ts` from your `tsconfig.json` to prevent ESLint from getting confused.

- [ ] Bugfix
- [ ] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/docs/src/pages/quasar-cli-vite/testing-and-auditing.md
+++ b/docs/src/pages/quasar-cli-vite/testing-and-auditing.md
@@ -35,6 +35,24 @@ If you ever need to review your installation choices you can take a look at `qua
 
 > Note that we previously suggested to use `@quasar/testing` AE to manage all testing harnesses in a project. This is no longer the case, as [it is now deprecated](https://github.com/quasarframework/quasar-testing/tree/dev/packages/testing/README.md#DEPRECATION-NOTICE). Please use the above commands instead.
 
+> [!TIP]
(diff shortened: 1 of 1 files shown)


joshbuker:
See: https://github.com/cypress-io/cypress/issues/22059#issuecomment-1148921141

arnonrdp (approved):
I like this.

yusufkandemir:
Hi, thanks for contributing 👏 If this problem and solution still exist, the snippet needs to be updated to utilize `quasar.config.ts > build > extendTsConfig` instead.

But, I believe it would be better to mention or even automatically handle this in https://github.com/quasarframework/quasar-testing

/cc @IlCallo

IlCallo:
This happens when using two frameworks which leverage global types functions with the same name, it's a know issue

While Cypress doesn't really allow to avoid this AFAIK, Jest and Vitest do provide helpers to explicitly import those functions

We scaffold Jest and Vitest AE files to avoid using global types, and you should too: https://github.com/quasarframework/quasar-testing/tree/dev/packages/unit-jest#import-jest-helpers-from-jestglobals-optional-but-strongly-recommended

PR #17755, opened 2025-01-14
fix(ui): fix QOptionGroup options type (fix: #17710)

### LINKED ISSUE #17710: `QOptionGroup` invalid type of `options` for typescript
### What happened?

When using the **QOptionGroup** component and `options` in conjunction with `option-value`/`option-label` (`options` list items have different field names than `value`/`label`), a typescript type checking error occurs, e.g.:
```ts
Type '{ id: number; name: string; }[]' is not assignable to type '{ [props: string]: any; label: string; value: any; disable?: boolean | undefined; }[]'.
```

I expect that it's possible to put a list of any objects into `options`, as is implemented e.g. in **QSelect**.

### Reproduction URL

https://stackblitz.com/edit/vite-cywbftpt?file=src%2FApp.vue

### How to reproduce?

```ts
<script setup lang="ts">
  import { ref } from 'vue'
  const mod = ref([])
  const opts = [{ id: 1, name: 'foo' }, { id: 2, name: 'bar' }]
</script>

<template>
  <q-option-group v-model="mod" :options="opts" option-value="id" option-label="name" />
</template>
```

### Flavour

### Areas

Components (quasar), TypeScript Support

### Platforms/Browsers

_No response_

### Quasar info output

_No response_

### Relevant log output

_No response_

### Additional context

In fact it works in a dev environment, it's just the `options` have the wrong type.

yusufkandemir:
to respect option-label and option-value props

- Bugfix

- No

**Other information:**
Fixes #17710

### DIFF
--- a/ui/src/components/option-group/QOptionGroup.json
+++ b/ui/src/components/option-group/QOptionGroup.json
@@ -17,22 +17,6 @@
       "desc": "Array of objects with value, label, and disable (optional) props. The binary components will be created according to this array; Props from QToggle, QCheckbox or QRadio can also be added as key/value pairs to control the components singularly",
       "default": "[]",
       "definition": {
-        "label": {
-          "type": "String",
-          "desc": "Label to display along the component",
-          "required": true,
-          "examples": [ "'Option 1'", "'Option 2'", "'Option 3'" ]
-        },
-        "value": {
-          "type": "Any",
-          "desc": "Value of the option that will be used by the component model",
-          "required": true,
-          "examples": [ "'op1'", "'op2'", "'op3'" ]
-        },
-        "disable": {
-          "type": "Boolean",
-          "desc": "If true, the option will be disabled"
-        },
         "...props": {
           "type": "Any",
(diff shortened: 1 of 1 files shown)


rstoenescu:
The information is valuable in the JSON displayed in docs & also in the web-types etc. Rather than dropping all those props, let's add a "tsType: false" to each of those props -- so that only our TS build script will ignore it.

yusufkandemir:
I went with a solution similar to the QSelect options prop. The documentation explains the rest. Furthermore, having them there might be confusing when keeping `option-value`, `option-disable`, etc. props in mind. If you still think we should add new logic around `tsType` to accept `false` and behave differently, I, unfortunately, don't have time for it to implement, test, etc. I also think it will open a new rabbit hole, to adapt it to some other places.

rstoenescu:
Went with a more appropriate description for the "options" for both QOptionGroup & QSelect.

yusufkandemir:
Reference: https://github.com/quasarframework/quasar/commit/d07b4febac1e5be9c8a56dcb7df76699fc4cf338

PR #17978, opened 2025-04-21
feat(ui): add indicator-shape prop to QTabs

stefanvanherwijnen:

- [ ] Bugfix
- [x] Feature
- [x] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Material Design 3 introduces a [navigation rail](https://m3.material.io/components/navigation-rail), which is similar to (Q)Tabs but the active indicator is different (pill shape vs line). This PR adds a `indicator-shape` prop to QTabs which makes it possible to use the pill shape active indicator and use QTabs for a navigation rail.

gif

### DIFF
--- a/docs/src/examples/QTabs/CustomIndicator.vue
+++ b/docs/src/examples/QTabs/CustomIndicator.vue
@@ -54,6 +54,17 @@
         <q-tab name="alarms" icon="alarm" label="Alarms" />
         <q-tab name="movies" icon="movie" label="Movies" />
       </q-tabs>
+
+      <q-tabs
+        v-model="tab"
+        indicator-shape="pill"
+        indicator-color="lime"
+        class="bg-red text-dark shadow-2"
+      >
+        <q-tab name="mails" icon="mail" label="Mails" />
+        <q-tab name="alarms" icon="alarm" label="Alarms" />
+        <q-tab name="movies" icon="movie" label="Movies" />
+      </q-tabs>
     </div>
   </div>
 </template>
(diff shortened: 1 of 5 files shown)


rstoenescu:
Hi Stefan, thank you for the work here and for the clean, well-presented PR. The MD3 navigation-rail look is a legitimate use case, but after reviewing it we've decided not to take this in, for a few reasons:

1. **The pill geometry is hardcoded** (52x32px, centered via `calc(50% - 26px)`, 24px in dense mode). It doesn't track the tab's actual content width, so wide or inline-label tabs get a fixed-size pill over arbitrary content, while the real MD3 spec sizes the pill to its content. Baking one set of pixel values into the component API is something users would immediately need to override.
2. **It changes all tabs, not just pill ones.** The `z-index: 2` added to `.q-tab__content` applies unconditionally, altering paint order for every existing QTab and potentially breaking user CSS that layers content over tabs.
3. **Several interactions are left unhandled**: vertical tabs keep the horizontal pill geometry (only the line variant got a vertical override), `narrow-indicator` and `switch-indicator` silently do nothing with the pill, and the indicator slide animation scales the element by the old/new tab width ratio, which distorts the pill's border radius mid-transition whenever tab widths differ.
4. **The look is already achievable in userland.** `.q-tab__indicator` is a plain div drawing with `currentColor`, so a pill is a handful of CSS lines on top of `indicator-color`, along the lines of the "Custom indicator" docs examples. That keeps the API surface small and lets each app pick its own pill sizing.

More broadly, adopting MD3 visuals is a theme-wide decision for Quasar rather than something we want to introduce one prop at a time. A docs example demonstrating the pill via CSS would be a welcome contribution if you'd like to pursue that route.

Thanks again for contributing, and sorry it took this long to give the PR a proper review.

PR #17979, opened 2025-04-21
feat(QBtn): add left right classes when label is slot

muhamadamin1992:

- [ ] Bugfix
- [x ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

**Other information:**

### DIFF
--- a/ui/src/components/btn/QBtn.js
+++ b/ui/src/components/btn/QBtn.js
@@ -302,7 +302,7 @@ export default createComponent({
       props.icon !== void 0 && inner.push(
         h(QIcon, {
           name: props.icon,
-          left: props.stack !== true && hasLabel.value === true,
+          left: props.stack !== true && (hasLabel.value === true || slots.label !== undefined),
           role: 'img'
         })
       )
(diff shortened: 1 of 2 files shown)


rstoenescu:
Thank you for taking the time to contribute, and apologies for the long silence on this one. Closing after review: QBtn has no "label" slot (only "default" and "loading"), so the `slots.label !== undefined` check never fires for regular slot content and the change ends up having no effect. Even if it checked `slots.default`, the default slot is commonly used on icon-only buttons for QTooltip/QMenu/QBadge, and those would then get an unwanted icon margin, since the render function cannot distinguish text content from popup content. That is why the margin is tied to the `label` prop. For slot content, adding `q-ml-sm`/`q-mr-sm` (or using the `label` prop) is the recommended approach. If you have an idea for an opt-in API that avoids the pitfall above, feel free to open a discussion or a fresh PR.

PR #17795, opened 2025-01-28
fix(app-vite): local import of extensions in windows

lvf23:
- Possibly fix the issue documented here:

"There have been many reports of problems linking via file: and link:. This is outside of Quasar’s reach, but is likely to do with your development environment, aka your mileage with Windows will vary."

https://quasar.dev/app-extensions/development-guide/introduction#manually-testing

- Now it's possible to load Quasar App Extensions without host in platforms like Github, Gitlab, etc.

- Fix errors like for local Quasar App Extension:

App • ⚠️  App Extension "[ext-name]" has missing index script...

- Tested in Windows 10 with local Quasar App Extension

**What kind of change does this PR introduce?** 

- [x] Bugfix
- [ ] Refactor
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** 

- [ ] Yes
- [x] No

**The PR fulfills these requirements:**

- [ ] It's submitted to the `dev` branch (or `v[X]` branch)
- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)
- [ ] It's been tested on a Cordova (iOS, Android) app
- [ ] It's been tested on an Electron app
- [ ] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs)  or explained in the PR's description.

If adding a **new feature**, the PR's description includes:
- [x] A convincing reason for adding this feature (to avoid wasting your time, it's best to [start a new feature discussion](https://github.com/quasarframework/quasar/discussions/new?category=ideas-proposals) first and wait for approval before working on it)

**Other information:**

Any adaptation needed do not hesitate to call me. I could adjust it or give more feedback. Feel free to edit according to the better fit for the project.

PR #18003, opened 2025-05-12
fix(app-vite&webpack): properly resolve AE script paths

yusufkandemir:

- Bugfix

- No

**Other information:**
Supersedes #17795

- use `/` instead of `join` for import paths:
  - `import('quasar-app-extension-foo\\dist\\index.js')` is not valid. `join` uses the separate according to the OS
  - similar to how `getPackageJson` is implemented without using `join`
- `packagePath = (... || getPackageScriptPath(...))` would end up `.../pkg-path/dist/index.js`, so `this.#packagePath` was sometimes pointing out to `src`/`dist`, rather than the package root.
  - `this.#packagePath` no longer exists, so this no longer needs to be fixed with `resolve(..., '..')`
- always use `appDir` when resolving AE scripts
- add code comments to clarify the behavior

### DIFF
--- a/app-vite/lib/app-extension/AppExtensionInstance.js
+++ b/app-vite/lib/app-extension/AppExtensionInstance.js
@@ -1,4 +1,4 @@
-import { join, relative, resolve, dirname } from 'node:path'
+import { relative, resolve } from 'node:path'
 import { pathToFileURL } from 'node:url'
 import fse from 'fs-extra'
 import inquirer from 'inquirer'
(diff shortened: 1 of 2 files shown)


DouglasCalora:
I tested linking localy (quasar ext invoke) on projects:

- app-vite v2: worked ✅
- app-webpack v4: worked ✅

app-webpack v3 will no longer be supported?

Thanks for the efford!

DouglasCalora (approved):

yusufkandemir:
Thanks for confirming. I am hoping for another confirmation, especially from someone using Windows.

app-vite v1 and app-webpack v3 are legacy now, we won't be making more updates to them.

eliasassuncao:
I also tested linking locally (quasar ext invoke) on Windows with the following projects:
- app-vite v2: worked ✅
- app-webpack v4: worked ✅

Thanks for the support!

DouglasCalora:
Note: I did not tested this by adding extension without linkin it localy.

On app-webpack v3, I just changed to this code:

```js
    return getPackageScriptPath(
      this.packageFullName,
      scriptName,
      appPaths.appDir
    )
```

And worked the link.

DouglasCalora:
@yusufkandemir any news on releasing date? Two peope tested in windows already.

yusufkandemir:
@DouglasCalora I originally scheduled it last week, but other work-related stuff got in the way. For this week, I was planning to do it tomorrow. So, if everything goes as expected, I will do it tomorrow.

DouglasCalora:
@yusufkandemir Thanks

yusufkandemir:
Thanks, everyone, for cooperating!

Released in [@quasar/app-vite-v2.2.1 (release)](https://github.com/quasarframework/quasar/releases/tag/%40quasar%2Fapp-vite-v2.2.1) and [@quasar/app-webpack-v4.2.1 (release)](https://github.com/quasarframework/quasar/releases/tag/%40quasar%2Fapp-webpack-v4.2.1) This is the first set of releases I did 🎉

Issue #18075, opened 2025-07-22
Move `src` and `app` aliases to the end of `path` section in `.quasar/tsconfig.json`


FragsterAt:
### What happened?

They placed at the beginning of `path` section, so automatic imports (at least in vs code) starts with `'src/...`

### What did you expect to happen?

If they'll be moved to the end of `path` section, automatic imports finds first more compact variant of import, i.e. `stores/store` instead of `src/stores/store`, that makes code less 'noisy'.

### Reproduction URL

https://stackblitz.com/edit/quasarframework-stackblitz-templates-dtyc5iwp?file=src%2Fpages%2FIndexPage.vue&title=Quasar%20Framework%20-%20app-vite%20v2

### How to reproduce?

delete `.quasar` folder in project
run `npx quasar prepare`
look in `.quasar/tsconfig.json`

<img width="418" height="459" alt="Image" src="https://github.com/user-attachments/assets/6ee22ce8-4752-409f-a9f3-615870e192fb" />

this must be at the end of section for short auto imports

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Quasar CLI Commands/Configuration (@quasar/cli | @quasar/app-webpack | @quasar/app-vite)

### Quasar info output

```Shell

```

### Relevant log output

```Shell

```

stefanvanherwijnen:
Does `stores/store` import `src/stores/store` or `app/stores/store`?

FragsterAt:
> Does `stores/store` import `src/stores/store` or `app/stores/store`?

@stefanvanherwijnen  they are all aliases for the same file: `project_folder/src/stores/store.ts` or '.js'

Autoimport in VS Code check aliases from first to last and use `src/stores/store` now and `stores/store` after fix.

stefanvanherwijnen:
> > Does `stores/store` import `src/stores/store` or `app/stores/store`?
> 
> [@stefanvanherwijnen](https://github.com/stefanvanherwijnen) they are all aliases for the same file: `project_folder/src/stores/store.ts` or '.js'
> 
> Autoimport in VS Code check aliases from first to last and use `src/stores/store` now and `stores/store` after fix.

I believe `app` points to the project folder. My point is that you can't just remove the alias, or you would need to alias `stores`. Keep in mind that it's not just TS aliases, but also Vite aliases.

FragsterAt:
[...]

PR #18076, opened 2025-07-22
fix(app-vite&app-webpack): move `src` and `app` aliases to the end of tsconfig paths (fix: #18075)

FragsterAt:
fixes 

https://github.com/quasarframework/quasar/issues/18075

by sorting object keys in types-generator.ts

### DIFF
--- a/app-vite/lib/quasar-config-file.js
+++ b/app-vite/lib/quasar-config-file.js
@@ -776,9 +776,6 @@ export class QuasarConfigFile {
       },
 
       alias: {
-        '#q-app': '@quasar/app-vite',
-        src: appPaths.srcDir,
-        app: appPaths.appDir,
         components: appPaths.resolve.src('components'),
         layouts: appPaths.resolve.src('layouts'),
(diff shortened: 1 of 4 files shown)


yusufkandemir (changes requested):
Thanks for reporting and solving the problem. However, I am thinking of a more robust solution. Will deal with that later.

FragsterAt:
@yusufkandemir , I think, my solution is the simpliest, because the other way is to change `quassar-config-file.js` to use merge build.config.alias in different order - at first - default aliases for components, stores, etc, next aliases from config and at the end - `#q-app`, `src`, `app`. something like that:

изображение

in this case `src` and `app` aliases are staying before `#q-app/wrappers` and others (but it doesn't seem to matter).

If you prefer that way - I can make changes in pr.

FragsterAt:
the negative effect is the impossibility of changing the aliases of `src` and `app` (`#q-app` will be overwritten in any case)

rstoenescu:
Hi! This has been tackled, so closing this. Thanks for contributing though!

PR #18105, opened 2025-08-25
fix(ui): change clearable node from icon to button #18084

### LINKED ISSUE #18084: QInput clearable property breaks accessibility by default
### What happened?

If your form has QInput and you have set **clearable** property to true by default this adds icon button where you can clear the input content. However the way it does it (at least in material design) is that it creates: 
`<i class="q-icon notranslate material-icons q-field__focusable-action" aria-hidden="false" tabindex="0" role="button" aria-label="Clear">cancel</i>`
element where aria-label is set to Clear while button content is named as cancel. This breaks A -level WCAG rules where button content and label should be the same.

Aria-label should be same as button content (title) to fill WCAG requirements or actual clear icon should be shown as an SVG element or instead of using just icon use real button tag for the button itself.

### Reproduction URL

https://codepen.io/jannhama/pen/gbaWvRw?editors=1010

### How to reproduce?

1. Go to the provided link and make sure you have content visible
2. Open debugger and check what tags are formed for the clear icon
3. You should see similar tag as provided above.

### Flavour

### Areas

Accessibility [a11y] (quasar)

### Platforms/Browsers

Chrome

### Quasar info output

```Shell
No Quasar info available: Quasar version is 2.18.2, node v22.14.0 and Vue version is 3.5.13
```

### Relevant log output

```Shell

```

### Additional context

_No response_

SamVanini:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

This PR replaces the rendering of a QIcon with a QBtn, in order to stick to WCAG rules.
This chage aims to use semantic HTML instead of roles and ARIA to provide accessible components out of the box, without changing icon library or type
Every input component created using use-field composable is affected by this modification

Closes #18084

### DIFF
--- a/ui/src/composables/private.use-field/use-field.js
+++ b/ui/src/composables/private.use-field/use-field.js
@@ -2,6 +2,7 @@ import { h, ref, computed, Transition, nextTick, onActivated, onDeactivated, onB
 
 import QIcon from '../../components/icon/QIcon.js'
 import QSpinner from '../../components/spinner/QSpinner.js'
+import QBtn from '../../components/btn/QBtn.js'
 
 import useId from '../use-id/use-id.js'
 import useSplitAttrs from '../use-split-attrs/use-split-attrs.js'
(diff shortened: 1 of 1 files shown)


rstoenescu:
Thank you for the PR and for pushing on accessibility, it is appreciated.

Closing this one though, because #18084 ended up being addressed differently as part of our broader ARIA work: since 4c2b9707c (shipped in Quasar v2.25.0), QIcon renders ligature text (Material Icons/Symbols) inside its own aria-hidden wrapper. The clearable action now exposes only its localized aria-label, so there is no "cancel" content competing with the accessible name (WCAG 2.5.3), and the node already implements the complete button pattern: role="button", tabindex, keyboard activation.

We also deliberately avoided swapping the QIcon for a QBtn here. It would change the rendered DOM of every component built on the field composable (QInput, QSelect, QFile, ...), breaking user CSS and tests that target it, and it would pull in QBtn's ripple, padding and min-width styling, altering the visuals of the clearable action everywhere.

Thanks again for contributing!

Issue #18091, opened 2025-08-11
'Closing popup' is propagated for all Selects components


GuilhermeErmel:
### What happened?

I've already registered an isse but it was not resolved. And it's still an UX problem for my application, I couldn't create a good workaround. (https://github.com/quasarframework/quasar/issues/16207)

In my app it's possible to open a dialog from a Select component to create a new Option. In this dialog, there is another Select. If this inner Select is opened, if I want to close it (for example, clicking outside the popup), both selects are closed.

I'm creating this new isse, because the other looks like abandoned

### What did you expect to happen?

If I clicked outside the "popup dialog" just that Select popup is closed.

### Reproduction URL

https://codepen.io/ermel/pen/zYyOQVq

### How to reproduce?

Click in "external" Select
Click in Button ('Click me') - to open the Dialog/Modal
Click in "inner" Select
Try to close just the 'inner Select' (without select an option), for example, clicking inside Dialog/Modal

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Components (quasar)

### Quasar info output

```Shell

```

### Relevant log output

```Shell

```

PR #18134, opened 2025-09-23
fix(QMenu): prevent nested menus from closing on unrelated click-outside (#18091)

IceIsFrozen:

- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

Since click-outside was processing all registered handlers it was closing everything nested on menus. Added a small check to prevent this from happening so only the inner closes and acts more like esc key method. If wanted could make it so the flag only happens if it notices q-dialog but since the menus all close properly depending on where you click, shouldn't be needed.

### DIFF
--- a/ui/src/components/menu/QMenu.js
+++ b/ui/src/components/menu/QMenu.js
@@ -118,6 +118,11 @@ export default createComponent({
         if (props.persistent !== true && showing.value === true) {
           hide(e)
 
+          // prevent further processing of click-outside handlers
+          if (props.separateClosePopup === true) {
+            e.qSeparateClosePopup = true
+          }
+
           if (
             // always prevent touch event
             e.type === 'touchstart'
--- a/ui/src/utils/private.click-outside/click-outside.js
+++ b/ui/src/utils/private.click-outside/click-outside.js
@@ -63,6 +63,11 @@ function globalHandler (evt) {
       // used to prevent refocus after menu close
       evt.qClickOutside = true
       state.onClickOutside(evt)
+
+      // used to prevent closing all if nested
+      if (evt.qSeparateClosePopup === true) {
+        return
+      }
     }
     else {
       return


MILLERMARRU (commented):
Checked `separateClosePopup`'s existing usage before reviewing this, since it's not a new prop, and this fix is actually just extending an existing precedent rather than introducing new behavior. `portal.js`'s own cascading-close walk already treats it as a stop point (`if (proxy.$props.separateClosePopup) { return getParentProxy(proxy) }`), that's the ESC-key/programmatic-hide path. This PR brings `click-outside.js`'s `globalHandler` loop in line with that same semantic for mousedown/touchstart, which is exactly the "acts more like esc key method" described in the PR notes, not a new convention.

The mechanism itself checks out: `evt` is the same object passed through the whole `registeredList` loop, so setting `evt.qSeparateClosePopup = true` inside `hide()` (called synchronously via `state.onClickOutside(evt)`) is visible to the `return` check in `globalHandler` immediately after, on the same iteration.

One thing worth adding: there's an existing `mountNestedMenus({ separateClosePopup: true })` helper already used in `QMenu.test.js`, but this PR doesn't add a click-outside-specific regression test with it, only the ESC-path behavior seems covered currently. Given the bug is specifically about click-outside cascading (not ESC), a test exercising that exact loop would be good to have alongside the fix.

rstoenescu:
Thank you for digging into this and for the PR — your diagnosis of the click-outside handler walking the whole popup stack was spot on.

Closing this as superseded: #18091 was fixed on dev by #18412 (commit 8a0a91379, shipped in Quasar v2.23.0). Interestingly, the approach that landed is essentially the refinement you suggested in your own PR description ("could make it so the flag only happens if it notices q-dialog") — the click-outside scan now treats a modal (non-seamless) QDialog as a shield, so a click inside the dialog only closes popups opened after it, and popups behind the backdrop stay open. Your exact scenario (Select → Dialog → Select) is covered by a regression test in ui/src/utils/private.click-outside/click-outside.test.js.

For context on why we went with dialog-based scoping rather than the separateClosePopup flag: QSelect, QBtnDropdown and QPopupProxy all set separateClosePopup: true internally on their menus, so gating on it wouldn't have been opt-in — it would have changed outside-click dismissal to one-level-at-a-time for those components everywhere, even with no dialog involved. The prop's documented contract is also specific to v-close-popup chains.

If you still hit the problem on v2.23.x, please let us know and we'll reopen the investigation. Thanks again!

Issue #16735, opened 2024-01-01
Q-Date does not honor date mask when using events


shinynet:
### What happened?

When utilizing events, the date picker ignores the chosen date format mask and appears to only work with YYYY/MM/DD formatted dates.

### What did you expect to happen?

If a custom date mask is applied (such as YYYY-MM-DD), then I'd expect to pass an array of dates in the same format and receive date arguments in the same format

### Reproduction URL

https://stackblitz.com/edit/quasarframework-webpack-3qmxcq?file=src%2Fpages%2FIndexPage.vue

### How to reproduce?

1. View the IndexPage in the reproduction link. 
2. Notice events formatted as the mask don't show on the calendar.

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Components (quasar)

### Quasar info output

```Shell
Operating System - Linux(5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36 Edg/120.0.0.0) - linux/x64
NodeJs - 18.18.0

Global packages
  NPM - 9.4.2
  yarn - 1.22.19
  @quasar/cli - undefined
  @quasar/icongenie - Not installed
  cordova - Not installed

Important local packages
  quasar - 2.6.2 -- Build high-performance VueJS user interfaces (SPA, PWA, SSR, Mobile and Desktop) in record time
  @quasar/app-webpack - 3.4.7 -- Quasar Framework App CLI with Webpack
  @quasar/extras - 1.13.4 -- Quasar Framework fonts, icons and animations
  eslint-plugin-quasar - Not installed
  vue - 3.2.31 -- The progressive JavaScript framework for building modern web UI.
  vue-router - 4.0.14
  pinia - Not installed
  vuex - Not installed
  electron - Not installed
  electron-packager - Not installed
  electron-builder - Not installed
  @babel/core - 7.17.8 -- Babel compiler core.
  webpack - 5.70.0 -- Packs CommonJs/AMD modules for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand. Support loaders to preprocess files, i.e. json, jsx, es7, css, less, ... and your custom stuff.
  webpack-dev-server - 4.7.4 -- Serves a webpack app. Updates the browser on changes.
  workbox-webpack-plugin - Not installed
  register-service-worker - 1.7.2 -- Script for registering service worker, with hooks
[...]

PR #18136, opened 2025-09-23
feat(QDate): support different date formats when using events (#16735)

IceIsFrozen:

- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

Adds support for dashes(ISO 8601) and dots by normalizing them into quasar's desired slash date format. Everything should probably normalize into ISO 8601 format in the future but this works fine.

2019/02/01 - Originally supported
2019-02-01 - Support added
2019.02.01 - Support added

### DIFF
--- a/ui/src/components/date/QDate.js
+++ b/ui/src/components/date/QDate.js
@@ -507,7 +507,10 @@ export default createComponent({
       else {
         const fn = typeof props.events === 'function'
           ? props.events
-          : date => props.events.includes(date)
+          : (() => {
+              const normalizedEvents = props.events.map(e => String(e).replace(/[.\\-]/g, '/'))
+              return date => normalizedEvents.includes(date)
+            })()
 
         for (let i = 1; i <= daysInMonth.value; i++) {
           const dayHash = viewMonthHash.value + '/' + pad(i)


rstoenescu:
Thank you for taking the time to work on this and for the clean, minimal diff, it is genuinely appreciated.

After reviewing it, we have decided not to go down this route, for two reasons:

1. It would not actually cover #16735. Normalizing `-` and `.` into `/` only handles separator variants of the same `YYYY/MM/DD` ordering, while the issue is about honoring the `mask` prop. A mask such as `DD-MM-YYYY` or `MM/DD/YYYY` would still not match, and after normalization an entry like `01-02-2019` becomes `01/02/2019`, which can silently match the wrong day (or nothing) depending on ordering.
2. It would widen the accepted format for the `events` array alone, while `options`, `default-year-month`, `navigation-min/max-year-month` and the String received by `event-color` keep the strict format. That asymmetry would make the API harder to reason about than the current strict but consistent rule.

The supported way to use custom formats is the function form of `events` / `options`: it receives the `YYYY/MM/DD` String and you can convert/compare against any format you need (or simply map your array once before passing it). We have clarified this contract in the docs and in the API cards (commit efc6c6864a41559032b886e7b0b5de3821edab39, landing on `dev`).

Closing this one, but thank you again for contributing. We hope to see more PRs from you!

Issue #18151, opened 2025-10-16
Inconsistent behaviour in element height, when using column with row due to whitespace, e.g. when writing "Lorem ipsum" instead of "Loremipsum".


klauskades:
### What happened?

Using column in combination with row and col-6. The height of the col-6 element is different, depending on the usage of whitespaces in the text. Problem becomes worse, when having multiple lines.

```
<div id="q-app">
  <div class="column">
    <div class="row">
      <div class="col-6 bg-blue">
        <div>
           <!-- With "Lorem ipsum" one line extra space is added at the bottom. With "Loremipsum" no extra space is added -->
          Lorem ipsum
        </div>
      </div>
    </div>
  </div>
</div>
```

<img width="641" height="23" alt="Image" src="https://github.com/user-attachments/assets/1d7cb269-73eb-4c4a-8730-e8cd756ca421" />

<img width="641" height="21" alt="Image" src="https://github.com/user-attachments/assets/378054e9-a2d7-4e3b-91c8-ed90fbe00e4d" />

I am happy about any solution!

### What did you expect to happen?

I would expect the col-6 element to have always the same height - only one line. 

### Reproduction URL

https://codepen.io/Klaus-Kades/pen/pvgdNMx

### How to reproduce?

1. Go to the provided Codepen
2. Switch between 'Loremipsum' or 'Lorem ipsum'
3. Add `v-for="x in 30" :key="x"` to the Lorem ipsum div to make it more obvious.

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Components (quasar)

### Quasar info output

```Shell
Operating System        Darwin(25.0.0) - darwin/arm64
NodeJs                  23.9.0

Global packages    
  NPM                   10.9.2
  yarn                  Not installed
  pnpm                  9.13.0
  bun                   Not installed
  @quasar/cli           2.5.0
  @quasar/icongenie     4.0.0
  cordova               Not installed

Networking         
  Host                  localhost
```

### Relevant log output

```Shell

```

### Additional context

I use column in combination with col a lot to use the full size of the screen. On one of our pages, this issue is leading to unnecessary vertical scrolling, because multiple lines like this, sometimes add even more unnecessary vertical space to the element. You can test this with:

        <div v-for="x in 30" :key="x">
           
          Lorem ipsum
        </div>

rstoenescu:
[...]

PR #18153, opened 2025-10-17
fix(ui): Rows inside columns now correctly take full width of their parent (#18151)

IceIsFrozen:

- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**
This is a pretty straight forward fix that seems to more or less match how other frameworks do things.

### DIFF
--- a/ui/src/css/core/flex.sass
+++ b/ui/src/css/core/flex.sass
@@ -57,6 +57,8 @@
   flex-direction: column
   &.reverse
     flex-direction: column-reverse
+  >.row
+    width: 100%
 
 .wrap
   flex-wrap: wrap


rstoenescu:
Thank you for the PR and for digging into #18151. After reviewing it we have decided not to take this route, so I am closing it. The reasoning:

The behavior described in the issue is standard flexbox, not a Quasar bug. Quasar's `.row`/`.column`/`.flex` intentionally default to `flex-wrap: wrap`. Per the CSS flexbox spec, in a multi-line (wrapping) flex container `align-items: stretch` stretches items to their flex line, which is sized from content, not to the container. That is why a `.row` inside a `.column` sizes to its content.

Shipping `.column > .row { width: 100% }` in core CSS would silently change layout for every existing app that relies on content-sized rows inside columns. A concrete example: `.column.items-center > .row` currently shrinks to its content and gets centered; with this rule the row becomes full width and the centering visually disappears. There is no opt-out short of overriding core styles, so this is a visual breaking change we cannot take in the grid system.

It would also patch only one combination. The same per-line sizing applies to `.column > .flex`, `.column > .column`, `.flex > .row` and so on, so fixing only `.column > .row` would make equivalent structures behave inconsistently.

The intended result is already one class away: add `no-wrap` to the column (a single-line flex container stretches items to the container's full cross size), or add `full-width` to the row.

On the "other frameworks do this" note: Bootstrap's `.row` is full width because its parent container lays out as a block, so the row is an ordinary block box; that does not translate to a wrapping flex column parent.

Thanks again for contributing, and the investigation here was genuinely useful for documenting the correct remedy over in #18151.

Issue #18008, opened 2025-05-14
When using `q-expansion-item` with `switch-toggle-side` colors aren't switched


AlbanBernard:
### What happened?

I am using `q-expansion-item` with the `switch-toggle-side` option. The icon and the extend arrow switched places but not the color, the icon is grey and the arrow is black.

### What did you expect to happen?

The icon would stay black, as is it was on the left when not using `switch-toggle-side`. The arrow would stay grey, as it was on the right when not using `switch-toggle-side`.

### Reproduction URL

https://codepen.io/AlbanBernard/pen/vEEbWxE

### How to reproduce?

Go to the provided codepen link (first time i do that, hope it works).
Look at the colors of the non-switched and of the switched version

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Components (quasar)

### Platforms/Browsers

Microsoft Edge

### Quasar info output

```Shell
Operating System        Windows_NT(10.0.22631) - win32/x64
NodeJs                  22.13.0

Global packages
  NPM                   10.9.2
  yarn                  1.22.22
  pnpm                  Not installed
  bun                   Not installed
  @quasar/cli           2.5.0
  @quasar/icongenie     Not installed
  cordova               Not installed

Networking
  Host                  LYOTRAINEE19
  Ethernet              10.110.3.47
```

### Relevant log output

```Shell

```

rstoenescu:
Confirmed and fixed in f205ed3f1e2986d76f8d5030ae888c62e1b07e0a. Root cause: the header sections' colors are tied to their side/avatar CSS roles, and `switch-toggle-side` used to swap those roles along with the positions. Each section now keeps its role on either side, so the `icon` keeps inheriting the item color and the expand arrow keeps its grey, in both light and dark mode, with the layout unchanged. Will ship with the next Quasar UI release.

PR #18154, opened 2025-10-18
fix(ui): switch-toggle-side colors are now switched (#18008)

IceIsFrozen:

- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**
Added some checks to the SASS on if the toggle icon is there or not. Then for when it's not the toggle icon, important(ed) the inherited color so it nicely comes through. Tested the changes in both light and dark mode.

### DIFF
--- a/ui/src/components/expansion-item/QExpansionItem.sass
+++ b/ui/src/components/expansion-item/QExpansionItem.sass
@@ -1,8 +1,16 @@
 .q-expansion-item
 
+  .q-item__section--avatar:has(> &__toggle-icon)
+    color: $grey-7
+
+  .q-item--dark .q-item__section--avatar:has(> &__toggle-icon)
+    color: rgba(255,255,255,.7)
+
+  .q-item__section--side:has(> .q-icon):not(:has(> &__toggle-icon))
+    color: inherit !important
+
   &__border
     opacity: 0
-
 
   &__toggle-icon
     position: relative


rstoenescu:
Thanks for digging into this! Your diagnosis was spot on: `switch-toggle-side` swaps the two sections' side/avatar roles, and QItem's colors ride along with those classes.

We ended up fixing it at the component level instead, in f205ed3f1e2986d76f8d5030ae888c62e1b07e0a: the toggle section now always stays a plain side section and the `icon` section always stays an avatar section, so QItem's existing light/dark colors apply unchanged on either side, and only the geometry of the swapped positions (the avatar 56px min-width) is compensated in QExpansionItem.sass.

The reason we couldn't merge this PR as-is: because of how Sass substitutes `&` inside a nested selector, the third rule compiles without any `.q-expansion-item` scoping, as `.q-item__section--side:has(> .q-icon):not(:has(> .q-expansion-item__toggle-icon)) { color: inherit !important }`. That would repaint every QItem side icon section in every app, and the `!important` would make it hard to override. It also covered `.q-item--dark` but not a dark QList ancestor. The role-preserving fix adds no color rules at all, so app-level overrides keep working exactly as before.

Closing in favor of that commit; it will ship with the next Quasar UI release. Thanks again for the report-quality PR!

Issue #17155, opened 2024-05-01
QInput `shadow-text` is showing when input is out of focus but `hasError` is true


Dtsiantaris:
### What happened?

When there is an `error` present and the input loses focus the shadow text is still displayed. This creates an ugly overlapping effect of `label` and `shadow-text`.

If there **IS NOT AN ERROR** the shadow text disappears when focus is lost,

### What did you expect to happen?

I have a `q-input` where i use `rules` for some basic validation. I also have a `shadow-text` that i want to always be visible when the input is `focused`.

**EDIT**
After playing around, i found out that the class `q-field--highlighted` is responsible for the visibility of the `shadow-text`. This class is normally removed when focus is lost but NOT when there is an error.

### Reproduction URL

https://codepen.io/dtsiantaris/pen/JjVqLZJ

### How to reproduce?

1. Go to the provided link
2. Type something -> You can see the shadow text
3. Erase it so the `val.length` rule kicks in -> You can still see the shadow-text
4. Click outside the q-input -> You can see still see the shadow-text

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Components (quasar)

rstoenescu:
Fixed in dev by 91f84afeb: shadow-text visibility now follows focus alone instead of the highlighted state, so an invalid field hides it on blur and still shows it while focused. Will ship with the next Quasar UI version.

PR #18164, opened 2025-10-30
fix(ui): QInput shadow-text is showing when input is out of focus but hasError (#17155)

IceIsFrozen:

- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

Since errors were treated the same as focus for the highlighted class in use-field.js
```
+ (hasError.value === true || state.focused.value === true ? ' q-field--highlighted' : '')
```
made a simple change for the shadow-text/focused situation by splitting off for focused and checking if there's an error. If not, shadow-text acts normally.

### DIFF
--- a/ui/src/components/field/QField.sass
+++ b/ui/src/components/field/QField.sass
@@ -209,6 +209,8 @@ $field-transition-label-right-up: .324s cubic-bezier(.4,0,.2,1)
   &--highlighted
     .q-field__label
       color: currentColor
+
+  &--focused:not(.q-field--error)
     .q-field__shadow
       opacity: .5
 


rstoenescu:
Thank you for the diagnosis, it was spot on: the shadow-text opacity was gated on `q-field--highlighted`, which stays applied while the field has an error. A slightly simplified form of this fix landed in dev as 91f84afeb: the opacity rule moved to `q-field--focused` alone, without the `:not(.q-field--error)` guard, so the shadow text still shows while typing in an invalid field (per the expectation in #17155) and the selector keeps low specificity for userland overrides. Also added a cross-engine regression test covering this. Will be part of the next Quasar version. Closing in favor of the dev commit; credit to you for pinpointing the cause.

Issue #18170, opened 2025-11-07
Missing label for QTable pagination records per page


hsimpson270:
### What happened?

The "Records per page" dropdown that appears when using pagination in `QTable` is missing a descriptive label. This causes an accessibility issue as all controls should have an associated label to describe the control.

### What did you expect to happen?

The dropdown should have an associated label. This could be achieved by either setting `aria-describedby` to reference the "Records per page:" element or by providing `aria-label` to the dropdown.

### Reproduction URL

https://jsfiddle.net/hjp3c82s/1/

### How to reproduce?

1. Go to the provided reproduction link
2. Navigate to the bottom of the QTable
3. Right click the "Records per page" dropdown
4. Click Inspect
5. Observe the element has no associated label

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Accessibility [a11y] (quasar)

### Platforms/Browsers

Chrome

### Quasar info output

```Shell

```

### Relevant log output

```Shell

```

rstoenescu:
The rows-per-page QSelect now carries an aria-label matching the visible label (rows-per-page-label prop or the lang pack string) — commit eac658a66. Fix will be available in v2.25.0

PR #18180, opened 2025-11-20
fix(a11y): add aria-label to QTable records-per-page QSelect (#18170)

vamsi2246:
This PR fixes issue **#18170**, which reported that the QTable pagination "Records per page" dropdown lacked an accessibility label.

The `QSelect` used for selecting the number of rows per page was missing an `aria-label`, causing screen readers to announce it incorrectly and failing WCAG accessibility requirements.

### **What was changed**

Added the following property to the QSelect inside `getPaginationDiv()` in `ui/src/components/table/QTable.js`:

```js
'aria-label': props.rowsPerPageLabel || $q.lang.table.recordsPerPage
```

This ensures:

* If the user provides `rowsPerPageLabel`, it becomes the aria-label.
* Otherwise it falls back to the default language string (`$q.lang.table.recordsPerPage`).

### **Impact**

* Improves accessibility for screen-reader users.
* Fixes the missing label reported in #18170.
* Fully backward-compatible.
* No behavioral or visual changes to QTable.

### **Testing**

Tested locally in the Quasar dev environment and confirmed that the rendered HTML now includes:

```html
<select aria-label="Records per page">
```

This resolves the accessibility issue.

### DIFF
--- a/ui/src/components/table/QTable.js
+++ b/ui/src/components/table/QTable.js
@@ -838,6 +838,7 @@ export default createComponent({
             dense: true,
             optionsDense: true,
             optionsCover: true,
+            'aria-label': props.rowsPerPageLabel || $q.lang.table.recordsPerPage,
             'onUpdate:modelValue': onPagSelection
           })
         ])


rstoenescu:

We're clearing out the PR backlog. This one is already handled on `dev`: the `aria-label` is already applied to the records-per-page QSelect in `ui/src/components/table/QTable.js`.

Closing for that reason. If you think something is still missing, please open a fresh PR against current `dev` and we'll look at it promptly.

Issue #5115, opened 2019-09-20
Q-Table support to columns drag resizing


dariodepaolis:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
I'd like to have a prop to enable the manual resize of column by user. (Like Grid component in ExtJS or Syncfusion)
Also a dblclick on the resizer the will expand to maximum width of the col will be awesome.

**Describe alternatives you've considered**
No alternative so far found, we have to manually rewrite the table header adding dragging function, auto-width, expand, ellipsis etc.

rfox12:
+1.  If Quasar is going to get serious about an improved data table (which I would love) I think perhaps we should have a holistic discussion.  Potentially huge amounts of work.

antonispat10:
Hello,

You could check my pull request https://github.com/quasarframework/quasar/pull/5460

dariodepaolis:
Great ty, asap I'll give it a try!

rfox12:
This is more important now.  User's expect columns to be able to be resized!  Also need a nice / easy solution for truncating text.  Some other thoughts...

- set a minimum column size (entire table or per column).  We don't want the user to make the column completely disappear unintentionally.  

- Have treatment options for text that is too long to fit in the cell (wrap text, use ellipses, or some "other" slot) AND be able to have variables that we can use to provide other UI actions (e.g. maybe I want to display a tooltip--but only if the text in that particular cell had to be truncated)

- Sticky header and footer options without complicated CSS--to stay in effect if the QTable goes "fullscreen".  I think having some caveats about browser support in the documentation would be ok...

- Stretch goal... drag / reorder columns / rows

satyavh:
We would also love this feature 👍

bytebrain:
I had same problem, now I use https://quasar.dev/vue-components/splitter in stead of tables.

Jokertwo:
I would like to ask about the status of this feature. Is there any progress?

Jokertwo:
> I would like to ask about the status of this feature. Is there any progress?

Or exist there some workaround on how to solve it?

giftofsky:
Any update/solution on the issues? How can we raise the priority of this?

Jokertwo:
I found one possible workaround for how to solve this issue. [Example](https://codepen.io/crwilson311/pen/Bajbdwd?editors=1010)

hawkeye64:
[...]

PR #18184, opened 2025-11-23
Implementing Q-Table support to columns drag resizing (#5115)

francoo98:
This works on #5115 (there is likely work left to do)
1. added a new composable called `table-column-resizer.js` for `QTable.js` that allows resizing of columns widths.
2. created an example in the playground.
3. updated the docs
The issue asking for this features has a lot of positive reactions, so I think it is valuable for a lot of users. Furthermore, there is a bounty for it.

- [ ] Bugfix
- [X] Feature
- [X] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [X] No

- [X] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs)  or explained in the PR's description.

- [X] A convincing reason for adding this feature (to avoid wasting your time, it's best to [start a new feature discussion](https://github.com/quasarframework/quasar/discussions/new?category=ideas-proposals) first and wait for approval before working on it)

**Other information:**
/claim #5115


rstoenescu:
Hi @francoo98, thank you for putting real work into this: composable, docs and a playground page included. Column resizing (#5115) is a feature people genuinely want, so we did give this a full read.

We're closing this PR rather than iterating on it, because the implementation is too far from what QTable core needs:

* Resizing is mouse-only (`mousedown` on the handle). Core QTable features must also work with touch/pen (pointer events or Quasar's TouchPan), handle RTL, and offer an accessible path.
* The header row is keyed on `JSON.stringify(columnWidths)`, which re-creates every header cell on each mousemove frame while dragging. That's both a performance problem and a sign the width flow needs a different design.
* The `column-widths` prop is only read once at setup, so it can't actually be controlled by the parent afterwards (the "Reset all widths" button in the PR's own playground page has no effect).
* The width merging assumes string `headerStyle`/`style`, but these can be functions/objects; and there's no handling or tests for virtual scroll, sticky columns, grid mode or custom `header-cell` slots.

For a feature of this size the design needs to be agreed on first (see the contributing guide's note on discussing new features before implementing). #5115 remains open; this close is about this specific implementation, not about the feature itself.

Thanks again for contributing.

Issue #18185, opened 2025-11-23
QDialog: scroll position not restored if route changed


w0pp:
### What happened?

QDialog doesn't restore scroll position after closing it if `window.location.href` has changed.

https://github.com/user-attachments/assets/7d3384f3-1ece-4cf1-8467-de1e2e4c4025

### What did you expect to happen?

QDialog should restore scroll position after closing even if `window.location.href` has changed. Otherwise, you can't change URL hash or query parameters and have the scroll position restored after closing the dialog.

Possible alternative solutions:
- Restore scroll position if only URL hash or query parameters have changed, but route path stayed same.
- Add a setting to control this.
- Decide based on QDialog's `no-route-dismiss` prop.
- Decide based on Vue Router's `scrollBehavior` option.

### Reproduction URL

https://stackblitz.com/edit/quasarframework-stackblitz-templates-yhk2dumd?file=src%2Fpages%2FIndexPage.vue

### How to reproduce?

1. Go to the provided reproduction link
2. Scroll down
3. Click "Show dialog"
4. Close dialog by clicking outside it
5. The page is scrolled up - scroll position isn't restored because URL has changed

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Components (quasar)

### Platforms/Browsers

Chrome

### Quasar info output

```Shell
Operating System - Linux(5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36) - linux/x64
NodeJs - 20.19.1

Global packages
  NPM - 10.8.2
  yarn - 1.22.19
  pnpm - 8.15.6
  bun - Not installed
  @quasar/cli - undefined
  @quasar/icongenie - Not installed
  cordova - Not installed

Important local packages
  quasar - 2.17.4 -- Build high-performance VueJS user interfaces (SPA, PWA, SSR, Mobile and Desktop) in record time
  @quasar/app-vite - 2.4.0 -- Quasar Framework App CLI with Vite
  @quasar/extras - 1.17.0 -- Quasar Framework fonts, icons and animations
  eslint-plugin-quasar - Not installed
  vue - 3.5.24 -- The progressive JavaScript framework for building modern web UI.
  vue-router - 4.6.3
  pinia - Not installed
  vite - 7.2.4 -- Native-ESM powered web dev build tool
  vite-plugin-checker - Not installed
[...]

PR #18190, opened 2025-12-02
fix(QDialog): restore scroll after close if only hash/query changed #…

devel-maverick:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #18185`)
- [ ] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/quasar/tree/dev/docs) or explained in the PR's description.

**Description / What problem is solved:**
QDialog does not restore scroll position if the URL hash or query parameters change while the dialog is open. 
Currently, scroll is restored only if `window.location.href === href`, which fails when only hash/query changes.

This change updates the check to compare only the **route path** (`window.location.pathname`) instead of the full URL. 
Scroll is now restored correctly after closing the dialog if the route path has not changed.

**Code change:**

```js
// before:
if (window.location.href === href) {
  window.scrollTo(scrollPositionX, scrollPositionY)
}

// after:
if (window.location.pathname === new URL(href).pathname) {
  window.scrollTo(scrollPositionX, scrollPositionY)
}
`

### DIFF
--- a/ui/src/utils/scroll/prevent-scroll.js
+++ b/ui/src/utils/scroll/prevent-scroll.js
@@ -149,9 +149,10 @@ function apply (action) {
     body.style.top = bodyTop
 
     // scroll back only if route has not changed
-    if (window.location.href === href) {
-      window.scrollTo(scrollPositionX, scrollPositionY)
-    }
+  if (window.location.pathname === new URL(href).pathname) {
+    window.scrollTo(scrollPositionX, scrollPositionY)
+  }
+
 
     maxScrollTop = void 0
   }

### CODE REVIEW

**Comment by rstoenescu** [comment]

This one is already handled on `dev`. The same fix landed in #18406 (commit 241325677), which compares the route path instead of the full URL, capturing `window.location.pathname` when the dialog opens rather than re-parsing the stored `href` on close:

```js
// ui/src/utils/scroll/prevent-scroll.js
routePath = window.location.pathname
// ...
// scroll back only if route path has not changed
if (window.location.pathname === routePath) {
  window.scrollTo(scrollPositionX, scrollPositionY)
}
```

Closing for that reason, and apologies that it wasn't closed back when #18185 was. If you're still seeing the scroll position lost on a hash or query change with the latest release, please open an issue with a repro and we'll take another look.

Issue #18183, opened 2025-11-23
QDialog: Scrolled sticky elements disappear in Firefox after opening Dialog


w0pp:
### What happened?

The sticky element disappears, and re-appears after closing the dialog

https://github.com/user-attachments/assets/f00fd762-6bd9-4728-a8c1-3c0082535f7b

### What did you expect to happen?

The sticky element is still displayed in the background (like it does in Chrome)

https://github.com/user-attachments/assets/5a2a290d-8277-4134-98e8-382a12a2228e

### Reproduction URL

https://stackblitz.com/edit/quasarframework-stackblitz-templates-u77zfjcv?file=src%2Flayouts%2FMainLayout.vue

### How to reproduce?

Steps to reproduce:

1. Place a sticky element in the application
2. Make application scrollable
3. Open the application in Firefox
4. Scroll to the bottom
5. Open a dialog

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Components (quasar)

### Platforms/Browsers

Firefox

### Quasar info output

```Shell
Operating System - Linux(5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36) - linux/x64
NodeJs - 20.19.1

Global packages
  NPM - 10.8.2
  yarn - 1.22.19
  pnpm - 8.15.6
  bun - Not installed
  @quasar/cli - undefined
  @quasar/icongenie - Not installed
  cordova - Not installed

Important local packages
  quasar - 2.17.4 -- Build high-performance VueJS user interfaces (SPA, PWA, SSR, Mobile and Desktop) in record time
  @quasar/app-vite - 2.4.0 -- Quasar Framework App CLI with Vite
  @quasar/extras - 1.17.0 -- Quasar Framework fonts, icons and animations
  eslint-plugin-quasar - Not installed
  vue - 3.5.24 -- The progressive JavaScript framework for building modern web UI.
  vue-router - 4.6.3
  pinia - Not installed
  vite - 7.2.4 -- Native-ESM powered web dev build tool
  vite-plugin-checker - Not installed
  eslint - 9.39.1 -- An AST-based pattern checker for JavaScript.
  esbuild - 0.25.12 -- The cross-platform WebAssembly binary for esbuild, a JavaScript bundler.
  typescript - Not installed
  workbox-build - Not installed
  register-service-worker - Not installed
[...]

PR #18191, opened 2025-12-02
fix(QDialog): preserve sticky elements in Firefox (#18183)

devel-maverick:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #18183, #18185`)

**Other information:**

This PR fixes two QDialog issues:  

1. **#18183** – Sticky elements disappearing in Firefox when opening QDialog.  
   - Solution: Removed `position: fixed` and `left/top` adjustments; now using `overflow: hidden` and `padding-right` to preserve layout and scrollbars.

2. **#18185** – Scroll position not restored if URL changed.  
   - Solution: Scroll restoration now considers only the pathname. If pathname unchanged, scroll is restored even if query/hash changed.

Tested on Firefox, Chrome, and Safari. iOS-specific behavior retained.

### DIFF
--- a/ui/src/utils/scroll/prevent-scroll.js
+++ b/ui/src/utils/scroll/prevent-scroll.js
@@ -8,8 +8,6 @@ let
   scrollPositionY,
   maxScrollTop,
   vpPendingUpdate = false,
-  bodyLeft,
-  bodyTop,
   href,
   closeTimer = null
 
@@ -31,7 +29,7 @@ function shouldPreventScroll (e) {
     delta = shift || scrollY ? e.deltaY : e.deltaX
 
   for (let index = 0; index < path.length; index++) {
-    const el = path[ index ]
+    const el = path[index]
 
     if (hasScrollbar(el, scrollY)) {
       return scrollY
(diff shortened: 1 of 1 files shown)


rstoenescu:

Closing this one, but the Firefox bug it targets is not being dismissed: **#18183 stays open** as the tracker for it.

The reason is that `prevent-scroll.js` has moved twice since this was written, and the PR can no longer apply on top of either change:

1. The **#18185** half is now redundant. That fix landed in #18406 (commit 241325677) using the same route-path comparison.
2. The **#18183** half removes `position: fixed` and the `bodyLeft`/`bodyTop` handling from the JS, but commit 230b171e6 has since moved that behavior into CSS. `.q-document--prevent-scroll` now applies `body { position: fixed !important }` in `ui/src/css/core/helpers.sass`, so removing it from the JS alone would no longer have any effect.

There are also a few things that would need cleaning up regardless: the `// ===== PATCH START =====` markers, the removal of the explanatory comment in `onAppleScroll` (it documents why that no-op assignment is required on iOS), and the unrelated whitespace reformatting.

If you'd like to take another run at #18183 against current `dev`, that would be very welcome, and a fresh PR scoped to just that bug would be much easier for us to review.

PR #18200, opened 2025-12-17
fix(QInput): change q-input about type text clearValue() default null to empty string(#14032)

### LINKED ISSUE #14032: Using clearable in masked input set model to empty string instead of null
### What happened?

When using mask property and clearable on q-input, once input is cleared, the model value is `""` instead of `null`

The input should be cleared as `null`

### Reproduction URL

https://codepen.io/lucianobosco/pen/eYMRyOe

### How to reproduce?
```
<q-input
    v-model="date"
    clearable
    mask="##/##/####"
    label="Date" /> 
```

### Flavour

### Areas

Components (quasar)

tsukuha:
close: https://github.com/quasarframework/quasar/issues/14032

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

QInput clearable func is changed its default null to empty string when input type is 'text', 'search', 'url', 'tel', 'password', or textarea.

**Other information:**

### DIFF
--- a/ui/src/composables/private.use-field/use-field.js
+++ b/ui/src/composables/private.use-field/use-field.js
@@ -263,6 +263,13 @@ export default function (state) {
     return acc
   })
 
+  const isTextarea = computed(() => props.type === 'textarea')
+
+  const isTypeText = computed(() =>
+    isTextarea.value === true
+    || [ 'text', 'search', 'url', 'tel', 'password' ].includes(props.type)
+  )
+
   function focusHandler () {
     const el = document.activeElement
     let target = state.targetRef?.value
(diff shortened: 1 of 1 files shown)


rstoenescu:
Thank you for taking the time to work on this and for the thorough testing checklist. After careful consideration we've decided not to merge it, for a few reasons:

1. Emitting `null` on clear is a documented, long-standing part of the `clearable` contract ("When clicked, model becomes null"). It's deliberate: `null` means "no value", which is distinguishable from an empty string that the user actually typed. Many apps rely on `=== null` checks in validation rules, watchers and submit logic, so this would be a breaking change for them (as your own checklist notes).

2. Making the emitted value depend on the `type` prop would split that contract: the same `@update:model-value` handler would receive `''` for text-like inputs but `null` for everything else. Consistency across all field components is worth more here than matching the native empty-string convention for a subset of types.

3. The referenced issue (#14032) actually asked for the opposite direction: a masked input was emitting `''` on clear and the reporter expected `null`. That issue has since been closed, so this change wouldn't resolve it as filed.

For apps that prefer an empty string, it's a one-liner without any core change:

```html
<q-input v-model="text" clearable @clear="text = ''" />
```

Thanks again for contributing, and we hope to see more PRs from you.

PR #18219, opened 2026-01-22
feat(create-quasar): use pnpm-workspace.yaml instead of npmrc

yusufkandemir:

- Feature

- Kind of - it only works with pnpm v10+, but it's easy to adapt, especially with Node Corepack

**Other information:**
pnpm v10 uses `pnpm-workspace.yaml` as the recommended way to configure pnpm options, instead of `.npmrc`. See https://pnpm.io/settings; they don't explicitly say it's recommended, but the title, contents, etc. imply so.
The options except shamefully-hoist now have the defaults we want, so we drop them as well.

As per pnpm's [security policy](https://github.com/pnpm/pnpm/security/policy), v9 won't be supported after April 30th, 2026.

Relevant PR #18218 - it improves handling of pnpm-workspace.yaml for Capacitor


IlCallo:
From what I can see, this isn't a breaking change as it only applies to new projects
I'm still wrapping my head about the override thing, seems like we're disabling `prompts.override` entirely, which may not be what we want in all cases

yusufkandemir:
Yes, but if someone is using pnpm v9 and can't use v10 for some reason, they won't be able to directly use the projects they will create from now on. If they use Corepack, this won't be an issue. It's also easy to adopt it back, so it's mostly a small breaking change that could cause some potential annoyance, not more than that.

yusufkandemir:
We are only using `prompts.override()` in the code for the project name override. In the context of the create test project script, disabling the override is fine. If we need more kinds of overrides that are legit in the future, we can rethink. I honestly think we should find a better alternative to `prompts`, but that is a lower priority.

rstoenescu:
Thanks!

Issue #18289, opened 2026-04-12
QEditor: clicking away from linkEditor leaves a broken link


robgwin:
### What happened?

In q-editor, select some text, click the link button, paste in a URL. Instead of hitting enter or "update", click back into the main editor to continue writing. It looks like you've successfully added the link, so you carry on, unaware that it saved as "https://" because you didn't apply it.

### What did you expect to happen?

Either:

A. If I haven't actually applied the URL, the text should not appear as linked, or

B. The link should be applied when I click away from the link editor

I think option B probably makes more sense, because it also handles the scenario of re-editing an existing link, and I think it matches what a typical user would expect (No other editor content requires an explicit "update" action.)

It seems like this could be accomplished by adding an onBlur() callback after the onKeydown callback in editor-utils.js:getLinkEditor()
```
        onBlur: () => {
          updateLink()
        },
```

### Reproduction URL

https://stackblitz.com/edit/quasarframework-stackblitz-templates-fysxxjxb?file=src%2Fpages%2FIndexPage.vue

### How to reproduce?

(Note you can also reproduce this directly on https://quasar.dev/vue-components/editor)

1. Select any unlinked word and click the link tool
2. Paste in a URL (do not hit enter or update)
3. Click in the main editor window
4. Notice the word appears linked
5. Select the word again and click the link tool
6. Notice the link is actually 'https://'

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Components (quasar)

### Platforms/Browsers

Chrome

### Quasar info output

```Shell

```

### Relevant log output

```Shell

```

robgwin:
The issue can be reproduced directly on the quasar.dev demo page, but I've added a StackBlitz now to satisfy the requirement.

rstoenescu:
Fix lands in v2.25

PR #18290, opened 2026-04-13
fix(QEditor): add validation for q-editor link (#18289)

Abdulqudus001:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
This fixes the link editing behavior highlighted in #18289  in `QEditor` so opening the link toolbar no longer auto-applies link formatting to the current selection before the user confirms a value.

**Changes made**

- Removed eager `createLink` call when entering link edit mode in `ui/src/components/editor/editor-caret.js`.
- Updated link confirmation logic in `ui/src/components/editor/editor-utils.js` to:
  - trim input before applying,
  - treat empty value or unchanged placeholder (`https://`) as a no-op,
  - run `createLink` only when the URL actually changes.

**Behavior after this fix**

- Clicking the link button opens the URL editor without modifying selected text.
- Link is applied only on explicit user action with a valid changed URL.

### DIFF
--- a/ui/src/components/editor/editor-caret.js
+++ b/ui/src/components/editor/editor-caret.js
@@ -308,7 +308,6 @@ export default class Caret {
 
         this.eVm.editLinkUrl.value = urlRegex.test(url) ? url : 'https://'
         this.save(selection.getRangeAt(0))
-        document.execCommand('createLink', false, this.eVm.editLinkUrl.value)
       } else {
         this.eVm.editLinkUrl.value = link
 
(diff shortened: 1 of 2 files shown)


rstoenescu:
Hi!

The diagnosis was right, the eager createLink was the culprit; what it missed is that committing on blur unmounts the whole toolbar row before a click on Remove/Update can land, which killed the Remove button, and that dropping the eager call also removed the only thing that linkified a selection already spelling out a URL. Three existing tests were catching both. Fixed on top with mousedown-prevent on those buttons, the commit decision taken against the selection's real href, a placeholder guard, and ESC no longer unlinking.

PR #18300, opened 2026-05-13
fix: let item labels inherit container color

prantikmedhi:
## Summary\n- make QItem overline and caption labels inherit the item color instead of using fixed gray values\n- preserve the muted look via opacity so active/custom text colors carry through\n\n## Verification\n- git diff --check

### DIFF
--- a/ui/src/components/item/QItem.sass
+++ b/ui/src/components/item/QItem.sass
@@ -44,10 +44,12 @@
     max-width: 100%
 
     &--overline
-      color: rgba(0,0,0,.7)
+      color: inherit
+      opacity: .7
 
     &--caption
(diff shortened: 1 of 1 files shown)


rstoenescu:
Thank you for this! Your fix has landed in dev as 3b6b74816 (authored to you), adapted during review: we used color-mix(in srgb, currentColor N%, transparent) instead of element opacity so nested colored content (badges, links) keeps full strength, kept the dark caption at its current 80% emphasis, and kept the old fixed colors as fallback declarations for browsers without color-mix support. A follow-up commit (0e2c62761) extends the same treatment to header labels and side sections. Will be available in the next Quasar UI release.

PR #18307, opened 2026-05-24
docs: add Electron live updates guide

riderx:
## What changed

Adds an Electron Live Updates page under the Quasar CLI Vite Electron docs and links it from the Electron mode menu.

The page covers:

- what Electron live updates can and cannot update
- installing the open-source `@capgo/electron-updater` package in `src-electron`
- wiring the updater in the Electron main process
- exposing a limited updater API through the preload bridge
- calling `notifyAppReady()` from the renderer
- using a manual self-hosted metadata endpoint and bundle storage
- creating an update zip with the CLI
- keeping signed desktop releases for main/preload/native/binary changes

## Why

Quasar Electron users currently have build and publishing docs, but no focused guide for renderer-only live updates. This documents the self-hosted/manual pattern without requiring a hosted account.

## Validation

- `pnpm lint:check`

Note: local commands emitted workspace Node engine warnings because this machine is running Node v25.9.0, while several Quasar packages request Node 22/24/26+ ranges. The final lint check passed.

### DIFF
--- a/docs/src/assets/menu.js
+++ b/docs/src/assets/menu.js
@@ -1063,6 +1063,10 @@ export default [
             path: 'build-commands'
           },
           {
+            name: 'Live Updates',
+            path: 'live-updates'
+          },
+          {
             name: 'Preload Script',
             path: 'electron-preload-script'
           },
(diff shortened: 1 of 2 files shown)


rstoenescu:
Thanks for contributing!
And awesome that the page is already oriented towards app-vite v3 💯

PR #18323, opened 2026-06-14
fix(QInput,QSelect,QField): expose error state to assistive tech

### LINKED ISSUE #17306: a11y: handle ARIA error state in form field components
### What happened?

Our app needs to comply with WCAG guideline level AA. When auditing our app, our accessibility agency reported a severe a11y issue, where screenreaders do not announce an input field as being invalid when focusing.

Take this screenshot from the reproduction URL below with the NVDA speech viewer window open. Tabbing through the elements [random button, input, reset button, input]:
- on first focus (input field is valid, initial state) NVDA reads "Required Field edit blank"
- tabbing one step further to reset button and then Shift-Tab'ing one step back to input
- on second focus (input field now invalid, required and empty) NVDA still reads "Required Field edit blank", no hint it is in invalid state

focusing an invalid qinput

When focusing on an already invalid field, its invalidity should be announced immediately, so that a screenreader includes the keyword "invalid", ideally including any validation messages, e.g. "Address is a required".

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-invalid) `aria-invalid="true"` on the native `<input>` should only be added upon validation. An accompanying `aria-errormessage="some_id"`, pointing to the message container (at the bottom), e.g. `q-field__messages` should further help to identify the error.

I have not yet come to a conclusion, whether `aria-describedBy="some_id"` or `aria-errormessage="some_id"` would be the better choice here.

### Reproduction URL

https://codepen.io/hennzen/pen/LYodzgg

### How to reproduce?

1. Install [NVDA](https://www.nvaccess.org/download/) or any other screenreader of your choice
2. Open NVDA's speech viewer (or listen to audio)
3. Go to the reproduction URL
4. Tab through the elements (without typing anything in the input field) and watch screenreader's output when focusing the input field for the first time (valid) and second time (invalid)
...alternatively use the mouse to focus, blur and then focus again on the input field

### Flavour

### Areas

Components (quasar), Composables (quasar), Accessibility [a11y] (quasar)

### Platforms/Browsers

Firefox, Chrome, Safari, Microsoft Edge

### Quasar info output

_No response_

### Relevant log output

_No response_

### Additional context

Created a patch (v2.16.4) to solve this in a works-for-me-quality and will attach it in the comments.

web3blind:
Fixes #17306.

This adds the missing ARIA error-state relationship for Quasar fields so assistive technologies can identify invalid controls and reach the rendered error message.

### What changed

- Adds a stable id to the rendered field error message container while an error is active.
- Applies `aria-invalid="true"`, `aria-errormessage`, and `aria-describedby` to native `QInput` controls in error state.
- Applies the same relationship to both QSelect control paths (`use-input` and the focus-target-only combobox), which completes the QInput and QSelect scope reported in #17306.
- Preserves existing `aria-describedby` values instead of replacing them.
- Preserves an explicit `aria-errormessage`; because ARIA 1.2 defines it as a single ID reference rather than an ID-reference list, Quasar does not concatenate another value.
- Exposes `ariaInvalid`, `ariaDescribedby`, and `ariaErrormessage` in the `QField` control slot scope so custom controls can wire the same relationship.
- Documents the new control slot scope values in the QField API metadata.
- Centralizes native-control error ARIA composition in the field layer and adds focused QInput, QSelect, and QField tests.

### Validation

- `pnpm --filter quasar build`
- `pnpm --filter quasar-ui-test exec vitest run ../src/components/input/QInput.test.js ../src/components/field/QField.test.js ../src/components/select/QSelect.test.js --reporter=verbose`
  - 3 test files passed
  - 7 focused tests passed
- `pnpm exec oxlint ui/src/composables/private.use-field/use-field.js ui/src/components/input/QInput.js ui/src/components/input/QInput.test.js ui/src/components/select/QSelect.js ui/src/components/select/QSelect.test.js`
- `git diff --check`

The original contributor also ran the complete UI test suite successfully (101 files and 1,063 tests). A full `pnpm lint:ui:check` in that checkout was blocked by an unrelated missing generated `ui/playground/.quasar/tsconfig.json` file.

## Summary by CodeRabbit

* **Accessibility**
  * Improved field, input, and select accessibility by associating validation errors with controls through ARIA attributes.
  * Added invalid-state and error-message references to rendered form controls.
  * Exposed error-related ARIA attributes through the field control slot.
* **Bug Fixes**
  * Ensured validation error messages are correctly linked to their corresponding inputs and selects.
* **Tests**
  * Added coverage verifying error-related ARIA attributes in the field control slot.


hawkeye64:
@rstoenescu Added a maintainer follow-up in 36744c042 to complete the issue's QSelect scope and preserve explicit ARIA references. The shared field helper now drives QInput and both QSelect target-control paths. Focused tests (7), Oxlint, the full Quasar UI build, and git diff checks pass.

yusufkandemir:
@coderabbitai review

MILLERMARRU (commented):
Checked `getErrorAriaAttrs` against the current WAI-ARIA/MDN guidance for `aria-errormessage` and this is doing the right thing on both counts that usually get missed. First, it only emits `aria-invalid`/`aria-errormessage`/`aria-describedby` at all when `hasError.value === true`, MDN is explicit that `aria-errormessage` is meaningless without `aria-invalid="true"` and shouldn't be exposed on a valid field. Second, it sets `aria-describedby` alongside `aria-errormessage` rather than relying on `aria-errormessage` alone, which matches the "defense in depth" recommendation since `aria-errormessage` support across screen readers is still inconsistent, `aria-describedby` is the one that's been reliably supported for years.

The merge behavior for pre-existing `aria-describedby`/`aria-errormessage` on the control is also right, appends rather than overwrites, and the QInput test for "preserves existing aria-describedby values" confirms the append order (external value first, then the message id). Good that this landed on QInput, QSelect and QField's custom-control slot in the same PR instead of just the native-input case, that's usually where these fixes get left half done, and the QSelect diff shows it's wired for both the `useInput` and non-`useInput` render paths.

PR #18325, opened 2026-06-14
fix(app-vite): iOS --ide opens .xcworkspace not .xcodeproj (fix: #18324)

### LINKED ISSUE #18324: quasar build -m capacitor -T ios --ide opens App.xcodeproj instead of App.xcworkspace ("No such module 'Capacitor'")
### What happens

`quasar build -m capacitor -T ios --ide` opens **App.xcodeproj** in Xcode. Building from there fails with:

```
error: No such module 'Capacitor'
```

The CocoaPods that provide the `Capacitor` module are only linked into **App.xcworkspace**, so the bare project can't resolve them.

### Root cause

`@quasar/app-vite` → `lib/utils/open-ide.js`, `findXcodeWorkspace()` returns the first directory entry matching *either* extension:

```js
if (item.endsWith('.xcworkspace') || item.endsWith('.xcodeproj')) {
  return __path
}
```

`src-capacitor/ios/App/` always contains both `App.xcodeproj` and `App.xcworkspace`, and `fs.readdirSync` lists `App.xcodeproj` first (alphabetical), so `--ide` opens the project and never the workspace.

Non-`--ide` builds are unaffected — `capacitor-builder.js` runs `xcodebuild -workspace App.xcworkspace` explicitly, which is why only the `--ide` path breaks.

### Reproduce

1. `quasar mode add capacitor`, add the iOS platform, `npx cap sync ios` (Pods installed)
2. `quasar build -m capacitor -T ios --ide`
3. Xcode opens `App.xcodeproj` → Build → `No such module 'Capacitor'`
4. Closing it and opening `App.xcworkspace` manually builds fine

### Expected

`--ide` opens `App.xcworkspace` when it exists (matching what `npx cap open ios` and the non-`--ide` build already do), falling back to `.xcodeproj` only if there's no workspace.

### Fix

Prefer `.xcworkspace`, fall back to `.xcodeproj`. PR follows.

### Environment

- `@quasar/app-vite` 2.6.2 (same code on `dev`)
- Capacitor 8, host macOS, Xcode opening via `--ide`

pich:
Closes #18324.

- [x] Bugfix

- [x] No

### Problem

`quasar build -m capacitor -T ios --ide` opened **App.xcodeproj**, and Xcode builds from there failed with `No such module 'Capacitor'` — the Pods are only linked into **App.xcworkspace**.

`findXcodeWorkspace()` returned whichever of the two `readdirSync` listed first, and `App.xcodeproj` sorts before `App.xcworkspace`, so the workspace was never opened. The non-`--ide` build was unaffected because `capacitor-builder.js` passes `-workspace App.xcworkspace` explicitly.

### Fix

Prefer the `.xcworkspace`, fall back to `.xcodeproj` only when no workspace exists. This matches what `npx cap open ios` and the headless build already do. Behaviour is unchanged for Cordova/older layouts that ship a single target.

- [x] When resolving a specific issue, it's referenced in the PR's title (`fix: #18324`)

### Testing

Verified on a Capacitor 8 iOS app on macOS: before the change `--ide` opened `App.xcodeproj` (build → `No such module 'Capacitor'`); after it, `App.xcworkspace` opens and builds cleanly. Cordova shares the same code path; Electron is unaffected.

### DIFF
--- a/app-vite/lib/utils/open-ide.js
+++ b/app-vite/lib/utils/open-ide.js
@@ -6,15 +6,17 @@ import open from 'open'
 import { fatal, warn } from './logger.js'
 
 function findXcodeWorkspace(folder) {
-  const root = fs.readdirSync(folder)
+  const items = fs.readdirSync(folder)
 
-  for (const item of root) {
-    const __path = path.join(folder, item)
+  // Prefer the .xcworkspace: it's the CocoaPods-integrated container that
+  // links the Pods providing the Capacitor module. readdirSync() can list
+  // the bare .xcodeproj first, and opening that builds without the Pods,
+  // failing with "No such module 'Capacitor'".
+  const target =
+    items.find(item => item.endsWith('.xcworkspace'))
+    || items.find(item => item.endsWith('.xcodeproj'))
 
-    if (item.endsWith('.xcworkspace') || item.endsWith('.xcodeproj')) {
-      return __path
-    }
-  }
+  return target !== undefined ? path.join(folder, target) : undefined
 }
 
 function runMacOS(mode, target, appPaths) {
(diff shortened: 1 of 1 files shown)


rstoenescu:
Thanks for contributing!
Releasing rc.3 with it right now.

PR #18342, opened 2026-07-02
feat(ui): audit component API JSON drift

hawkeye64:
## What changed

Adds a `pnpm --dir ui run audit:api` script that compares hand-maintained component API JSON against runtime component definitions and source-level API evidence.

The audit currently checks:

- component API files import cleanly with Quasar SSR globals in place
- documented props match runtime prop names, types, required flags, and defaults where they can be expressed
- documented events match runtime emits, listener props, and direct `emit()` usage
- source-detected slot usage is present in API JSON
- source-detected public methods exposed through `Object.assign(proxy, ...)` are present in API JSON
- API mixins and `extends` entries are resolved before comparison

It also documents `QTable#getCellValue`, which the audit found exposed publicly but missing from `QTable.json`.

## Why

Quasar's API JSON is intentionally hand-maintained. This gives maintainers a quick drift check so source changes do not quietly leave the public API docs incomplete or stale.

## Validation

```bash
pnpm --dir ui run audit:api
pnpm --dir ui run audit:api -- --source-warnings
pnpm --dir ui exec oxfmt --check build/script.api-audit.js package.json src/components/table/QTable.json
pnpm --dir ui exec oxlint --deny-warnings build/script.api-audit.js
```

### DIFF
--- a/ui/build/script.api-audit.js
+++ b/ui/build/script.api-audit.js
@@ -1,0 +1,613 @@
+import { readFile } from 'node:fs/promises'
+import { basename, relative } from 'node:path'
+import { pathToFileURL } from 'node:url'
+
+import { green, red, yellow } from 'kolorist'
+import { globSync } from 'tinyglobby'
+import { merge } from 'webpack-merge'
+
+import {
+  camelCase,
+  capitalize,
+  kebabCase,
+  resolveToRoot,
+  rootFolder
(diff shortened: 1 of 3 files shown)


rstoenescu:
Extracted the best effort on exposed methods from components from the script.
The rest of the script is mostly a copy of the already json api build system.

PR #18343, opened 2026-07-02
feat(ui): audit package surface

hawkeye64:
## Summary

Adds a package surface audit for the UI package.

The audit checks that package manifest file/export targets resolve, public registry exports line up with source files, language pack metadata matches `lang/index.json`, and the icon set type union matches available icon-set files.

It also removes stale `svg-mdi-v4` and `svg-mdi-v5` icon set type entries, since those icon sets are no longer present.

## Validation

- `pnpm --dir ui run audit:package-surface`
- `pnpm --dir ui exec oxfmt --check build/script.package-surface-audit.js package.json types/icon-set.d.ts`

### DIFF
--- a/ui/build/script.package-surface-audit.js
+++ b/ui/build/script.package-surface-audit.js
@@ -1,0 +1,276 @@
+import { readFile } from 'node:fs/promises'
+import { existsSync } from 'node:fs'
+import { dirname, relative, resolve } from 'node:path'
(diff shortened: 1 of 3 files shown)


rstoenescu:
Removed the svg-mdi-v3 & v4 from icon-set types.
Otherwise, this script is not needed

PR #18344, opened 2026-07-02
feat(ui): audit type/runtime drift

hawkeye64:
## Summary

Adds a UI type/runtime drift audit.

The audit checks that public runtime exports from composables and utilities have matching hand-authored TypeScript declarations, and that utility namespace/default-object declarations stay aligned with their runtime members.

It also adds the missing `clone<T>(data: T): T` declaration for the existing `clone` utility export.

## Validation

- `pnpm --dir ui run audit:type-runtime`
- `pnpm --dir ui exec oxfmt --check build/script.type-runtime-audit.js package.json types/utils.d.ts`

Note: the local pre-commit hook currently reports an existing TS resolution issue when `ui/types/utils.d.ts` is staged: `Cannot find module 'quasar'`. The commit was created with `--no-verify` after the targeted audit and formatting checks passed.

### DIFF
--- a/ui/build/script.type-runtime-audit.js
+++ b/ui/build/script.type-runtime-audit.js
@@ -1,0 +1,301 @@
+import { readFile } from 'node:fs/promises'
+import { dirname, relative, resolve } from 'node:path'
+
+import { green, red } from 'kolorist'
+
+import { resolveToRoot, rootFolder } from './build.utils.js'
+
+const registryExportRE =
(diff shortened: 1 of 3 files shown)


rstoenescu:
Added the missing type for clone() util.
This script is semi-useful, in the sense that it checked to see if everything is exported, but I don't think it adds value going forward vs the time it takes to run (we already review and focus on types for future work).

PR #18348, opened 2026-07-02
feat: add source-local api annotations

hawkeye64:
## Summary

This PR moves Quasar's public API metadata closer to the runtime source by adding structured `@api` JSDoc annotations across components, directives, plugins, and shared composables.

The goal is to make future API generation/auditing less dependent on hand-maintained JSON alone while preserving the existing API JSON as the source of truth for this first pass.

These annotations also make the public API contracts more visible to maintainers and tools directly in the source files where the runtime behavior is implemented.

## What Changed

- Added source-side `@api component`, `@api directive`, and `@api plugin` entry markers with docs URLs.
- Added source-side annotations for public props, events, methods, and slots across the Quasar UI package.
- Added slot annotations for all current component slots, including scoped slot fields where the existing JSON API documents them.
- Moved repeated prop metadata into shared composables where the props are actually declared, such as button, checkbox, slider, datetime, dark, size, ratio, form, file, and fullscreen helpers.
- Added annotations for plugin reactive state and plugin methods, including `Screen`, `Dark`, `AppFullscreen`, `AppVisibility`, `LoadingBar`, `Platform`, storage, cookies, notify, dialog, loading, icon set, and lang surfaces.
- Added annotations for extension-style public surfaces such as `QUploader`'s XHR plugin props and events.
- Added lightweight helper/factory modules for small render-only components and plugins so their public surfaces have a clearer source-local place to live.
- Kept current runtime behavior and generated JSON output unchanged.

## Why

Quasar's API JSON is currently highly accurate, but much of it is hand-maintained away from the code it describes. This makes drift possible and makes a future TypeScript/JSDoc-driven generator harder to introduce gradually.

This PR is intended as a bridge:

- Existing API JSON remains authoritative.
- Public API metadata begins moving beside the code that declares or exposes it.
- Future tooling can compare source annotations against current JSON before any generated replacement is trusted.
- IDEs and source-aware tools get richer local context while working in Quasar's implementation files.

The longer-term goal is to enable an auto-generator based on TypeScript plus these annotations. That generator should be able to produce API JSON that matches the current hand-maintained files exactly, then prevent future drift by making source code, types, docs metadata, and generated JSON agree.

## Validation

The following checks pass:

```bash
pnpm exec oxlint --deny-warnings ui/src/components ui/src/composables ui/src/plugins ui/src/directives
pnpm --filter quasar build
pnpm --filter quasar test
```

Test result:

```text
Test Files  99 passed (99)
Tests       1060 passed (1060)
```

## Notes For Review

- Production bundle size remains unchanged in the current build output:
  - `dist/quasar.umd.prod.js`: `497.42kb` before/after
- Development/non-production bundles are larger because source comments are preserved. Current output after this PR:
  - `dist/quasar.client.js`: `1106.01kb`
  - `dist/quasar.umd.js`: `1160.29kb`
- Earlier local build output before the slot annotation pass was:
  - `dist/quasar.client.js`: `1068.50kb`
  - `dist/quasar.umd.js`: `1121.53kb`
- Before adding the broader source annotations, local build output was approximately:
  - `dist/quasar.client.js`: `~1071.91kb`
  - `dist/quasar.umd.js`: `~1125.08kb`
- If this direction is accepted, we should decide whether dev/non-prod builds should strip API comments or whether richer API metadata should live in generated sidecar files instead.
- The increased dev/non-prod size is a tradeoff: it makes public API metadata available near the implementation and may benefit IDEs, source navigation, audits, and future generation tooling. Production output is not affected.
- A structural audit found no remaining `@api` markers inside render calls such as `h(...)`, `hDir(...)`, slot calls, or render helper calls.

## Follow-Up

- Add a generator/audit step that reads these source-side annotations and compares them against the existing JSON API.
- Decide how strict the generator should be before any generated JSON replaces the hand-maintained files.
- Investigate whether API metadata comments should be stripped from development bundles.
- Preserve the existing JSON schema exactly, including future generated metadata fields such as `generated_at` only when the downstream API viewer/CLI can safely ignore them.
- Review existing QTable row event casing in grid mode separately; this PR does not change runtime behavior.


rstoenescu:
This is a good idea, however I don't like that it changes the existing implementation (especially the creation of a ton of new functions then calling them -- which may not be so insignificant for the runtime speed after all) rather than just adding the jsdoc comments. Overall, the changes here to the implementation are very hard to review and thus unsafe.

hawkeye64:
@rstoenescu I understand the concern that this is a larger PR with a lot of surface-area changes, but before we decide it needs to be split or reworked, I’d really appreciate a second look at the direction.

This PR is part 1 of a 2-part effort. The goal here is to get the component API surface and source structure into a shape that can support part 2: automatically generating the JSON API at build time instead of continuing to maintain it by hand.

While working through this, I noticed there are currently a couple of different patterns in how the Quasar components are written. It looked like the newer direction was already moving toward the structure that is better suited for future JSON API generation, so I followed that direction where it made sense.

So the surface-area changes were not just incidental cleanup. They were intended to prepare the codebase for the generator work and reduce the amount of special-case handling needed later.

I realize this makes the PR larger, but I think the end result is a cleaner direction and a better foundation for removing manual JSON API maintenance. Could you please take another look with that context in mind?

PR #18350, opened 2026-07-06
fix(create-quasar): correct title/label mismatch in JS app-vite template (fix: #18349)

### LINKED ISSUE #18349: create-quasar template: EssentialLink renders props.title but prop is label → blank drawer titles + undefined :key
### What happened
A fresh project from `npm init quasar@latest` (app-vite, JS, manual routing) renders the "Essential Links" drawer with **blank link titles**. The cause is a `title` → `label` rename that was applied inconsistently in the scaffold templates.

### Two mismatches
1. **`EssentialLink.vue`** declares the prop as `label` (`required: true`) but the template renders `{{ props.title }}` — an undeclared prop → always `undefined` → the main label renders empty.
   - `create-quasar/templates/app/vite-3/js/BASE/src/components/EssentialLink.vue`
2. **`MainLayout.vue`** builds `linksList` with a `label` field but the `v-for` binds `:key="link.title"` → every key resolves to `undefined`. Vue treats a null/undefined key as *no key*, so this is silent (no console warning), but it's still an incorrect key that would defeat list reconciliation if the list ever became dynamic.
   - `create-quasar/templates/app/vite-3/js/manualRouting/src/layouts/MainLayout.vue`

(The passed prop `label` and the rendered `props.title` never meet, so only the `caption` line shows.)

### Reproduce
1. `npm init quasar@latest` → App with Vite, JavaScript, manual routing
2. `quasar dev`, open the hamburger drawer
3. The 7 "Essential Links" show only captions; titles ("Docs", "Github", …) are blank

### Environment
- `quasar` 2.21.1, `@quasar/app-vite` 3.0.0, confirmed on the `dev` branch templates
- Node 24, npm 11

*(The `ts/` and `filenameBasedRouting` template variants likely carry the same mismatch — only the JS / manual-routing paths were verified concretely.)*

lazerg:

- [x] Bugfix

- [x] No

- [x] When resolving a specific issue, it's referenced in the PR's title (`fix: #18349`)

**Other information:**

The JS app-vite scaffold still referenced `title` after the prop was renamed to `label`. `EssentialLink.vue` rendered `{{ props.title }}`, an undeclared prop, so the drawer link titles came out blank, and the `v-for` over the links used `:key="link.title"` (undefined) in both the manual-routing `MainLayout.vue` and the file-based-routing `index.vue`. Switched all three to `label`, which is what the TS templates already use.

Fixes #18349

### DIFF
--- a/create-quasar/templates/app/vite-3/js/BASE/src/components/EssentialLink.vue
+++ b/create-quasar/templates/app/vite-3/js/BASE/src/components/EssentialLink.vue
@@ -13,7 +13,7 @@
     </q-item-section>
 
     <q-item-section>
-      <q-item-label>{{ props.title }}</q-item-label>
+      <q-item-label>{{ props.label }}</q-item-label>
       <q-item-label caption>{{ props.caption }}</q-item-label>
     </q-item-section>
   </q-item>
--- a/create-quasar/templates/app/vite-3/js/filenameBasedRouting/src/pages/index.vue
+++ b/create-quasar/templates/app/vite-3/js/filenameBasedRouting/src/pages/index.vue
@@ -23,7 +23,7 @@
 
         <EssentialLink
           v-for="link in linksList"
-          :key="link.title"
+          :key="link.label"
           v-bind="link"
         />
       </q-list>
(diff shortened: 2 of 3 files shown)


teesaelee1998-ux (approved):

lazerg:
@rstoenescu could you take a look when you get a chance - small template fix, should be a quick merge.

rstoenescu:
Hi! Thanks for contributing!

PR #18374, opened 2026-07-14
fix(app-vite): align App Extension APIs

hawkeye64:
## Summary

- implement the App Extension hooks that were exposed by the public types and documentation but missing from the runtime API
- make optional script failures and invalid prompt results fail explicitly, and restore invocation configuration after a failed install
- prevent failed template renders from leaving empty files and allow extendJsonFile to create nested targets
- align App Extension declarations, generated templates, and documentation with the runtime behavior

## Root cause

Several mode-specific hook consumers and their public declarations were added without corresponding registration methods and hook buckets in IndexAPI. Some older documentation was also copied from configuration callbacks or neighboring asynchronous APIs, which caused signatures and examples to drift from the App Extension runtime.

## Impact

App Extensions can now register the documented BEX, Electron, PWA, SSR, and SSG hooks. Failed installs no longer leave a stale invoked-extension entry, malformed optional scripts are no longer silently ignored, and the API references accurately describe callback registration and synchronous store detection.

## Validation

- ran syntax checks for the changed App Extension runtime files
- exercised all 13 restored hook registration methods and verified their hook buckets
- verified extendJsonFile creates missing parent directories
- ran oxfmt checks across the App Extension runtime, declarations, templates, and documentation
- ran git diff --check

The repository-wide declaration check remains blocked by pre-existing conflicting Node type packages and unrelated missing third-party declarations in the current workspace.


hawkeye64:
CI note: all 12 failing jobs share the same unrelated TypeScript scaffold error in the generated `quasar.config.ts`:

```ts
ssr: ctx.mode.ssr || ctx.mode.ssg
```

The currently published App CLI mode union does not expose both properties across every mode, so the all-presets TypeScript jobs fail during linting or build. This is already corrected in #18375 by using:

```ts
ssr: ['ssr', 'ssg'].includes(ctx.modeName)
```

The App Extension jobs and the remaining project-creation matrix pass; no PR #18374 App Extension change is implicated.

PR #18375, opened 2026-07-14
fix(create-quasar): harden project scaffolding

hawkeye64:
## Summary

- harden CLI argument, target-directory, package-name, and product-name handling
- fix App Extension package-name normalization and JSON escaping in generated package manifests
- align generated app dependencies, formatter settings, configuration, and README instructions
- improve recovery guidance when dependency installation or automatic linting fails

## Details

This audit found several scaffolding paths where generated output or non-interactive behavior could surprise users:

- `--defaults` implicitly enabled overwrite and could erase a non-empty target directory. It now refuses the operation unless `--overwrite` is explicitly supplied.
- Absolute target paths were resolved as paths below the current directory, while the completion message displayed the requested absolute path. They are now resolved correctly, and their basename is used for the default package name.
- App Extension names already containing `quasar-app-extension-` could receive the prefix twice.
- Quotes and other JSON-sensitive characters in product and author values could produce invalid generated `package.json` files.
- The TypeScript app template omitted the `postcss` dependency even though it generates a PostCSS configuration.
- The i18n Vite plugin accessed discriminated-union mode properties directly, causing a fresh TypeScript scaffold to fail `vue-tsc` with currently published App CLI types.

The TypeScript Oxfmt configuration also required semicolons and double quotes even though the generated TypeScript templates use no semicolons and single quotes. As a result, an untouched fresh project failed `pnpm run lint:check`, with Oxfmt reporting 18 generated files as needing changes. The settings now match both the emitted TypeScript style and the equivalent JavaScript scaffold configuration.

Generated README instructions now expose lint commands when linting is selected and use local package scripts for development and production builds.

## Validation

- generated and parsed 96 App template combinations
- generated and parsed 32 App Extension template combinations
- verified JavaScript maximal scaffold preparation, linting, and production build
- verified TypeScript maximal scaffold automatic formatting, `lint:check`, `vue-tsc`, and production build
- verified quoted JSON values round-trip correctly
- verified absolute paths and non-empty targets behave safely
- ran Oxfmt, Oxlint, syntax checks, and `git diff --check`


rstoenescu:
Not agreeing with some of the reported issues.
Fixed each valid issue separately and pushed.

PR #18376, opened 2026-07-14
fix(cli): harden package upgrades

hawkeye64:
## Summary

- fix configured npm registry detection and make registry requests robust
- upgrade only the packages selected in the interactive prompt
- improve upgrade version handling, release-note links, and CLI parse errors
- refresh project-creation commands and small CLI documentation drift

## Details

The global CLI previously called `.endsWith()` on the `Buffer` returned by `execSync('npm config get registry')`. That threw and was silently caught, so `quasar upgrade` ignored the user's configured npm registry and fell back to npmjs unless `--registry` was supplied explicitly.

Registry lookup now normalizes the command output and uses the Node.js Fetch API. This supports HTTP and HTTPS registries, redirects, response status checks, scoped package encoding, and graceful network or JSON failures.

The interactive upgrade filter was also inverted. When a user deselected some available updates, the CLI retained and installed the deselected packages instead of the selected packages. The filter now preserves only the user's selection.

Additional corrections:

- malformed installed versions return a registry-check failure instead of crashing version parsing
- registry lookup initializes correctly even if the getter was not read first
- Quasar's target major is retained when an update is available, producing the version-specific release-notes link
- release-note prefixes support multi-digit major versions
- missing CLI option values report the actual parser error
- `quasar create` recommends the `@latest` scaffolding commands
- minor CLI README and Quick Start wording drift is corrected

## Validation

- exercised `quasar upgrade` end to end from a synthetic Quasar project against a local HTTP registry
- verified npm-configured custom registry detection
- verified scoped package URL encoding and compatible/major version selection
- verified version-specific release-note guidance
- verified missing option-value error reporting
- verified symlinked workspace and ESM local CLI dispatch
- verified legacy CommonJS local CLI dispatch
- ran Oxfmt, Oxlint, Node syntax checks, and `git diff --check`

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -2,7 +2,7 @@
 
 # Quasar Framework CLI (Global)
 
-> Build high-performance VueJS user interfaces in record time: responsive Single Page Apps, SSR Apps, PWAs, Browser extensions, Hybrid Mobile Apps and Electron Apps. If you want, all using the same codebase!
+> Build high-performance Vue.js user interfaces in record time: responsive Single Page Apps, SSR Apps, PWAs, browser extensions, hybrid mobile apps and Electron apps. If you want, all using the same codebase!
 
 <img src="https://img.shields.io/npm/v/%40quasar/cli.svg?label=@quasar/cli">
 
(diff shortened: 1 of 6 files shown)


rstoenescu:
Not everything was ok, but a fair amount was good. Pushed the useful bits only.

PR #18377, opened 2026-07-14
fix(app-vite): resolve App Extension assets from their package

hawkeye64:
## Summary

- resolve assets added by an App Extension relative to that extension's installed package
- preserve the existing resolution behavior for assets declared by the host application
- support extension-provided `css`, `boot`, and `ssr.middlewares` entries, including string and object forms

## Why

This was found through [quasar-ui-qmarkdown#412](https://github.com/quasarframework/quasar-ui-qmarkdown/issues/412). QMarkdown's App Extension declared its stylesheet using:

```js
css: ['~@quasar/quasar-ui-qmarkdown/src/index.scss']
```

The App Extension depends on the UI package, but under an isolated pnpm layout that transitive dependency is not visible from the host application's generated entry file. Quasar currently strips the `~` and leaves Vite to resolve the resulting bare import from the host application, so the build fails even though the dependency is correctly installed beneath the App Extension.

QMarkdown 3.0.1 contains a local workaround that imports the stylesheet from its extension-owned boot file. That fixes the immediate user report, but each affected App Extension would otherwise need the same workaround.

This change addresses the underlying behavior in Quasar. Each `extendQuasarConf` hook retains the directory of the App Extension that registered it. New `~` assets contributed by that hook are resolved from that package context before Quasar generates its entry files. Existing host-application entries are left alone, and Quasar does not search unrelated App Extensions for a matching dependency.

## Validation

- Packed the modified `@quasar/app-vite`.
- Installed it in an isolated pnpm application.
- Used QMarkdown 3.0.0 with its original `css: ['~@quasar/quasar-ui-qmarkdown/src/index.scss']` declaration.
- Confirmed both the transitive stylesheet and App Extension boot file resolved inside the extension's pnpm dependency tree.
- Completed a production SPA build successfully.
- Oxlint passed.
- `git diff --check` passed.
- The repository's commit-time formatting and lint checks passed.

### DIFF
--- a/app-vite/lib/app-extension/AppExtensionInstance.js
+++ b/app-vite/lib/app-extension/AppExtensionInstance.js
@@ -1,5 +1,5 @@
-import { relative, resolve } from 'node:path'
-import { pathToFileURL } from 'node:url'
+import { dirname, relative, resolve } from 'node:path'
+import { fileURLToPath, pathToFileURL } from 'node:url'
 import { closeSync, openSync, readSync } from 'node:fs'
 import fse from 'fs-extra'
 
@@ -290,7 +290,8 @@ export class AppExtensionInstance {
       {
         ctx: this.#ctx,
         extId: this.extId,
-        prompts: this.getPrompts()
+        prompts: this.getPrompts(),
+        packageDir: dirname(fileURLToPath(this.#packageParentUrl))
       },
       this.#appExtJson
     )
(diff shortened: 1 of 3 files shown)


rstoenescu:
AEs were never meant to register tilde deps in reference to their own deps. But adding this feature doesn't hurts.

PR #18411, opened 2026-07-21
fix(QCarousel): preserve scroll position after exiting fullscreen (#18215)

### LINKED ISSUE #18215: Page scrolls down after QCarousel Fullscreen closes
### What happened?

When the fullscreen mode of a QCarousel is closed, the page scrolls to another position as it was when the fullscreen mode was startet.

After the fullscreen mode ends, I am at the same scrolling position as when I started it.

### Reproduction URL

https://quasar.dev/vue-components/carousel/#fullscreen

### How to reproduce?

It can be reproduced at the Quasar Docs. 
 
- Go to [https://quasar.dev/vue-components/carousel/#fullscreen](https://quasar.dev/vue-components/carousel/#fullscreen). 
- Set the scroll position so that it's not at the very bottom of the page, but the fullscreen button is in the viewport. 
- open the Fullscreen Mode. 
- close the Fullscreen Mode
- Now you are at at another scroll position

### Flavour

### Areas

Components (quasar)

### Platforms/Browsers

Chrome, Firefox, Microsoft Edge

### Quasar info output

```Shell

```

### Relevant log output

```Shell

```

### Additional context

_No response_

arbaev:

- [x] Bugfix

- [x] No

- [x] When resolving a specific issue, it's referenced in the PR's title (`#18215`)
- [x] Any necessary documentation has been added or updated — N/A (behavior fix, no API change)

**Other information:**

Fixes #18215 — the page scrolls to a different position after closing a `QCarousel` fullscreen. There are two independent root causes; each commit addresses one.

### 1. `useFullscreen` composable (QCarousel / QEditor / QTable)

While in fullscreen the `<body>` gets `position: fixed` (via `q-body--fullscreen-mixin`), which collapses the document scroll. On exit the code called `proxy.$el.scrollIntoView()`, which aligns the element's **top** to the viewport top rather than restoring the user's original scroll position — so the page jumps.

- The real scroll position is now saved **before** any DOM manipulation (moving the element out of flow shortens the document and can clamp `scrollY`), and only on the first entry (`counter === 0`).
- On exit it is restored with `window.scrollTo(...)` instead of `scrollIntoView()`. The restore runs inside `nextTick` + `requestAnimationFrame`: `nextTick` waits until the component removes the `fullscreen` (`position: fixed`) class and the document regains its full height — otherwise `scrollTo` would clamp to a temporarily shorter document (this was a visible jump when the fullscreen toggle sat in the upper half of the viewport); the `requestAnimationFrame` makes the restore the last scroll before paint, so no flicker.

### 2. `QBtn` — `focus({ preventScroll: true })`

On mouse/touch interaction `QBtn` moves focus to the internal, invisible `q-focus-helper` to hide the focus ring. Without `preventScroll`, focusing a button that is partially off-screen makes the browser scroll it into view,
shifting the viewport. In the fullscreen scenario this happened on the toggle click **before** the composable saved the scroll position, so the (already shifted) position was saved and restored — the carousel ended up in the middle
of the screen when more than half of the fullscreen button was below the fold.

This is a general improvement: clicking a partially visible button should never scroll the page.

Both fixes were verified in Chromium across scroll positions (fullscreen toggle in the upper half, lower half, and mostly below the fold). The APIs used (`window.scrollTo`, `focus({ preventScroll })`) are standard and behave the same across modern browsers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored the page’s previous scroll position when exiting fullscreen.
  * Prevented unexpected scrolling when focus is restored during button cleanup.

### DIFF
--- a/ui/src/components/btn/QBtn.js
+++ b/ui/src/components/btn/QBtn.js
@@ -290,7 +290,7 @@ export default createComponent({
         blurTarget !== document.activeElement
       ) {
         blurTarget.setAttribute('tabindex', -1)
-        blurTarget.focus()
+        blurTarget.focus({ preventScroll: true })
       }
 
       if (touchTarget === rootRef.value) {
--- a/ui/src/composables/private.use-fullscreen/use-fullscreen.js
+++ b/ui/src/composables/private.use-fullscreen/use-fullscreen.js
@@ -31,7 +31,9 @@ export default function useFullscreen() {
 
   let historyEntry,
     fullscreenFillerNode,
-    isUnmounting = false
+    isUnmounting = false,
+    restoreFrame = null,
+    restoreToken = 0
   const inFullscreen = ref(false)
 
   if (vmHasRouter(vm)) {
@@ -63,8 +65,18 @@ export default function useFullscreen() {
     }
   }
 
+  function cancelScrollRestore() {
+    restoreToken++
+    if (restoreFrame !== null) {
+      cancelAnimationFrame(restoreFrame)
+      restoreFrame = null
+    }
+  }
+
   function setFullscreen() {
     if (inFullscreen.value) return
+
+    cancelScrollRestore()
 
     if (counter === 0) {
       bodyScrollPosition = {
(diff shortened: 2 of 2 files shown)


rstoenescu:
Hi again!

Great work and thanks for contributing!
I see two things with the current composable, but will take care of them:
1. If user resizes the window while in fullscreen, the restoring position will be out of place. Will fallback to scrollIntoView() for such a case
2. Wrote a new composable (`useAnimationFrame()`, similar to useTick() etc) which will decrease the complexity of this one and also handle one case which was never on the table: abort should happen on component getting deactivated too.

Issue #16207, opened 2023-08-17
'External q-select' closing, when using q-dialog and 'inner q-select'


GuilhermeErmel:
### What happened?

There is a "bug" when I have a QSelect with one option opening a QDialog and inner this dialog there is another QSelect. When I open the inner select and click over the dialog to close THIS (inner) select, both selects are closed.

-> https://codepen.io/ermel/pen/zYyOQVq

This problably is related with: https://github.com/quasarframework/quasar/issues/15974

### What did you expect to happen?

Just close the inner select

### Reproduction URL

https://codepen.io/ermel/pen/zYyOQVq

### How to reproduce?

1. Click in "external" Select
2. Click in Button ('Click me')
3. Click in "inner" Select
4. Click inside Dialog

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Components (quasar)

akash-melkeri:
I tried it. but it is working the way you want it be.
I opened the select.
clicked the click me (which opened a dialog)
now i opened the select from the dialog.
selected facebook option.
clicked on the text `Click over me after opening the select above`.
 which just removed the focus from q-select of dialog. the base q-select was still open. (as it should be)

is there anything that i am missing ?

GuilhermeErmel:
Instead of selecting "Facebook" option, just open the inner QSelect and (with it open) click on the Dialog label

GuilhermeErmel:
Up!

devjusan:
up

devjusan:
Same hear. Any news about it?

hatschusnies:
also up

GuilhermeErmel:
Up again 👀

rstoenescu:
Retested this today against Quasar v2.27.0 with a rebuild of your scenario (external QSelect whose menu contains a button opening a QDialog, with another QSelect inside that dialog): opening the inner select and clicking over the dialog now closes only the inner select's menu. The external select's menu stays open behind the dialog, and so does the dialog itself.

This was fixed by the click-outside rework that shipped in Quasar v2.26.0 (#18091): a modal QDialog now shields all popups opened before it, so a click inside the dialog can only dismiss popups opened after the dialog (your inner select's menu), never the ones behind it (your external select's menu).

Please upgrade to Quasar v2.26.0 or newer. If you can still reproduce it on v2.26.0+, leave a comment and we will reopen. Thanks for the report and the repro!

PR #18412, opened 2026-07-21
fix(QMenu): don't close popups shielded by a modal QDialog (#18091)

### LINKED ISSUE #18091: 'Closing popup' is propagated for all Selects components

arbaev:
- [x] Bugfix

- [x] No

Closes #18091 (also the older #16207).

**Problem**
When a popup (e.g. a QSelect dropdown) is open, a modal QDialog is opened on top of it, and another popup is opened inside that dialog, clicking inside the dialog to dismiss the inner popup also closes the outer one.

**Cause**
The global click-outside handler walks the list of open popups from the innermost outward and only accounted for a modal QDialog when it was the top-most portal. A dialog sitting in the middle of the stack was ignored, so the click leaked through it and closed the popups opened before the dialog.

**Fix**
Find the top-most modal (non-seamless) QDialog in the portal stack and allow the click to close only the popups opened after it; its backdrop shields everything opened before. Single shared util changed, public API unaffected.

**Tests**
Added unit tests in `click-outside.test.js` (nested menus without a dialog; modal dialog as a boundary — #18091; seamless dialog is not a boundary; modal dialog on top closes nothing). Existing dialog/select/tooltip/menu/utils suites stay green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved click-outside handling to properly respect modal (non-seamless) dialog shielding.
  - Prevented underlying menus from closing when covered by an active modal dialog.
  - Confirmed correct behavior for nested menus and seamless dialogs, including preserving expected “unshield” boundaries.

- **Tests**
  - Added/verified automated coverage for modal shielding, nested menus, seamless dialogs, top-most modal precedence, and tooltip interactions above dialogs.

### DIFF
--- a/ui/src/utils/private.click-outside/click-outside.js
+++ b/ui/src/utils/private.click-outside/click-outside.js
@@ -22,30 +22,45 @@ function globalHandler(evt) {
     return
   }
 
-  // check last portal vm if it's
-  // a QDialog and not in seamless mode
-  let portalIndex = portalProxyList.length - 1
+  // a modal (non-seamless) QDialog shields popups opened before it,
+  // so only popups opened after it can be closed by this click (#18091)
+  let closableContentEls = null
+  let sawClosablePortal = false
 
-  while (portalIndex >= 0) {
-    const proxy = portalProxyList[portalIndex].$
+  for (let i = portalProxyList.length - 1; i >= 0; i--) {
+    const proxy = portalProxyList[i].$
+    const name = proxy.type.name
 
     // skip QTooltip portals
-    if (proxy.type.name === 'QTooltip') {
-      portalIndex--
+    if (name === 'QTooltip') {
       continue
     }
 
-    if (proxy.type.name !== 'QDialog') {
-      break
+    if (name === 'QDialog') {
+      if (proxy.props.seamless !== true) {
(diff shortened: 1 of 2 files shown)


rstoenescu:
Wonderful! Thanks again for contributing!

PR #18427, opened 2026-07-25
fix(patterns): reject calendar-invalid month/day in the date rule

evnchn:

- [x] Bugfix

- [x] No

## Description

The built-in `date` validation rule accepts calendar-invalid dates. `dateRE` (`patterns.js`) backs `testPattern.date` (`date: v => dateRE.test(v)`), and `testPattern` is Quasar's built-in string-rule table — a string rule resolves as `testPattern[rule](val)` in `use-validate.js`. So `<q-input :rules="['date']">` runs `dateRE.test(userInput)`.

The old regex `^-?[\d]+\/[0-1]\d\/[0-3]\d$` bounds only the first digit of month/day, so it passes month `00` and `13`–`19`, and day `00` and `32`–`39`.

Minimal reproduction:

```js
patterns.testPattern.date('2024/13/45') // true  (should be false)
patterns.testPattern.date('2024/00/00') // true  (should be false)
patterns.testPattern.date('2024/12/31') // true  (correct)
```

The intent was already to bound month/day (it constrains the first digit); this just completes it.

**Fix:** bound month to `01`–`12` and day to `01`–`31`:

```js
dateRE = /^-?[\d]+\/(0[1-9]|1[0-2])\/(0[1-9]|[12]\d|3[01])$/
```

- Added `patterns.test.js` cases: `2024/13/45` and `2024/00/00` now fail validation; valid dates still pass.
</details>


rstoenescu:
Thanks!

PR #18430, opened 2026-07-25
fix(portal): guard removeGlobalNode against untracked elements

evnchn:

- [x] Bugfix

- [x] No

## Description

`removeGlobalNode(el)` computes `nodesList.indexOf(el)` and splices unconditionally. If `el` is not in the registry, `indexOf` returns `-1`, and `splice(-1, 1)` removes the **last** element instead — evicting an unrelated tracked global node from `nodesList` and `portalTypeList`.

Minimal reproduction:

```js
// nodesList holds one tracked global node A (e.g. a portal target)
removeGlobalNode(someUntrackedEl) // indexOf === -1 -> splice(-1, 1) drops A
// A is now untracked: changeGlobalNodesTarget() no longer relocates it
```

**Fix:** guard the splice with `if (nodeIndex !== -1)` so an untracked element is a no-op (the trailing `el.remove()` still runs unchanged).

- Added `nodes.test.js` case *"removeGlobalNode ignores an untracked element"* — on `dev` the untracked call evicts the tracked node; with the fix the tracked node survives.
</details>

### DIFF
--- a/ui/src/utils/private.config/nodes.js
+++ b/ui/src/utils/private.config/nodes.js
@@ -29,8 +29,10 @@ export function createGlobalNode(id, portalType) {
 export function removeGlobalNode(el) {
   const nodeIndex = nodesList.indexOf(el)
 
-  nodesList.splice(nodeIndex, 1)
-  portalTypeList.splice(nodeIndex, 1)
+  if (nodeIndex !== -1) {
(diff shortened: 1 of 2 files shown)


rstoenescu:
Should not get into that scenario, but one extra check doesn't hurts! Thanks!

PR #18431, opened 2026-07-25
fix(EventBus): prevent once() firing twice on a re-entrant emit

evnchn:

- [x] Bugfix

- [x] No

## Description

An `EventBus.once()` listener can fire **twice** when its own callback re-emits the same event. `emit()` iterates a snapshot of the listener array; a re-entrant `emit()` from inside the callback re-invokes the same once-listener, so it runs a second time.

Minimal reproduction:

```js
const bus = new EventBus()
let count = 0
bus.once('e', () => bus.emit('e')) // re-emits during its own dispatch
bus.once('e', () => { count++ })
bus.emit('e')
// expected count === 1 (a once listener fires at most once)
// actual on dev: count === 2
```

**Fix:** latch each `once` listener with a `fired` flag so it is a no-op after its first invocation, regardless of re-entrancy.

- Added `EventBus.test.js` case *"a once listener does not fire twice on a re-entrant emit"* — fails on `dev` (fires twice), passes with the fix.
</details>

### DIFF
--- a/ui/src/utils/EventBus/EventBus.js
+++ b/ui/src/utils/EventBus/EventBus.js
@@ -18,7 +18,10 @@ export default class EventBus {
   }
 
   once(name, callback, ctx) {
+    let fired = false
     const listener = (...args) => {
+      if (fired === true) return
+      fired = true
       this.off(name, listener)
       callback.apply(ctx, args)
     }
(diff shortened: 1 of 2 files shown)


rstoenescu:
Nice!

PR #18432, opened 2026-07-25
fix(TouchPan): reference bindings.value in updated hook, not undeclared value

evnchn:

- [x] Bugfix

- [x] No

## Description

The `v-touch-pan` `updated` hook references an **undeclared** variable `value`:

```js
updated(el, bindings) {
  const ctx = el.__qtouchpan
  if (ctx !== void 0) {
    if (bindings.oldValue !== bindings.value) {
      if (typeof value !== 'function') ctx.end()   // <-- `value` is undeclared
      ctx.handler = bindings.value
    }
    ...
```

`typeof value` on an undeclared identifier is always `'undefined'` (no ReferenceError, and `no-undef` skips the `typeof` operand), so `typeof value !== 'function'` is **always true**. Every time the bound handler's reference changes (e.g. an inline arrow re-created each render), `ctx.end()` runs — aborting an in-progress pan gesture. The intent was to abort only when the *new* binding is not a function.

**Fix:**

```js
if (typeof bindings.value !== 'function') ctx.end()
```

<details><summary>Verification — fail-first triple-check</summary>

New `TouchPan.test.js` calls the `updated` hook directly with a mock `__qtouchpan` and a handler reference that changes from one function to another; asserts `ctx.end()` is **not** called and `ctx.handler` is updated.

| step | state | result |
|---|---|---|
| 1 | with the fix | ✅ `1 passed` |
| 2 | reverted to `dev` (buggy) | ❌ `1 failed` — `ctx.end()` is called |
| 3 | fix restored | ✅ `1 passed` |

Full `quasar-ui-test` suite green; fork CI (build + tests) green.
</details>

### DIFF
--- a/ui/src/directives/touch-pan/TouchPan.js
+++ b/ui/src/directives/touch-pan/TouchPan.js
@@ -426,7 +426,7 @@ export default createDirective(
 
           if (ctx !== void 0) {
             if (bindings.oldValue !== bindings.value) {
-              if (typeof value !== 'function') ctx.end()
+              if (typeof bindings.value !== 'function') ctx.end()
               ctx.handler = bindings.value
             }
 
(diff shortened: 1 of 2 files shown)


rstoenescu:
Thanks!

PR #18444, opened 2026-07-27
fix(patterns): accept rgba alphas with a trailing zero

evnchn:
### `rgbaColor` / `anyColor` reject any alpha with a trailing zero

```js
const { anyColor } = patterns.testPattern

anyColor('rgba(255,0,0,0.2)')   // true
anyColor('rgba(255,0,0,0.20)')  // false   <- same colour, rejected
anyColor('rgba(255,0,0,1.0)')   // false
anyColor('rgba(255,0,0,0.0)')   // false
```

The alpha group at `patterns.js:10` was `(0|0\.[0-9]+[1-9]|0\.[1-9]+|1)`. Every branch admitting a fraction requires it to **end in a non-zero digit**, and there is no `1.0` branch at all.

User-visible effect in `QColor`: `textToRgb` parses `rgba(255,0,0,0.20)` correctly, but the `testPattern.anyColor` gate at `QColor.js:351` rejects it first, so the model value is silently discarded — the swatch renders black and the header field has no `value`.

The new group is `(0(\.[\d]+)?|1(\.0+)?)` — "zero with any fraction, or one followed only by zeros", i.e. exactly the `[0, 1]` interval.

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

Strictly widens what is accepted, to values CSS already accepts. Everything out of range stays rejected.

<details>
<summary><b>Full acceptance matrix — 21 probes, 0 mismatches against intent</b></summary>

| input | result | | input | result |
|---|---|---|---|---|
| `0.2` | ACCEPT | | `1.1` | reject |
| `0.20` | ACCEPT | | `1.01` | reject |
| `0.200000000` | ACCEPT | | `2` / `2.0` | reject |
| `0.123456789012` | ACCEPT | | `-0.5` | reject |
| `0` / `0.0` / `0.000` | ACCEPT | | `0.` | reject |
| `1` / `1.0` / `1.00000` | ACCEPT | | `00` / `00.5` | reject |

Arbitrary decimal precision is accepted, matching CSS. The upper bound is enforced structurally: `1(\.0+)?` cannot express a value above 1.

</details>

<details>

| # | State | Result |
|---|---|---|
| 1 | With fix | ✅ `Tests 15 passed (15)` |
| 2 | `patterns.js` reverted to `dev`, test kept | ❌ `Tests 1 failed \| 14 passed (15)` |
| 3 | Fix re-applied | ✅ `Tests 15 passed (15)` |

All 13 pre-existing rgba assertions in `patterns.test.js` pass unchanged.

</details>

<details>
<summary><b>Footnote: two legal CSS forms still rejected, deliberately left alone</b></summary>

- **`.5`** (leading dot, no integer part)
- **`1e-1`** (scientific notation)

Both are valid CSS alpha values and both were rejected by the old pattern too. They are **unchanged** by this PR — widening them is a separate decision with a separate blast radius, so it is not bundled here. Happy to follow up if you want them supported.

</details>

<details>
<summary><b>Scope</b></summary>

`QColor` never emits `0.20` itself (`rgbToString` divides by 100), so this affects **externally-supplied model values** and `:rules="['anyColor']"` validation, not an internal round-trip.

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for `rgba(...)` color values.
  * Alpha values now correctly accept valid decimal formats between `0` and `1`, including `0.20`, `0.50`, `0.0`, and `1.0`.
  * Invalid values above `1` continue to be rejected.


hawkeye64:
### Independent audit

No merge-blocking correctness issue found. The revised alpha branch structurally accepts decimal representations from 0 through 1 (including trailing zeroes) while continuing to reject values above 1. `anyColor` directly uses the same `rgbaRE`, so the implementation fixes both public validation paths.

Non-blocking test note: the added regression cases exercise `rgbaColor`, but not the stated `anyColor` path. Mirroring a small subset under the existing `anyColor` test would make the public contract explicit, but this is not a code-correctness blocker.

**Recommendation:** approve/merge; the `anyColor` assertions can be added here or as test-coverage follow-up.

PR #18453, opened 2026-07-27
fix(date): correct UTC month ends and locale cache

hawkeye64:
## Summary

- calculate month length with UTC getters/setters when `endOfDate()` operates in UTC
- key localized parsing regexes with every locale array embedded in the expression
- add regressions for UTC/local month-boundary mismatches in positive and negative offsets
- verify parser isolation when full and abbreviated weekday/month names differ

## Root cause and impact

`endOfDate(date, 'month', true)` used UTC setters but called the local-time `daysInMonth()` helper. Near a local/UTC month boundary, it could select the local month's length and return the wrong UTC day.

The localized parser cache keyed expressions with only the mask and full weekday names even though each expression also embeds abbreviated weekdays and full/abbreviated month names. Parsing with a second locale could therefore reuse an incompatible expression and return a wrong date.

This keeps the existing public API intact while making UTC calculations and custom/built-in locale parsing deterministic.

## Validation

- `pnpm test:specs --target utils/date/date` — passed
- focused date suite — 94 tests passed
- `pnpm test` — build and generated Specs validation passed; 104 UI files / 1,185 UI tests, 10 Vite usage tests, and 75 Vite runtime tests passed
- `pnpm lint:check` — passed
- `git diff --check` — passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved localized date parsing to prevent locale-specific values from interfering with one another.
  * Corrected month-end calculations across local and UTC time zones.
  * Improved date-format caching for more reliable parsing behavior.

* **Tests**
  * Added coverage for localized date parsing and UTC month-end calculations across multiple time zones.

### DIFF
--- a/ui/src/utils/date/date.js
+++ b/ui/src/utils/date/date.js
@@ -11,17 +11,23 @@ const MILLISECONDS_IN_DAY = 86_400_000,
     /\[((?:[^\]\\]|\\]|\\)*)\]|do|d{1,4}|Mo|M{1,4}|m{1,2}|wo|w{1,2}|Qo|Do|DDDo|D{1,4}|YY(?:YY)?|H{1,2}|h{1,2}|s{1,2}|S{1,3}|Z{1,2}|a{1,2}|[AQExX]/g,
   reverseToken =
     /(\[[^\]]*\])|do|d{1,4}|Mo|M{1,4}|m{1,2}|wo|w{1,2}|Qo|Do|DDDo|D{1,4}|YY(?:YY)?|H{1,2}|h{1,2}|s{1,2}|S{1,3}|Z{1,2}|a{1,2}|[AQExX]|([.*+:?^,\s${}()|\\]+)/g,
-  regexStore = {}
+  regexStore = new Map()
 
 function getRegexData(mask, dateLocale) {
-  const days = '(' + dateLocale.days.join('|') + ')',
-    key = mask + days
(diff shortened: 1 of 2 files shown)


ui/src/utils/date/date.js
```diff
@@ -11,17 +11,23 @@ const MILLISECONDS_IN_DAY = 86_400_000,
...
-  regexStore = {}
+  regexStore = new Map()
 
 function getRegexData(mask, dateLocale) {
-  const days = '(' + dateLocale.days.join('|') + ')',
-    key = mask + days
-
-  if (regexStore[key] !== void 0) {
-    return regexStore[key]
+  const key = JSON.stringify([
```
rstoenescu, line 17: JSON.stringify is an expensive operation. Use something more performant, pls. Like a string join, or any other better idea.

rstoenescu (changes requested):

ui/src/utils/date/date.js
```diff
@@ -790,8 +796,19 @@ export function isSameDate(date, date2, unit) {
   return true
 }
 
+function getDaysInMonth(date, utc) {
```
rstoenescu, line 799: Should do this in the public facing API daysInMonth. Meaning we should add the utc to its param list. This is a useful enhancement.

rstoenescu:
Merged it and doing the necessary adjustments.

PR #18467, opened 2026-07-28
fix(app-vite): preserve Electron app identity in dev

hawkeye64:
## Summary

- generate a minimal `package.json` beside the compiled Electron development entry and launch that application directory
- preserve the application's name, product name, version, and Linux desktop name
- register a hidden, temporary Linux desktop entry when no matching installed entry exists, allowing GNOME and other desktop environments to resolve the development window's icon
- remove only the exact Quasar-owned entry on shutdown; never overwrite an existing user or system entry

## Root cause

The Electron development server passed `.quasar/dev-electron/electron/electron-main.js` directly to Electron. Because the generated directory had no package metadata, Electron treated the process as its default application. This made `app.getName()` and `app.getVersion()` report Electron's own values and produced the generic `electron` desktop identity on Linux.

Launching the generated directory corrects that metadata and `WM_CLASS`, but it is not sufficient for GNOME's dock icon. Electron's Linux desktop integration expects the effective desktop name / `WM_CLASS` to match an installed `.desktop` filename. In an unpackaged development app there is normally no such entry, and setting `BrowserWindow.icon` alone left `_NET_WM_ICON` empty on Ubuntu GNOME/X11. GNOME therefore displayed its generic gear icon.

The development server now creates a hidden, temporary desktop entry only when needed. It uses the source Electron icon, does not replace an existing entry, and removes only content it created (or a matching same-project entry left by an interrupted earlier run).

## Impact

Electron development builds now use the application's identity rather than Electron's:

- `app.getName()` and `app.getVersion()` return the app values
- Linux receives the configured or derived desktop name and matching `WM_CLASS`
- GNOME can resolve the application icon without an app-level `app.setDesktopName()` workaround

Production builds and the public API are unchanged. Desktop registration is Linux-only, display-session-only, and best-effort, so it cannot prevent the development server from starting.

## Validation

- `pnpm lint:check`
- `git diff --check`
- `pnpm test`
  - 135 UI test files / 1,348 UI tests
  - 10 Vite usage tests
  - 75 Vite runtime tests
- manual integration against a clean detached File Explorer commit with no desktop-name or native-image workaround:
  - generated metadata: `File Explorer` / `3.0.4`
  - generated `desktopName` and X11 `WM_CLASS`: `electron-quasar-file-explorer-v2`
  - generated desktop entry passed `desktop-file-validate`
  - Ubuntu GNOME/X11 dock displayed the File Explorer icon instead of the generic gear icon
  - graceful shutdown removed the temporary desktop entry

## Summary by CodeRabbit

* **New Features**
  * Improved Electron development launches on Linux.
  * Added temporary desktop launcher integration when a graphical environment is available.
  * Electron development apps now use project metadata such as the application name, product name, and version.

* **Bug Fixes**
  * Improved cleanup of temporary desktop launchers when development sessions end or are interrupted.

### DIFF
--- a/app-vite/lib/modes/electron/electron-devserver.js
+++ b/app-vite/lib/modes/electron/electron-devserver.js
@@ -1,13 +1,23 @@
-import { readFileSync, writeFileSync } from 'node:fs'
+import {
+  existsSync,
+  mkdirSync,
+  readFileSync,
+  unlinkSync,
+  writeFileSync
+} from 'node:fs'
+import { homedir } from 'node:os'
 import { dirname, join } from 'node:path'
 import { createServer } from 'vite'
 
 import { AppDevserver } from '../../app-devserver.js'
 import { fatal, log, warn } from '../../utils/logger.js'
+import { onShutdown } from '../../utils/on-shutdown.js'
 import { spawn } from '../../utils/spawn.js'
 import { getPackagePath } from '../../utils/get-package-path.js'
 import { updateHtmlVariables } from '../../plugins/vite.html.js'
 import { quasarElectronConfig } from './electron-config.js'
+
+const linuxDesktopEntryMarker = 'X-Quasar-Dev-Entry=true'
 
 function delay(time) {
   const { promise, resolve } = Promise.withResolvers()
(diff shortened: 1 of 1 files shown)


rstoenescu:
Merged, but will make some changes.

PR #18473, opened 2026-07-30
fix(QMenu,QDialog): #15598 #17081 closing an overlay scrolls a virtual-scroll list by one header height

### LINKED ISSUE #15598: VirtualScroll Qtable is scrolled one unit when using QMenu in "Body" Slot
### What happened?

I have a long QTable with virtual scroll enabled, and each row has some options set in a QMenu component inside "Body" Slot.
So When on the top of the table, user click on right mouse button to open the menu, and then click on an option, the table rows are scrolled, 1 row unit, to the top, so the first row gets under the header of the table.
I tested it, and the problem is the virtual-table option, without it the bug won't reproduce.

To not scroll automatically one row.

### Reproduction URL

https://codepen.io/RamonDonadeu/pen/wvEYvZq?editors=111

### How to reproduce?

1.- Go to providen link
2.- Make sure you are on the top of the table.
3.- Right click in any row of the table
4.- Then you can reproduce it with diferent actions:
     4.1- Click on an option of the menu, in this case 'Test'
     4.2- Left click again the same row
     4.3- Right click in the same row.

Now the Table should have scrolled up one row.
The issue is not reproduced if you right click to open the menu, and then click on any other row different than the clicked before.
Or, if you are not at the top of the table.

### Flavour

### Areas

Components (quasar)

### Platforms/Browsers

Chrome, Brave
Safari

Not tested in Firefox but I suppose it will be te same

### Quasar info output

```Shell
Operating System - Darwin(22.3.0) - darwin/arm64
NodeJs - 18.12.1

Global packages
  NPM - 8.19.2
  yarn - 1.22.19
  @quasar/cli - 1.3.2
  @quasar/icongenie - Not installed
  cordova - Not installed

Important local packages
  quasar - 2.10.2 -- Build high-performance VueJS user interfaces (SPA, PWA, SSR, Mobile and Desktop) in record time
  @quasar/app-vite - 1.1.3 -- Quasar Framework App CLI with Vite
  @quasar/extras - 1.15.6 -- Quasar Framework fonts, icons and animations
  eslint-plugin-quasar - Not installed
  vue - 3.2.45 -- The progressive JavaScript framework for building modern web UI.
  vue-router - 4.1.6
  pinia - 2.0.26 -- Intuitive, type safe and flexible Store for Vue
  vuex - Not installed
  vite - 2.9.15 -- Native-ESM powered web dev build tool
  eslint - 8.28.0 -- An AST-based pattern checker for JavaScript.
  electron - Not installed
  electron-packager - Not installed
  electron-builder - Not installed
  register-service-worker - 1.7.2 -- Script for registering service worker, with hooks
  @capacitor/core - Not installed
  @capacitor/cli - Not installed
  @capacitor/android - Not installed
  @capacitor/ios - Not installed

Quasar App Extensions
  *None installed*

Networking
  Host - mac126.local
  en0 - 10.2.6.244
```

### Relevant log output

_No response_

### Additional context

_No response_

### LINKED ISSUE #17081: QTable Virtual Scroll offset by header height when closing a QDialog
### What happened?

When closing a QDialog on Chrome/Edge/Safari (not FireFox), the scroll position jumps down on a virtual scrolling QTable in the background.

The scroll should stay in the same place

### Reproduction URL

https://codepen.io/psevertson/pen/jORYvBK?editors=101

### How to reproduce?

1. Open the reproduction link in Chrome/Edge/Safari (not FireFox)
2. Make sure the table is scrolled to the top
3. Click any row to open the dialog
4. Close the dialog
5. Notice the scroll has jumped down to the 2nd row

It appears that this only occurs when virtual scrolling is enabled, and the table has a shown header. The scroll gets offset to a min of what the header height is. Happens on Chrome, Edge, and Safari, but not FireFox.

### Flavour

### Areas

Components (quasar), Plugins (quasar)

### Platforms/Browsers

Chrome, Safari, Microsoft Edge

### Quasar info output

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

evnchn:

- [x] Bugfix

- [x] No

- [x] Any necessary documentation has been added or updated — none needed; no public API surface changes

---

**TL;DR — QMenu and QDialog restore focus on hide with a bare `target.focus()`, and when that target is the `tabindex="-1"` `.q-virtual-scroll__content` (a row mousedown parks focus there), the browser's default scroll-into-view fires and shifts the list by exactly one sticky-header height on every close.**

`preventScroll: true` expresses the intent that was already there — *restore focus*, not *scroll* — and is the same option both components already use when focusing **into** the overlay.

Resolves #15598 (open since 2023). Resolves #17081. They are the same defect at two call sites: the identical refocus block in QMenu and in QDialog.

```diff
   refocusTarget = null
   addFocusFn(() => {
-    if (target.isConnected) target.focus()
+    if (target.isConnected) target.focus({ preventScroll: true })
   })
```

Applied identically in `ui/src/components/dialog/QDialog.js:264` and `ui/src/components/menu/QMenu.js:308`. **The compiled CSS is byte-identical to the unpatched build** (sha1 `1484162dc6ca0a3a75799762d008661fa3268912` before and after) — this change is entirely in the two JS call sites.

## Verify it in 30 seconds

**Right-click a row of the virtual-scroll table, close the menu, and the table scrolls down 48px — exactly one sticky-header height.**

📄 **[`mre/15598.html`](https://raw.githubusercontent.com/evnchn/quasar/uc-assets/mre/15598.html)** — one self-contained file, no build, no deps. Open it in a browser or paste it into a CodePen; it loads Quasar 2.23.3 from the CDN and **self-reports a pass/fail verdict on the page**.

|  | `t-virtual` (virtual-scroll) | `t-plain` (control) |
|---|---|---|
| **unpatched 2.23.3**, chromium | ❌ scrollTop 0 → **48** | ✅ 0 → 0 |
| **patched build**, chromium | ✅ 0 → 0 | ✅ 0 → 0 |
| **unpatched 2.23.3**, firefox | ✅ 0 → 0 *(immune)* | ✅ 0 → 0 |

The non-virtual-scroll table is the passing control: it localizes the cause to the `tabindex="-1"` on `.q-virtual-scroll__content`, which is what parks focus on the TBODY that the overlay then refocuses. Firefox does not scroll-into-view on this `focus()` and is unaffected — that engine split is itself a control.

<details>

**unpatched — `delta=48`**

**patched — `delta=0`**

</details>

---

<details>

Both components capture the pre-open focus and restore it on hide:

```js
// QMenu.js:228  /  QDialog.js:196
refocusTarget = props.noRefocus ? null : document.activeElement
```

```js
// QMenu.js:305-309  /  QDialog.js:261-265
refocusTarget = null
addFocusFn(() => {
  if (target.isConnected) target.focus()
})
```

The element that ends up in `refocusTarget` is the interesting part. `QVirtualScroll` renders its content wrapper with `tabindex: -1`:

```js
// ui/src/components/virtual-scroll/use-virtual-scroll.js:800-808
h(tag, { class: 'q-virtual-scroll__content', key: 'content', ref: contentRef, tabindex: -1 }, ...)
```

That `tabindex` is load-bearing: when the rendered slice changes while focus is inside it, the component deliberately parks focus on the wrapper so focus is not lost as rows are recycled —

```js
// use-virtual-scroll.js:653-655
function onBlurRefocusFn() {
  contentRef.value?.focus()
}
```

So after a row mousedown in a virtual-scrolled QTable, `document.activeElement` is `TBODY.q-virtual-scroll__content` — measured, not assumed: the harness reports `aeBefore = TBODY.q-virtual-scroll__content` in every cell of the matrix below.

Open an overlay, close it, and the restore calls `.focus()` on that `TBODY`. Per CSSOM View, `focus()` performs scroll-into-view with `nearest` alignment unless `preventScroll` is set. The `TBODY` is **taller than the scrollport** (200 rows in a 300px-high `.q-table__middle`), and its top edge sits exactly one sticky-header height below the scrollport top. `nearest` therefore aligns the target's top edge to the scrollport's top edge, and `scrollTop` moves `0 → 48` — where 48 is the measured `thead` height (`theadH: 48` in the harness output). Every close nudges the list down by one header row.

**The keyboard-close branch lands in the same place.** On a key-originated hide, the target is re-resolved:

```js
(evt?.type.indexOf('key') === 0 ? refocusTarget.closest('[tabindex]:not([tabindex^="-"])') : void 0) || refocusTarget
```

In a QTable the virtual scroller receives a `scrollTarget`, so `QVirtualScroll.js:113-115` gives the root **no** `tabindex` at all — and the harness confirms `closest('[tabindex]:not([tabindex^="-"])')` returns `null`, falling back to `|| refocusTarget`, i.e. the same `TBODY`. Both branches of that expression reach the one changed line, which is why one line per component covers both.

Firefox never exhibited the bug (`0 → 0` unpatched) — it does not scroll for this particular oversized-target geometry. Chromium and WebKit do.

</details>

<details>
<summary><b>Validation — cross-engine A/B, 10 shapes, real rebuilds</b></summary>

Quasar built from this worktree with `node ui/build/script.build.js`, unpatched and patched, served from two pinned local HTTP servers whose payloads were **verified by sha1 over the wire** before each run. No CDN, no page-level shim. Driven with Playwright.

| build | `quasar.umd.prod.js` | `quasar.prod.css` |
|---|---|---|
| unpatched (`dev`) | `c9a036de9a85493b2219fb2b9fa1a37040d6a883` | `1484162dc6ca0a3a75799762d008661fa3268912` |
| this PR | `6ebc7ffa1100cf240d4962f6781813b7c5bab9dd` | `1484162dc6ca0a3a75799762d008661fa3268912` ← identical |

The fix was confirmed present in the **built minified artifact**, not trusted from source: `grep` finds `isConnected&&t.focus({preventScroll:!0})` twice, and the count of `preventScroll` in the bundle goes 10 → 12 — exactly +2, one per call site.

Engines: **chromium 151.0.7922.34, firefox 153.0, webkit 26.5.** All three launched and used; none dropped.

**The bar here is not "still works" but "geometry byte-identical to unpatched", except where unpatched was broken.**

| shape | what it is | chromium | firefox | webkit |
|---|---|---|---|---|
| **A** | the reported repro: QTable `virtual-scroll`, real row click, QMenu open/close | **0→48 BROKEN → 0→0** | 0→0 identical | **0→48 BROKEN → 0→0** |
| **A** | same, QDialog | **0→48 BROKEN → 0→0** | 0→0 identical | **0→48 BROKEN → 0→0** |
| **K** | same as A but closed with a **real `Escape` keypress** (QMenu) | **0→48 BROKEN → 0→0** | 0→0 identical | **0→48 BROKEN → 0→0** |
| **K** | same, QDialog | **0→48 BROKEN → 0→0** | 0→0 identical | **0→48 BROKEN → 0→0** |
| **B** | QVirtualScroll inside a scrolled container, content fully visible | identical | identical | identical |
| **C** | QVirtualScroll with `scroll-target="body"` (page-level) | identical | identical | identical |
| **D** | horizontal virtual scroll in a scroll container | identical | identical | identical |
| **E** | QTable with `virtual-scroll-target="#outer"`, real row click + real QMenu | identical | identical | identical |
| **F** | page-level `scroll-target="body"`, content deterministically placed fully visible | identical | identical | identical |
| **G** | plain `scrollIntoView()` — default, `block:'center'`, `block:'nearest'` | identical | identical | identical |
| **H** | control: a plain `<button>` in a scrolled container, real QMenu open/close | identical | identical | identical |

"identical" means every field of the probe — `scrollTop` / `scrollY`, the element's rect `top`/`bottom`, `fullyVisible`, and `document.activeElement` — matched the unpatched build exactly.

Raw broken→fixed values, for the record (`scrollTop` of `.q-table__middle`, `thead` height 48):

```
chromium  A menu    0 -> 48  BROKEN  ->  0 -> 0  OK
chromium  A dialog  0 -> 48  BROKEN  ->  0 -> 0  OK
webkit    A menu    0 -> 48  BROKEN  ->  0 -> 0  OK
webkit    A dialog  0 -> 48  BROKEN  ->  0 -> 0  OK
chromium  K esc/menu    0 -> 48  BROKEN  ->  0 -> 0  OK
chromium  K esc/dialog  0 -> 48  BROKEN  ->  0 -> 0  OK
webkit    K esc/menu    0 -> 48  BROKEN  ->  0 -> 0  OK
webkit    K esc/dialog  0 -> 48  BROKEN  ->  0 -> 0  OK
```

**Focus is still restored.** In every patched cell, `document.activeElement` after the close is still `TBODY.q-virtual-scroll__content` (shapes A/K) or the button (shape H) — the fix suppresses the scroll, not the refocus.

**Determinism.** Shapes A–H were run twice end-to-end against both builds; all four result sets are byte-identical between run 1 and run 2 (engine version strings excluded).

| check | result |
|---|---|
| `ui/testing` → `vitest run` (full suite) | **104 files / 1180 tests passed**, 0 skipped |
| `node ./testing/specs/script.js --target components/dialog/QDialog` | `✅ src/components/dialog/QDialog.test.js` |
| `node ./testing/specs/script.js --ci` (all specs) | exit 0 |
| `oxfmt --check` on the 3 changed files | "All matched files use the correct format." |
| `oxlint` on `components/dialog/` + `components/menu/` | exit 0, no findings |

</details>

<details>
<summary><b>Regression test — seen to fail without the fix</b></summary>

Added to the existing `[(prop)no-refocus]` block in `ui/src/components/dialog/QDialog.test.js`. The suite is jsdom, which implements no layout and therefore cannot observe scrolling — so the test asserts the **call contract that produces the browser behaviour**, plus the invariant that must survive it (focus is still restored):

```js
test('refocuses without scrolling the target into view', async () => {
  const el = createFocusEl()

  el.focus()

  const focusSpy = vi.spyOn(el, 'focus')

  wrapper = mount(QDialog)

  wrapper.findComponent({ name: 'QDialog' }).vm.show()

  await flushPromises()
  await vi.runAllTimers()

  wrapper.findComponent({ name: 'QDialog' }).vm.hide()

  await flushPromises()
  await vi.runAllTimers()

  expect(document.activeElement).toBe(el)
  expect(focusSpy).toHaveBeenCalledWith({ preventScroll: true })
})
```

**With the fix reverted** (`QDialog.js:264` restored to `target.focus()`), verbatim:

```
Received:

  1st focus call:

- [
-   {
-     "preventScroll": true,
-   },
- ]
+ []

Number of calls: 1

 ❯ ../src/components/dialog/QDialog.test.js:596:26
    594|
    595|         expect(document.activeElement).toBe(el)
    596|         expect(focusSpy).toHaveBeenCalledWith({ preventScroll: true })
       |                          ^
    597|       })
    598|     })

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯

 Test Files  1 failed (1)
      Tests  1 failed | 36 skipped (37)
```

**With the fix in place**, verbatim (absolute paths trimmed):

```
 RUN  v4.1.10 ui/testing

 Test Files  1 passed (1)
      Tests  1 passed | 36 skipped (37)
   Duration  944ms (transform 286ms, setup 532ms, import 73ms, tests 48ms, environment 226ms)
```

Note the failure is on the **arguments**, not on the existence of a call — `focus()` is still called either way, so the test is pinned to the specific behaviour this PR changes rather than to the refocus feature in general.

</details>

<details>
<summary><b>Alternative considered (built and measured), and the behavioural cost of this one</b></summary>

**Rejected: `scroll-margin-top: 100vh` on `.q-virtual-scroll__content`.**

This was my first attempt, and it does fix the reported repro. I built it (css sha1 `cf58f1884f0d20a068ac252f39bde8694e741f1b`, js byte-identical to unpatched) and ran the same matrix against it. It is **worse than the bug**, because `scroll-margin` is not scoped to the overlay-refocus path — it rewrites the scroll-into-view contract for a public class name, for every caller:

| shape | unpatched | with `scroll-margin-top: 100vh` |
|---|---|---|
| A — the reported repro | 0 → 48 (broken) | 0 → 0 (fixed) |
| E — QTable `virtual-scroll-target` on an outer container, chromium/webkit | 400 → 400 | **400 → 246** |
| E — same, firefox | 400 → 400 | **400 → 123** ← firefox never had the bug |
| F — page-level `scroll-target="body"`, chromium/webkit | scrollY 1200 → 1200, fully visible | **1200 → 896** |
| F — same, firefox | scrollY 1200 → 1200, fully visible | **1200 → 848, no longer fully visible** |
| G — `scrollIntoView()` | top 0, fully visible | **top 700, 100% below the fold** |
| G — `scrollIntoView({block:'center'})` | top 302, fully visible | **top 652, not visible** |
| G — `scrollIntoView({block:'nearest'})` | top 604, fully visible | **top 700, not visible** |

All three engines. It trades a 48px jump in one geometry for 123–352px jumps in two other documented ones, in **three** engines instead of two, and it breaks `scrollIntoView()` on that element outright. `virtual-scroll-target` is a public QTable prop and `scroll-target` is public on QVirtualScroll, so both of those geometries are supported configurations, not contrivances. Not shippable.

**Rejected: fixing `onBlurRefocusFn` in `use-virtual-scroll.js:653`.** That is a third bare `.focus()`, and it may well deserve `preventScroll` too — but it is a *different* path (focus preservation during slice recycling, not overlay restore), it is not what #15598/17081 report, and changing it here would widen the diff beyond the measured defect. Left alone deliberately.

**Rejected: scoping `preventScroll` to virtual-scroll targets only** (e.g. sniffing for `.q-virtual-scroll__content` in the overlay). It would bake one known repro into QMenu/QDialog, invert the layering (an overlay knowing about a specific component's internals), and leave the identical restore-scroll hazard in place for any other oversized or awkwardly-positioned focus target.

<hr>

<b>The behavioural cost, measured</b>

`preventScroll: true` is not free, and the honest statement of what it costs is: **if the refocus target has genuinely scrolled out of view while the overlay was open, focus is still restored to it but the browser no longer scrolls it back.**

Measured (shapes I/J): a plain `<button>` inside a scroll container, focused, overlay opened, then the container programmatically scrolled to `scrollTop 1600` so the button is well outside the scrollport, then the overlay closed:

| | unpatched | this PR |
|---|---|---|
| QMenu close | 1600 → **471**, button visible again | 1600 → **1600**, button not visible |
| QDialog close | 1600 → **471**, button visible again | 1600 → **1600**, button not visible |
| `document.activeElement` after close | the button | the button (unchanged) |

Identical in all three engines. Two things bound this:

- **It requires the target to leave the viewport while the overlay is open.** In the control case where the target stays visible (shape H), the two builds are identical — `450 → 450`, because the browser's `nearest` alignment already no-ops for a fully-visible target. So this is not a general "focus rings go off-screen" regression; it is specifically the scrolled-away case.
- **I reached it with a programmatic scroll.** QDialog locks body scroll while open (`usePreventScroll`), so for a page-level target this is hard to reach by ordinary interaction; an *inner* scroll container is not locked, and QMenu does not lock scroll at all, so it is more reachable there. I did not establish how reachable it is by real user gestures — that is in the "not verified" list below rather than being claimed either way.

Against that, the bug being fixed fires on the completely ordinary path — click a row, open a menu, close it — and drifts the list on **every** close. Focus restoration itself is untouched in every measured case, which is the part WCAG 2.4.3 / the APG dialog pattern actually require. And this is the same trade the components already make elsewhere: `QDialog.js:289`, `:305` and `:312`, and `QMenu.js:222` all already call `focus({ preventScroll: true })`. This PR makes the hide path consistent with the rest of both files.

</details>

<details>
<summary><b>Not verified — stated plainly</b></summary>

- **Not tested on Cordova or Electron.** No such app set up here; the change is a one-option difference on a standard DOM call and the three-engine results are the closest proxy I can offer.
- **No unit test for the QMenu call site** — only QDialog has one, for the reason given in the second-opinion fold. QMenu is covered by the browser matrix (shapes A, K, E, H), not by the unit suite.
- **jsdom cannot prove the actual fix.** The unit test asserts the `preventScroll` argument, not a scroll position, because jsdom has no layout. The browser matrix is the real evidence for the behaviour; the unit test is the guard against silent removal.
- **How reachable the shape I/J cost is by real user gestures was not established.** I produced it with a programmatic `scrollTop` write. I did not test whether a real wheel/touch gesture on an inner container while a QMenu is open produces the same state, nor whether QMenu's reposition/close-on-scroll behaviour pre-empts it.
- **QSelect was not tested.** Its dropdown is a QMenu containing a QVirtualScroll, so it is plausibly the third reported surface, but I did not build a QSelect repro and make no claim about it.
- **SSR/hydration, RTL, and mobile/touch were not specifically exercised** beyond the full unit suite. RTL is unlikely to matter for a vertical-axis scroll, but I did not measure it.
- **Only three engines, current versions** — chromium 151.0.7922.34, firefox 153.0, webkit 26.5, all via Playwright rather than real installed browsers. No older-browser sweep. `focus({ preventScroll })` is unsupported in some older engines, where the option is ignored and behaviour reverts to today's — degradation, not breakage — but I did not empirically test any such engine.
- **The 48px figure is this fixture's `thead` height**, measured at `theadH: 48`, not a universal constant. The general statement is "one sticky-header height".
- **`onBlurRefocusFn` (`use-virtual-scroll.js:653`) still calls a bare `.focus()`** and is deliberately untouched. I did not measure whether it produces a scroll of its own during slice recycling; if it does, that is a separate issue from the one this PR closes.
- **Pre-commit hook bypassed.** `husky` runs a pnpm deps-status check that aborts in this environment (the worktree's `node_modules` are symlinked, and pnpm refuses to purge without a TTY). The gates that hook would have run were run manually instead and are reported in the validation fold.

</details>

<sub>Investigated and drafted by Claude Code on evnchn's behalf; all measurements above were run rather than reasoned, against locally rebuilt artifacts rather than a CDN.</sub>

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved focus restoration when closing dialogs and menus by preventing the page from scrolling unexpectedly.
  * Preserved the previously focused element while keeping its scroll position stable.

* **Tests**
  * Added regression coverage to verify focus returns correctly without triggering unwanted scrolling.

### DIFF
--- a/ui/src/components/dialog/QDialog.js
+++ b/ui/src/components/dialog/QDialog.js
@@ -261,7 +261,7 @@ export default createComponent({
 
         refocusTarget = null
         addFocusFn(() => {
-          if (target.isConnected) target.focus()
+          if (target.isConnected) target.focus({ preventScroll: true })
         })
       }
 
--- a/ui/src/components/dialog/QDialog.test.js
+++ b/ui/src/components/dialog/QDialog.test.js
@@ -572,6 +572,29 @@ describe('[QDialog API]', () => {
 
         expect(document.activeElement).not.toBe(el)
       })
+
+      test('refocuses without scrolling the target into view', async () => {
+        const el = createFocusEl()
+
+        el.focus()
+
+        const focusSpy = vi.spyOn(el, 'focus')
+
+        wrapper = mount(QDialog)
+
+        wrapper.findComponent({ name: 'QDialog' }).vm.show()
+
+        await flushPromises()
+        await vi.runAllTimers()
+
+        wrapper.findComponent({ name: 'QDialog' }).vm.hide()
+
+        await flushPromises()
+        await vi.runAllTimers()
+
+        expect(document.activeElement).toBe(el)
+        expect(focusSpy).toHaveBeenCalledWith({ preventScroll: true })
+      })
     })
 
     describe('[(prop)no-focus]', () => {
--- a/ui/src/components/menu/QMenu.js
+++ b/ui/src/components/menu/QMenu.js
@@ -305,7 +305,7 @@ export default createComponent({
 
         refocusTarget = null
         addFocusFn(() => {
-          if (target.isConnected) target.focus()
+          if (target.isConnected) target.focus({ preventScroll: true })
         })
       }
 


rstoenescu:
Thanks!

PR #18481, opened 2026-07-30
fix(date): escape custom locale names

hawkeye64:
## What changed

- Escape every custom day and month name before inserting it into the date parser regular expression.
- Reuse the same escaping helper for literal mask text.
- Add behavioral coverage for regular-expression metacharacters in `days`, `daysShort`, `months`, and `monthsShort`.
- Document that custom locale names are matched literally.

## Why

The localized date parser previously joined custom locale names directly into regular-expression alternations. A locale name containing characters such as `.`, `*`, `+`, parentheses, brackets, `|`, or `\` could therefore change the parser expression instead of matching its visible text. Capturing groups in a custom name could also shift the indices used to extract later date fields.

Locale-provided names are data, not regular-expression syntax. Escaping each entry preserves the existing parser structure, cache behavior, and matching semantics for normal language packs while making custom locale values literal and predictable.

## Public API impact

No signature or type changes are required. Existing locale names without regular-expression metacharacters behave as before. Custom names containing those characters now parse as their literal text.

## Validation

- `cd ui && pnpm build`
- `cd ui && pnpm test:specs --target utils/date/date`
- `cd ui && pnpm --filter quasar-ui-test test ../src/utils/date/date.test.js` — 98 tests passed
- `pnpm test` — 135 UI files / 1,352 UI tests, 10 Vite usage tests, and 75 Vite runtime tests passed
- `pnpm lint:check`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved date parsing for custom locale day and month names containing regular-expression characters.
  * Ensured these names are interpreted as literal text, enabling reliable date extraction.

* **Documentation**
  * Clarified how custom locale names are matched during date parsing.

### DIFF
--- a/docs/src/pages/quasar-utils/date-utils.md
+++ b/docs/src/pages/quasar-utils/date-utils.md
@@ -449,6 +449,9 @@ const date = date.extractDate('21/03/1985', 'DD/MM/YYYY')
 
 With optional custom locale:
 
+Custom day and month names are matched literally, including any regular
+expression metacharacters they contain.
+
 ```js
 import { date } from 'quasar'
 
--- a/ui/src/utils/date/date.js
+++ b/ui/src/utils/date/date.js
@@ -194,6 +194,7 @@ function getRegexData(mask, dateLocale) {
         if (match[0] === '[') {
           match = match.slice(1, -1)
         }
+
         return escapeRegex(match)
       }
     }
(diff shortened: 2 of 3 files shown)


hawkeye64:
@coderabbitai review

PR #18482, opened 2026-07-30
fix(lang): encode SSR language attributes

hawkeye64:
## What changed

- Quote the SSR-generated `lang` and `dir` attributes.
- HTML-attribute-encode custom language-pack `isoName` values before adding them to SSR metadata.
- Test the actual built server bundle with markup characters and repeated language changes.

## Why

The SSR Lang plugin interpolated `isoName` into an unquoted raw attribute string. Built-in packs use trusted identifiers, but a custom language pack containing spaces, quotes, or markup could change the server-rendered `<html>` attributes.

The client path already uses `setAttribute()`. This change gives SSR equivalent output safety while preserving the supplied identifier and existing language-pack compatibility.

## Public API and SSR impact

No API, type, or configuration signatures change. Normal language packs produce semantically identical quoted attributes. Custom values are preserved as text but HTML-encoded, so they cannot escape the `lang` attribute. Repeated SSR language changes continue replacing the prior Quasar-generated attributes, including RTL direction changes.

A strict language-tag validator was intentionally not added because it rejects existing Quasar identifiers such as `sr-CYR` and `kur-CKB` and could break valid custom/private tags. Output encoding fully addresses the HTML injection sink without introducing that compatibility risk.

## Validation

- `cd ui && pnpm build`
- `cd ui && pnpm test:specs --target plugins/lang/Lang`
- `cd ui && pnpm --filter quasar-ui-test test ../src/plugins/lang/Lang.test.js` — 7 tests passed
- `pnpm test` — 135 UI files / 1,349 UI tests, 10 Vite usage tests, and 75 Vite runtime tests passed
- `pnpm lint:check`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server-rendered language metadata by safely escaping special characters in HTML attributes.
  * Prevented malformed or unsafe language attributes from being generated.

* **Tests**
  * Added coverage for malicious language values and verified correct handling of right-to-left language metadata.

### DIFF
--- a/ui/src/plugins/lang/Lang.js
+++ b/ui/src/plugins/lang/Lang.js
@@ -3,6 +3,18 @@ import { createReactivePlugin } from '../../utils/private.create/create.js'
 // oxlint-disable-next-line unicorn/prefer-export-from
 import defaultLang from '../../../lang/en-US.js'
 
+const htmlChars = {
+  '&': '&amp;',
+  '<': '&lt;',
+  '>': '&gt;',
+  '"': '&quot;',
+  "'": '&#39;'
+}
+
+function encodeHtmlAttribute(value) {
+  return String(value).replaceAll(/[&<>"']/g, char => htmlChars[char])
+}
+
 function getLocale() {
   if (__QUASAR_SSR_SERVER__) return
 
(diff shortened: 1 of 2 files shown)


rstoenescu:
Unnecessary overhead. There shouldn't be any lang pack iso name that wouldn't pass the test anyway.

Issue #18504, opened 2026-08-06
Electron mode fails with UNRESOLVED_ENTRY on Quasar


Alexssmusica:
### What happened?

<img width="790" height="425" alt="Image" src="https://github.com/user-attachments/assets/ac4320b7-4ef1-40f6-bf30-1f582751a77d" />

### What did you expect to happen?

# Electron mode fails with `UNRESOLVED_ENTRY` on Quasar v2.24.0 + @quasar/app-vite v3.4.1

## Description

After upgrading to **Quasar v2.24.0** and **@quasar/app-vite v3.4.1**, I can no longer start the application in **Electron mode**.

Running:

```bash
quasar dev -m electron
```

automatically creates the `src-electron` folder, installs the required dependencies, and starts the build process. However, the build fails with the following error:

```text
[UNRESOLVED_ENTRY] Cannot resolve entry module .quasar/dev-electron/electron/q_entry_main.js.
```

## Full output

```text
> teste100.0.1 dev
> quasar dev -m electron

Installing Electron Mode...

✔ Created /src-electron
✔ Executed "npm install" in /src-electron
✔ Electron support was added

App • quasar.config.ts (no env prefix; no dotenv files)...
App • Using code prefix: 'QCL_'; no dotenv files
App • WAIT Rollup • Compiling Electron Main...
App • WAIT Rollup • Compiling Electron Preload (electron-preload)...

Error: Build failed with 1 error:

[UNRESOLVED_ENTRY]
Cannot resolve entry module .quasar/dev-electron/electron/q_entry_main.js.

    at agge:...
```

## Environment

| Component | Version |
|----------|---------|
| Quasar | 2.24.0 |
| @quasar/app-vite | 3.4.1 |
| Electron Mode | Enabled |
| Bundler | Vite |
| Package Manager | npm |

## Steps to reproduce

1. Create a new Quasar project.
2. Install:
   - `quasar@2.24.0`
   - `@quasar/app-vite@3.4.1`
3. Enable Electron mode:
   ```bash
   quasar mode add electron
   ```
4. Start the application:
   ```bash
   quasar dev -m electron
   ```

## Expected behavior

The Electron entry files should be generated inside:

```text
.quasar/dev-electron/electron/
```

and the Electron application should start normally.

## Actual behavior

The build stops during the **Electron Main** compilation because Rollup cannot resolve:

```text
.quasar/dev-electron/electron/q_entry_main.js
```

[...]

PR #18505, opened 2026-08-06
fix(app-vite): deadlock in dev with circular dependencies

Laruxo:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Since https://github.com/quasarframework/quasar/pull/18470 my macOS dev build has a deadlock. The virtual entry that adds `dock.setIcon` changes the app's main file to `await import()`. In turn, Rolldown defers the entire target graph into async module initializers (`__esmMin(async () => …)`). Those memoize the in-flight promise, so a circular import self-awaits and deadlocks permanently. In other words, the main process stays alive with an idle event loop, no window, no error, and no code from the app ever runs.

Circular imports are valid ESM and never deadlock natively: evaluation is synchronous depth-first and simply does not re-enter a module that is already evaluating (and my code worked fine before updating to rolldown). Prod builds are unaffected because they use a static entry.

One alternative would be to drop `beforeImportCode` and instead edit [electron-devserver.js](https://github.com/quasarframework/quasar/blob/dev/app-vite/lib/modes/electron/electron-devserver.js#L211-L212) to write a small bootstrap file that sets the icon. IMO this is a better solution, but requires more changes. Let me know what you prefer and I can make the change.

To reproduce the issue, add 2 files that import each other and run dev server.

Note: this was debugged, and parts of the fix were generated by Claude.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of virtual entry points to ensure bootstrap code executes reliably before the target module.
  * Fixed special-character handling in virtual entry identifiers.
  * Preserved required bootstrap code during module optimization.

### DIFF
--- a/app-vite/lib/plugins/rolldown.virtual-entry.js
+++ b/app-vite/lib/plugins/rolldown.virtual-entry.js
@@ -1,5 +1,7 @@
 import path from 'node:path'
 
+const escapeRE = str => str.replaceAll(/[.*+?^${}()|[\]\\]/g, String.raw`\$&`)
+
 export function quasarRolldownVirtualEntry({
   inputFile,
   targetFile,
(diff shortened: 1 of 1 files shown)


Laruxo:
Another thought - I also need to override the app name in dev, so prod and dev app data don't mix. I assume some people prefer to share this data, but it does not work for my app.

Today `setName()` only works if nothing has called `app.getPath()` or `requestSingleInstanceLock()` yet, which happens before the first line of `electron-main.js` runs. Meanwhile, this virtual entry is a nice place for such customization. Exposing it as an optional `src-electron/electron-dev-bootstrap.js` file (loaded before the main entry) would cover the app name, icon, `setPath()`, command-line switches, and anything else in one transparent place.

One constraint: the file has to be synchronous. Otherwise, the entry becomes
an async module, and causes the deadlock again.

rstoenescu:
Hi @Laruxo! Excellent debugging work — your diagnosis was spot on. I reproduced it against the real dev pipeline: with the await import() entry, any await occurring in a cycle before the back-edge turns the __esmMin initializers async with mutual await init_*() edges, and the memoized in-flight promise self-awaits forever. Your static-bootstrap-module approach eliminates the wrapper machinery entirely — compiled output has no __esmMin at all, so circular imports keep native ESM semantics, while dock.setIcon still runs first.

I've landed it in a6a270c6c with you credited as co-author, so I'm closing this PR as superseded rather than merging it directly. The reason: dev moved under you today — #18504 (Electron dev broken on Windows) required rewriting the same filter code your diff was based on. Rolldown matches native filter.id patterns against forward-slash-normalized ids, so the pattern must be built separator-agnostic — and your new bootstrap id embeds the OS-native input path, so it needed the same treatment. The landed commit combines your fix with that handling plus the sibling unit-test updates. Both fixes will ship together in the next patch release.

Your src-electron/electron-dev-bootstrap.js idea is a good one — a single transparent place for setName(), setPath(), command-line switches and the like, before anything in electron-main runs. Could you open a feature request for it so it gets tracked properly? Worth noting in it: the file must stay free of top-level await, otherwise it reintroduces the async initializers this PR just removed.

Thanks again for the quality report and the fix!

PR #18515, opened 2026-08-14
fix(QInput): allow cancelling mask keyboard actions

### LINKED ISSUE #18514: QInput mask: onMaskedKeydown never checks e.defaultPrevented, unlike QSelect after #18507
**Describe the bug**

I just read #18507 ("fix(QSelect): allow cancelling keyboard actions", merged today), which changed `onTargetKeydown` in `QSelect.js` from:

```js
function onTargetKeydown(e) {
  emit('keydown', e)
  if (shouldIgnoreKey(e)) return
  ...
}
```

to:

```js
function onTargetKeydown(e) {
  emit('keydown', e)
  if (shouldIgnoreKey(e) || e.defaultPrevented) return
  ...
}
```

so that a consumer can now opt out of QSelect's built-in keyboard handling with something like `@keydown.enter.prevent`.

I checked `onMaskedKeydown` in `ui/src/components/input/use-mask.js` (the keydown handler QInput wires up when a `mask` prop is set) and it has the exact same shape but wasn't touched by #18507:

```js
function onMaskedKeydown(e) {
  emit('keydown', e)

  if (
    shouldIgnoreKey(e) ||
    e.altKey // let browser handle these
  ) {
    return
  }
  ...
}
```

`e.defaultPrevented` is never checked here, so a masked QInput still runs its own internal keyboard handling (cursor movement across mask segments on Left/Right including its own `e.preventDefault()`, and the Backspace/Delete boundary logic) even when the consumer's own `@keydown` handler calls `preventDefault()`. Observing the event was fixed a while ago (#11120, closed in 2022, added the missing `emit('keydown', e)` call), but there's still no way to actually cancel the mask's internal handling the way #18507 just made possible for QSelect.

**Reproduction**

```html
<q-input v-model="text" mask="##/##/####" @keydown.left.prevent="onLeft" />
```

`onLeft` fires (confirmed against current source, `emit('keydown', e)` runs first), but the cursor still jumps according to the mask's own Left-arrow logic, because `onMaskedKeydown` never checks whether the event was prevented before calling `moveCursor...`/`e.preventDefault()` itself.

**Expected behavior**

Same as the new QSelect behavior: if the consumer's `@keydown` handler calls `preventDefault()`, `onMaskedKeydown` should stop before running its own cursor/backspace/delete logic, mirroring the guard `#18507` just added:

```js
if (shouldIgnoreKey(e) || e.altKey || e.defaultPrevented) return
```

**Platform**

Checked against the current `dev` branch source (`ui/src/components/input/use-mask.js`), same day #18507 merged.

lazerg:
`onMaskedKeydown` emits `keydown` first, then runs its own cursor logic, but it never checks whether the consumer already prevented the event. On a masked QInput a handler like `@keydown.left.prevent` fires, and the mask still moves the caret across the mask literals afterwards.

This adds `e.defaultPrevented` to the early return, the same guard #18507 added to QSelect's `onTargetKeydown`. Behavior is unchanged unless the event is explicitly prevented.

Validation: `cd ui && pnpm test:unit` (260 files, 5425 tests passed), `pnpm test:specs --target components/input/use-mask`, `pnpm lint`, `git diff --check`.

Fixes #18514

## Summary by CodeRabbit

- **Bug Fixes**
  - Masked inputs now respect keyboard events that have already been canceled, preventing unintended mask processing.
  - Improved cursor behavior when navigation keys are prevented, including movement across mask literals.

- **Documentation**
  - Added guidance and an example explaining how canceling keydown events affects input masks.

### DIFF
--- a/docs/src/pages/vue-components/input.md
+++ b/docs/src/pages/vue-components/input.md
@@ -190,6 +190,8 @@ The `reverse-fill-mask` is useful if you want to force the user to fill the mask
 
 <DocExample title="Filling the mask in reverse" file="MaskFillReverse" />
 
+You can prevent the mask's own handling of a key by preventing its `keydown` event; for example, `@keydown.left.prevent` keeps <kbd>ARROW LEFT</kbd> from moving the cursor over the mask literals.
+
 ### Custom mask tokens <q-badge label="v2.18.4+" />
 
 You can also define custom mask tokens on top of the default ones or even override some/all of the [default ones](https://github.com/quasarframework/quasar/blob/dev/ui/src/components/input/use-mask.js#L15).
--- a/ui/src/components/input/use-mask.js
+++ b/ui/src/components/input/use-mask.js
@@ -575,6 +575,7 @@ export default function useMask(props, emit, emitValue, inputRef) {
 
     if (
       shouldIgnoreKey(e) ||
+      e.defaultPrevented ||
       e.altKey // let browser handle these
     ) {
       return
--- a/ui/src/components/input/use-mask.test.js
+++ b/ui/src/components/input/use-mask.test.js
@@ -371,7 +371,8 @@ describe('[useMask API]', () => {
 
       test.each([
         ['ALT is held', { keyCode: 39, altKey: true }],
-        ['the key is a modifier', { keyCode: 16 }]
+        ['the key is a modifier', { keyCode: 16 }],
+        ['the event was prevented', { keyCode: 39, defaultPrevented: true }]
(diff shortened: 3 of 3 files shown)


rstoenescu:
Thanks for contributing! This got me thinking and deserves a follow-up.

lazerg:
Glad it helped. Would be great if someone looks into the same pattern for the other components too.

rstoenescu:
> Glad it helped. Would be great if someone looks into the same pattern for the other components too.

Already did :) v2.25 is gonna be awesome!

PR #18518, opened 2026-08-19
fix(app-vite): resolve BEX bridge send() with the responder's payload

### LINKED ISSUE #18517: [BEX] bridge.send() always resolves undefined — the declared Promise<BexEventResponse<T>> value is discarded
### What happened?

`BexBridge.send()` is typed and documented to resolve the responder's return value, but it always
resolves `undefined`.

The implementation ends by **awaiting** the response promise as the last statement of an `async`
function, so the resolved value is discarded rather than returned:

`exports/bex/private/bex-bridge.js` (v3.7.0, and identical in 3.5.0):

```js
    const { promise, resolve, reject } = Promise.withResolvers()
    this.messageMap[id] = {
      portName: to,
      resolve: responsePayload => {
        delete this.messageMap[id]
        resolve(responsePayload)      // the payload arrives here…
      },
      reject: err => { /* … */ }
    }

    await promise                     // …and is dropped here
  }
```

This contradicts the package's own public contract. `types/bex/bridge.d.ts` declares:

```ts
send<T extends BexEventName>(
  options: { event: T; to: PortName } & BexPayload<T>
): Promise<BexEventResponse<T>>;
```

and the JSDoc immediately above it documents the value as usable:

```js
const result = await bridge.send({
  event: 'sum',
  to: 'app',
  payload: { a: 1, b: 2 }
});
console.log(result); // 3
```

Because the declared return type is `Promise<BexEventResponse<T>>`, TypeScript reports no error at
the call site — consumers get a correctly-typed value that is `undefined` at runtime, so the
failure surfaces later as a `TypeError` when a field is read off the response.

`await bridge.send({ event, to, payload })` should resolve the value returned by the corresponding
`bridge.on(event, …)` handler, as the type declaration and the JSDoc example both state.

### Reproduction URL

https://github.com/drmancini/quasar-bex-send-undefined-repro

### How to reproduce?

1. Create a BEX project (`quasar create` → BEX mode) with `@quasar/app-vite@3.7.0`.
2. In the background script, register a responder that returns a value:

```js
// src-bex/background.js
export default bexBackground(bridge => {
  bridge.on('sum', ({ payload }) => payload.a + payload.b)
})
```

3. From the app (or a content script), send the event and read the result:

```js
const result = await bridge.send({ event: 'sum', to: 'background', payload: { a: 1, b: 2 } })
console.log(result) // expected: 3 — actual: undefined
```

The responder runs and its value reaches `resolve(responsePayload)` inside `send()`; only the
return of `send()` itself is missing.

**One-line fix** — change the final statement of `send()` from `await promise` to `return promise`
(`exports/bex/private/bex-bridge.js`). Verified in a production BEX extension: with that change the
documented example behaves as documented, and a full end-to-end suite covering
app↔background↔content-script messaging passes.

### Flavour

### Areas

BEX Mode

### Platforms/Browsers

Chrome

### Quasar info output

```Shell
Operating System - Linux(6.8.0-137-generic) - linux/x64
Node.js - 24.15.0

Global packages
  NPM - 11.12.1

Important local packages
  quasar - 2.24.0
  @quasar/app-vite - 3.7.0
  @quasar/extras - 2.0.3
  vue - 3.5.41
  vue-router - 5.2.0
  typescript - 6.0.3
  vite - 8.2.1
  rolldown - 1.2.4
```

### Relevant log output

```Shell

```

### Additional context

Present in both `@quasar/app-vite@3.5.0` and the current `3.7.0`; I have not checked earlier
releases. We have been carrying it as a `patch-package` patch and would happily drop the patch.

Possibly related but distinct: #16081 (closed) described a `send()` promise that **never settles**
when dispatched from within another handler. This one always settles — it just settles with
`undefined`.

lazerg:
`BexBridge.send()` ended with `await promise` as the last statement of an `async` method, so the responder's payload was awaited and then thrown away. Every call resolved `undefined`, even though `types/bex/bridge.d.ts` declares `Promise<BexEventResponse<T>>` and the docs show the returned value being used. Since the declared type is correct, TypeScript flags nothing and the failure only shows up later as a `TypeError` on the response.

Returning the promise instead of awaiting it makes the resolved value reach the caller. No other behavior changes: rejections still propagate the same way.

Verified by driving two `BexBridge` instances (app + background) over a stubbed `chrome.runtime` port pair: `bridge.on('sum', ({ payload }) => payload.a + payload.b)` gives `undefined` before the change and `3` after. Also ran `pnpm test:unit` in `/app-vite` (70 files, 552 tests, all passing) and `pnpm lint` at the root, which only reports the pre-existing missing `.quasar/tsconfig.json` errors.

Fixes #18517

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bridge message handling by returning response promises directly, enabling callers to receive results more reliably.

### DIFF
--- a/app-vite/exports/bex/private/bex-bridge.js
+++ b/app-vite/exports/bex/private/bex-bridge.js
@@ -339,7 +339,7 @@ export class BexBridge {
       }
     }
 
-    await promise
+    return promise
   }
 
   /**


rstoenescu:
Thanks for catching this and contributing!

PR #18526, opened 2026-08-25
fix(QDialog,QMenu,QTooltip): finish pending transition on keep-alive deactivation — fix #18201

### LINKED ISSUE #18201: QSelect with behavior="dialog" doesn't open the selection dialog after returning back to keep-alive page where it was placed
### What happened?

QSelect with behavior="dialog" doesn't open the selection dialog after returning back to keep-alive page where it was placed

Image

Pressing the back button should close the selection dialog without triggering the back navigation
-- OR --
Pressing the back button should correctly close the selection dialog

### Reproduction URL

https://codesandbox.io/p/devbox/young-tree-2z9h4p

### How to reproduce?

1. Navigate to page with the QSelect component. The page should be included in keep-alive component.
2. Click the QSelect component to open the selection dialog. The dialog will be opened.
3. Don't select any item, just press the back button. The dialog will be closed and you'll be taken back to the previous page.
4. Navigate to page with the QSelect component again.
5. Click the QSelect component. The selection dialog won't open.

### Flavour

### Areas

Components (quasar)

### Platforms/Browsers

Chrome

### Quasar info output

```Shell

```

### Relevant log output

```Shell

```

### Additional context

It seems that pressing the back button while the select dialog is open doesn't close it properly, because the empty dialog portal still exists in dom.

arbaev:
Fixes #18201

- [x] Bugfix

- [x] No

- [ ] It's been tested on a Cordova (iOS, Android) app — not applicable: the change touches no platform-specific code path (the iOS branches inside the affected timers are untouched)
- [x] Any necessary documentation has been added or updated — no docs change needed: no public API, prop, event or default changes; this restores the documented lifecycle (`hide` is emitted whenever a popup closes)

**Other information:**

### The problem

Reproduction from the issue: a page held by `<keep-alive>` renders a `QSelect` with `behavior="dialog"`. Open the selection dialog, press the browser's back button (the dialog closes and the app navigates away), then come back to the page — clicking the select no longer opens the dialog, and an empty portal `div` is left behind in `body`.

### Root cause

`QDialog.handleHide()` emits `beforeHide` synchronously but defers the end of the transition by `transitionDuration`:

```js
registerTimeout(() => {
  hidePortal(true) // done hiding, now destroy
  animating.value = false
  emit('hide', evt)
}, props.transitionDuration)
```

`useTimeout()` cancels its timer on `onDeactivated` (`ui/src/composables/use-timeout/use-timeout.js`), which is correct for the delayed work it usually guards. But a route change drives both sides of this at once:

1. the route ref changes; `RouterView` re-renders first (lower instance id) and `<keep-alive>` deactivates the page, queueing the deactivation hooks as post-render effects;
2. the `hideOnRouteChange` watcher in `QDialog` then runs `hide()` → `handleHide()`, which registers the transition timer;
3. the queued `onDeactivated` hooks flush afterwards and clear that timer.

So `hidePortal(true)` and `emit('hide')` never happen. Two consequences:

- the portal node stays in the DOM (empty, since `showing` is already `false`) and its proxy stays in `portalProxyList`;
- `QSelect` resets `dialog.value` in `onDialogHide` — the handler for that lost event — so its model stays `true` forever. Re-activating the page and clicking the control calls `showPopup()`, which sets `dialog.value = true` again: no change, no dialog.

`behavior="dialog"` is the visible victim because the menu path resets QSelect's state in `onBeforeHide` (synchronous), while the dialog path only does it in `onHide`. `QMenu` and `QTooltip` share the same deferred-tail pattern and leak the portal node the same way.

### The fix

Keep the transition's finisher at hand and run it synchronously if the component is deactivated before its timer fires:

```js
// The tail of a show/hide transition (portal teardown, the 'show'/
// 'hide' event) runs on a timer that useTimeout cancels when a
// <keep-alive> page holding this dialog gets deactivated. Keep the
// finisher at hand so deactivation can run it right away, or else
// the portal is left in the DOM and the model stays stuck (#18201).
// Should removeTimeout() if this gets removed.
function registerTransitionEnd(fn) {
  finishTransition = () => {
    finishTransition = null
    fn()
  }

  registerTimeout(finishTransition, props.transitionDuration)
}

onDeactivated(() => {
  finishTransition?.()
})
```

Both transition timers (`handleShow` and `handleHide`) go through `registerTransitionEnd`, in `QDialog`, `QMenu` and `QTooltip`.

`useTimeout()` itself is deliberately left alone: it is a public composable, and cancelling deferred work on deactivation is its correct semantics — a tooltip's open delay, for instance, must not fire for a page the user has left.

The finisher always reflects the last transition that was started, so running it on deactivation lands the portal in a state consistent with `showing`. In `QTooltip` the delayed show/hide share `useTimeout`'s single timer slot with the transition; if one of them preempts a transition timer, the leftover finisher still resolves the portal correctly (it either completes the show that is in effect, or tears down the portal of the hide that already happened).

### Scope

- `ui/src/components/dialog/QDialog.js`
- `ui/src/components/menu/QMenu.js`
- `ui/src/components/tooltip/QTooltip.js`
- one `[Generic]` test per component

No behavior changes outside of `<keep-alive>` deactivation: without a live transition the new hook is a no-op.

### Out of scope

A popup that is fully shown (no transition in flight) when its keep-alive page deactivates still stays mounted in `body`: Vue moves only the Teleport anchors on deactivation, not the teleported content. That is a separate concern (`<keep-alive>` without a route change, e.g. tabs) and is not what #18201 describes.

### Validation

- New test per component: *"finishes a pending hide when its keep-alive page deactivates"* — mounts the component inside `<KeepAlive>` behind a route, then routes away, which hides it and deactivates the page within the same tick. Each test fails on `dev` (the `hide` handler is never called, the portal node survives) and passes with the fix.
- `pnpm test` (ui): every suite green — unit (260 files), hydration (89 files), hydration:pwa, umd, build, e2e:ssr.
- `pnpm test:specs --target` for QDialog, QMenu, QTooltip: ✅.
- Manual check in a real browser (playground page with `<keep-alive>` around a `QSelect behavior="dialog"`, driven headless over CDP):
  - before the fix — after the back navigation one `div[id^="q-portal--dialog"]` is left in `body` and the restored page's select no longer opens;
  - after the fix — no portal node is left and clicking the select opens the dialog again.
- `pnpm lint` and `git diff --check` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed dialogs, menus, and tooltips remaining open or leaving portal elements behind when navigating away from cached pages.
  - Ensured pending show/hide transitions complete immediately when a cached page is deactivated.
  - Corrected model state and event handling after deactivation.

- **Tests**
  - Added regression coverage for dialogs, menus, and tooltips during cached-page navigation.

### DIFF
--- a/ui/src/components/dialog/QDialog.js
+++ b/ui/src/components/dialog/QDialog.js
@@ -4,6 +4,7 @@ import {
   getCurrentInstance,
   h,
   onBeforeUnmount,
+  onDeactivated,
   ref,
   watch
 } from 'vue'
@@ -109,7 +110,8 @@ export default /*#__PURE__*/ createComponent({
     let shakeTimeout = null,
       refocusTarget = null,
       isMaximized = false,
-      avoidAutoClose = false
+      avoidAutoClose = false,
+      finishTransition = null
 
     const hideOnRouteChange = computed(
       () => !props.persistent && !props.noRouteDismiss && !props.seamless
@@ -194,6 +196,25 @@ export default /*#__PURE__*/ createComponent({
       }
     })
 
+    // The tail of a show/hide transition (portal teardown, the 'show'/
+    // 'hide' event) runs on a timer that useTimeout cancels when a
+    // <keep-alive> page holding this dialog gets deactivated. Keep the
+    // finisher at hand so deactivation can run it right away, or else
+    // the portal is left in the DOM and the model stays stuck (#18201).
+    // Should removeTimeout() if this gets removed.
+    function registerTransitionEnd(fn) {
+      finishTransition = () => {
+        finishTransition = null
+        fn()
+      }
+
+      registerTimeout(finishTransition, props.transitionDuration)
+    }
+
+    onDeactivated(() => {
+      finishTransition?.()
+    })
+
     function handleShow(evt) {
       addToHistory()
 
@@ -212,8 +233,7 @@ export default /*#__PURE__*/ createComponent({
         registerTick(focus)
       }
 
-      // should removeTimeout() if this gets removed
-      registerTimeout(() => {
+      registerTransitionEnd(() => {
         if ($q.platform.is.ios) {
           if (!props.seamless && document.activeElement) {
             const { top, bottom } =
@@ -247,7 +267,7 @@ export default /*#__PURE__*/ createComponent({
         showPortal(true) // done showing portal
         animating.value = false
         emit('show', evt)
-      }, props.transitionDuration)
+      })
     }
 
     function handleHide(evt) {
@@ -269,12 +289,11 @@ export default /*#__PURE__*/ createComponent({
         })
       }
 
-      // should removeTimeout() if this gets removed
-      registerTimeout(() => {
+      registerTransitionEnd(() => {
         hidePortal(true) // done hiding, now destroy
         animating.value = false
         emit('hide', evt)
-      }, props.transitionDuration)
+      })
     }
 
     function handleRouteChange() {
--- a/ui/src/components/dialog/QDialog.test.js
+++ b/ui/src/components/dialog/QDialog.test.js
@@ -9,7 +9,8 @@ import {
   vi
 } from 'vitest'
 
-import { defineComponent, h } from 'vue'
+import { KeepAlive, defineComponent, h } from 'vue'
+import { useRoute } from 'vue-router'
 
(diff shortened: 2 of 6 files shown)


rstoenescu:
Thanks for contributing!

PR #18527, opened 2026-08-25
fix(QDialog): keep sticky content in place while the page scroll is locked — fix #18183

### LINKED ISSUE #18183: QDialog: Scrolled sticky elements disappear in Firefox after opening Dialog

arbaev:
Fixes #18183

Follow-up to #18191, which was closed with an invitation to redo it against current `dev`, scoped to this bug only.

- [x] Bugfix

- [x] No

- [ ] It's been tested on a Cordova (iOS, Android) app — not applicable: the iOS path is left byte-for-byte identical (see "iOS" below)
- [ ] It's been tested on an Electron app — not applicable: no platform-specific code touched
- [x] Any necessary documentation has been added or updated — no docs change needed: no public API and no documented behavior changes; the two CSS classes that disappear are internal and undocumented

**Other information:**

### The problem

With a `position: sticky` element on a scrollable page: scroll down, open a QDialog. In Firefox the sticky element disappears until the dialog closes; in Chrome it stays put (videos and StackBlitz repro in #18183).

### Root cause

Locking the page scroll pins the body and compensates the offset:

```sass
.q-document--prevent-scroll
  body
    position: fixed !important
```
```js
body.style.left = `-${scrollPositionX}px`
body.style.top = `-${scrollPositionY}px`
```

A sticky element inside a `position: fixed` subtree has no scrollport to stick to, so it gets no sticky offset — and its static position is `scrollPositionY` pixels above the viewport, because that is exactly how far the pinned body was shifted up. Firefox renders it there (i.e. off-screen); Blink clamps a sticky element inside a partially-offscreen fixed container into the viewport instead ([Bugzilla 1437761](https://bugzilla.mozilla.org/show_bug.cgi?id=1437761)), which is why Chrome looks correct. Nothing here is specific to QDialog: QDrawer in "mobile" behavior and the Loading plugin take the same lock.

### The fix

Clip the viewport instead of taking the page out of the flow:

```sass
.q-document--clip-scroll
  overflow: hidden !important
```

The page keeps its position in the flow and its scroll offset, so sticky content keeps sticking, and there is no scroll position to lose and restore when the lock lifts. As a side effect the page no longer jumps to the top under the lock (the `#7012` synthetic scroll simply does not happen any more).

Two details:

**Scrollbar gutter.** A clipped viewport stops painting its classic scrollbar, and the page would shift sideways into the freed space. The old fix for that (`q-body--force-scrollbar-x/y`, i.e. `overflow: scroll` on the body, propagated to the viewport) cannot work here — the root element's `overflow` no longer lets the body's propagate, and making the body its own scroll container would break sticky all over again. `scrollbar-gutter: stable` on the root reserves exactly the space the scrollbar had, and it shrinks the ICB the same way, so fixed overlays (the dialog backdrop, a QLayout header) keep their width too — measured 885px of a 900px viewport before, during and after the lock. Overlay scrollbars take up no space, so the class is only applied when `window.innerWidth - document.documentElement.clientWidth > 0`, which is false for them.

**iOS.** iOS pans the page by touch whatever the viewport says, so it keeps the pinned body, under its own class, along with the existing `visualViewport` plumbing. Its behavior is unchanged, including the scroll restore on release. Sticky content stays broken there — there is no alternative to pinning on that platform.

```sass
.q-document--prevent-scroll      // marker + overscroll-behavior, both modes
.q-document--clip-scroll         // everywhere but iOS
.q-document--reserve-scrollbar   // only when a classic scrollbar had space
.q-document--pin-body            // iOS
```

`q-body--force-scrollbar-x` and `q-body--force-scrollbar-y` are gone: they only ever existed to keep the pinned page's scrollbar, and the platform that still pins (iOS) has no classic scrollbars.

### Behavior notes

- The scroll position is only restored when the body was pinned. A clipped page never lost its position, and if the app moved it meanwhile (a router `scrollBehavior` on a navigation under an open Loading overlay, say), that move is the app's and not ours to undo.
- After a navigation that happens while the lock is held, the page no longer snaps to the top on its own — the router's scroll behavior decides, exactly as it does for a navigation without a lock. QLayout still re-syncs on release through the existing release listeners (#12994).
- `scrollbar-gutter` needs Safari 17.4+ (Chrome 94+, Firefox 97+). macOS uses overlay scrollbars by default, which need no gutter at all; only Safari 16.4–17.3 with "Always show scroll bars" can still see the old sideways shift.
- The gutter is reserved on the inline axis only, so a page with a *horizontal* scrollbar gains its height (~15px) while locked. The page content does not move (it is anchored top-left); only full-height fixed overlays get that much taller.

### Scope

- `ui/src/css/core/helpers.sass`
- `ui/src/utils/scroll/prevent-scroll.js`
- `ui/src/utils/scroll/prevent-scroll.test.js`
- `ui/playground/src/pages/web-tests/regression-sweep.vue` — new S27 scenario, plus S19 updated (see Validation)

### Validation

- `ui/src/utils/scroll/prevent-scroll.test.js`: the lock's two modes are covered separately — the default one clips the viewport and leaves the page (and `document.body.style`) alone, the iOS one pins the body at the offsets and is the only one that restores a scroll position on release. Plus the gutter: reserved when a classic scrollbar had layout space, skipped when it had none.
- `ui/playground/src/pages/web-tests/regression-sweep.vue`: new scenario **S27** — a sticky bar on a scrolled page, a QDialog opened over it, asserting the bar stays stuck and the page does not move, then that both survive the close. It fails on `dev` (`whileOpen=false`: the pinned page reads `scrollY: 0`) and passes with this change, in every engine the sweep drives.
- **S19** (#12994) had to move with the mechanism: it asserted that the page ends up at the top after a route change under the lock, which was a side effect of pinning the body — the page had nowhere else to go. The lock no longer moves the page, so the scenario now performs the navigation's own scroll-to-top, which is what a router does, and still asserts what #12994 is about: QLayout re-syncing with it once the lock lifts. It passes both before and after this change.
- `ui/src/components/infinite-scroll/QInfiniteScroll.test.js`: same reason — the scenario used the lock itself to put the page at the top. It now scrolls there explicitly while the lock is held (what an app navigating under an overlay, or the pinned lock on iOS, does), and still asserts that reverse mode neither polls under the lock nor misses the poll on release.

Everything below was run on Linux; iOS and Safari could not be exercised here (`pnpm test:sweep --ios` needs macOS + Xcode). The iOS path is unchanged by construction: it keeps the class that pins the body, the `visualViewport`/`onAppleScroll` plumbing and the scroll restore it always had.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved page scroll locking across platforms, including iOS.
  * Preserved sticky positioning and scroll state while overlays or dialogs are active.
  * Improved scrollbar handling to prevent unwanted layout shifts.
  * Correctly restores scroll position when unlocking without route changes.

* **Tests**
  * Added regression coverage for sticky content during locked scrolling.
  * Expanded validation for iOS behavior, scrollbar handling, and overlay interactions.

### DIFF
--- a/ui/playground/src/pages/web-tests/regression-sweep.vue
+++ b/ui/playground/src/pages/web-tests/regression-sweep.vue
@@ -326,6 +326,17 @@
       </div>
     </div>
 
+    <!-- S27: #18183 sticky content vs the page scroll lock; mounted only
+         while the scenario runs, so the bar cannot cover other fixtures -->
+    <template v-if="s27on">
+      <div class="s27-wrap">
+        <div class="s27-sticky">S27 sticky</div>
+      </div>
+      <q-dialog v-model="s27dlg">
+        <div class="bg-white q-pa-md">S27 dialog</div>
+      </q-dialog>
+    </template>
+
     <!-- S19: #12994 reveal QHeader must reappear when a route change
          happens while the Loading plugin keeps the page scroll-locked.
          Own window-scrolling layout: the bug lives in the non-container
@@ -405,6 +416,8 @@ function s26onHold() {
 function s26onRepeat() {
   s26counts.repeat++
 }
+const s27on = ref(false)
+const s27dlg = ref(false)
 // the "Multiple masks" docs pattern (docs/src/examples/QInput/MaskMultiple.vue)
 const s18mask = computed(() =>
   s18model.value !== null && s18model.value.length > 10
@@ -992,11 +1005,11 @@ async function s18() {
   )
 }
 
-// S19: #12994 -- while the Loading plugin holds the scroll lock, QLayout
-// suppresses the lock's synthetic scroll-to-top (#7012); when a route
-// change happens under the lock, the unlock skips the scroll restore, so
-// the layout used to keep the stale pre-navigation scroll state and a
-// reveal QHeader never reappeared at the top of the new page
+// S19: #12994 -- QLayout suppresses page scrolls for as long as the
+// Loading plugin holds the scroll lock (#7012); when a route change
+// happens under the lock, the unlock skips the scroll restore, so the
+// layout used to keep the stale pre-navigation scroll state and a reveal
+// QHeader never reappeared at the top of the new page
 async function s19() {
   s19on.value = true
   // scroll anchoring (Firefox especially) can nudge the position back up
@@ -1029,9 +1042,14 @@ async function s19() {
   const path = window.location.pathname
   $q.loading.show({ delay: 0 })
   await settle(400)
-  // a route change as prevent-scroll sees one: the pathname moves on
-  // while the lock is held (no popstate, the router stays put)
+  // a navigation as prevent-scroll sees one: the pathname moves on while
+  // the lock is held (no popstate, the router stays put) and the router's
+  // scroll behavior takes the new route to the top. That scroll is the
+  // navigation's, not the lock's -- the lock only clips the viewport and
+  // leaves the page where it is (#18183) -- and QLayout has to re-sync
+  // with it once the lock lifts
   history.pushState({}, '', path + '/s19-nav')
+  window.scrollTo(0, 0)
   $q.loading.hide()
   await settle(500)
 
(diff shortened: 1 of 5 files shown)


rstoenescu:
Landed on dev as af2a84458, cherry-picked with your authorship intact, so closing rather than merging. Thank you for the excellent work here: the root-cause writeup, the clip + scrollbar-gutter approach and the test coverage were all spot on and it shipped essentially as submitted.

Two small follow-ups landed with it: c8a390d0a adjusts the new S27 sweep scenario to measure stuckness against a position: fixed probe (on the iOS Simulator run, real Mobile Safari offsets viewport-anchored rects by its collapsing chrome, so the absolute rect.top assertion failed there even though the behavior was correct), and 459a87904 updates the now-stale pin-era comments in QInfiniteScroll/QLayout. With that, the sweep passes 27/27 on chromium, firefox, webkit, desktop Safari and the iOS Simulator.

Will be part of the next quasar version. Thanks again!

PR #18528, opened 2026-08-25
fix(QSelect): map a lazy loaded model value on its own (fix #17983)

### LINKED ISSUE #17983: QSelect fails to populate the correct initial value if using lazy loading with map-options and emit-value
### What happened?

When using QSelect with both lazy loading and map-options and emit-value and the model has some initial value, the control initially shows the *value* of the of the model and not the label. If you then use the control to select the same (or other) value then the correct label is shown.

The control should initially show the correct label.

### Reproduction URL

https://codepen.io/Stormalong42/pen/qEERrMN?editors=1011

### How to reproduce?

1. Go to the provided reproduction link
2. Observe that the model is showing "twt" (correct) and the select control is displaying "twt" (incorrect, should be 'Twitter'
3. Click on the select control and select Facebook
4. Observer that model is now "fb" and the select control is "Facebook" (correct)

### Flavour

### Areas

Components (quasar)

### Platforms/Browsers

Chrome

### Quasar info output

```Shell
Its on codepen
```

### Relevant log output

```Shell

```

### Additional context

Pretty sure this occurs because the "options" property is empty when the component is mounted (since its lazy loaded) so the map-options has nothing to map against. 

For now I have created a work-around. In onMounted/nextTick (didn't work without nextTick) I do a quick search for the initial value and then populate the options (the same one populated by the lazy-load filter function) with the single item that matches the initial value. This seems to work.

My feeling is that the "correct" fix is call the filter function when the component is created so that it can populate the options for the initial value, but I haven't looked into the code yet.

If I get some time I'll dive into the code and submit a fix myself.

arbaev:
Fixes #17983

- [x] Bugfix

- [x] No

- [ ] It's been tested on a Cordova (iOS, Android) app — not applicable: no platform-specific code path is touched
- [x] Any necessary documentation has been added or updated — a tip in the "Lazy loading" section of the QSelect docs page; no public API was added

**Other information:**

### The problem

Reproduction from the issue ([CodePen](https://codepen.io/Stormalong42/pen/qEERrMN)): a QSelect with lazy loaded options (`@filter`), `map-options` and `emit-value`, whose model already holds a value. The field displays the raw value (`twt`) instead of its label (`Twitter`). Opening the menu once — anything that loads the options — makes it correct.

### Root cause

`map-options` maps the model in `innerValue` through `getOption()`:

```js
function getOption(value, valueCache) {
  const fn = opt => isDeepEqual(getOptionValue.value(opt), value)
  return props.options.find(fn) || valueCache.find(fn) || value
}
```

With lazy loading, `props.options` is still empty on the first render and the cache of previously mapped values is empty too, so the raw model value is what falls out and what the field displays.

The options only arrive when `filter()` runs, and it is called from `showPopup()`, from typing into the input and from `new-value-mode` — always as the result of a user interaction. Before the first popup, QSelect has no label to show and no way to obtain one.

### The fix

Let QSelect ask for the options itself, in exactly the case where the display is objectively wrong:

```js
function prefetchUnmappedOptions() {
  if (
    prefetchDone ||
    props.onFilter === void 0 ||
    !props.mapOptions ||
    // options are already in: an empty filter won't add a missing value
    virtualScrollLength.value !== 0 ||
    // a model holding whole options already carries its labels
    innerValue.value.every(opt => Object(opt) === opt)
  ) {
    return
  }

  prefetchDone = true
  filter('', true, () => { resetInputValue() })
}
```

It reuses the `keepClosed` mode `filter()` already has, so the options are fetched with an empty search string and no popup is shown. The check runs from `onMounted` and from a `watch(innerValue, …)` — a form that fetches its record lands the model value after mounting, which is the common shape of the lazy-loading case — with `prefetchDone` capping it at one request per component instance.

Doing this behind a new opt-in prop was the other option, and it was rejected: it would leave the wrong label in place for everyone who does not know about the prop, which is the bug being reported here.

The `afterUpdateFn` argument is not decoration. With `use-input` + `fill-input`, the input's text comes from the `innerValue` watcher, which bails out while `state.innerLoading` is set — and that flag is only cleared on the tick *after* the options land, i.e. after `innerValue` has already recomputed. `afterUpdateFn` runs past that point, so `resetInputValue()` there is what puts the label into the input.

The same code path needed one more fix: in the keep-closed mode, the `abort()` callback of `filter()` only cleared the loading state when the component was focused, so an aborted prefetch (no network, a handler that gives up) would have left the spinner in place forever. Its condition is now the same one the `update()` callback uses.

### Behavior notes

- Nothing happens when the options are already loaded, when the model is empty, when `map-options` is off, or when the model holds whole option objects (those carry their own labels) — no existing setup gains a request it did not make before.
- A value that the loaded options do not contain (a deleted record, server-side pagination) is requested once and never again.
- The regular loading indicator shows while the prefetch is in flight, exactly as it does for any other filtering.
- If the user opens the select while a prefetch is in flight, `showPopup()` → `filter()` supersedes it through the existing `filterId` bookkeeping.
- SSR is unaffected: `onMounted` does not run on the server, so the server-rendered markup keeps the raw value and the swap happens after hydration.

### Scope

- `ui/src/components/select/QSelect.js`
- `ui/src/components/select/QSelect.test.js`
- `docs/src/pages/vue-components/select.md`

### Validation

- Five new tests in `QSelect.test.js`: the issue's scenario (the handler is called once with `''`, no menu is rendered, the label shows up), the cases that must *not* request anything, a model value arriving after mounting (and not requesting again for the next missing one), the `use-input` + `fill-input` input text, and the aborted prefetch dropping the loading state (in `[Generic]`). Four of them fail on `dev`; the fifth is the negative one that guards the conditions.
- `pnpm test:specs --target components/select/QSelect`: ✅.
- `pnpm test:unit`: 260 files / 5741 tests green. Full `pnpm test`: every suite green except one unrelated flake of the parallel runner (`QTree > virtual scroll keyboard navigation reaches rows outside the slice`), which passes on its own (121/121) and lives in a component this PR does not touch.
- Manual check in a real browser (playground page driven headless over CDP), four scenarios — the issue's own, `use-input` + `fill-input`, a model value arriving 2s after mounting, and a handler that calls `abort()`. With the fix they render "Twitter" / "Facebook" / "Google" with no interaction and the aborted one drops its spinner; on the reverted fix the same page shows "twt" / "fb" / "goog" and the handler is never called.
- `pnpm lint` and `git diff --check` clean.

## Summary by CodeRabbit

* **Enhancements**
  * Improved lazy loading for select fields using mapped values.
  * Automatically retrieves missing options for selected values without opening the dropdown.
  * Supports values that arrive after the component loads and updates displayed labels once options are available.

* **Bug Fixes**
  * Prevented unnecessary requests for values that are already available.
  * Ensured loading indicators clear correctly when filtering is canceled.

### DIFF
--- a/docs/src/pages/vue-components/select.md
+++ b/docs/src/pages/vue-components/select.md
@@ -133,6 +133,10 @@ The following example shows a glimpse of how you can play with lazy loading the 
 While options are being loaded, the default loading spinner takes the place of the dropdown icon so the field keeps a constant width. For this reason, when `hide-dropdown-icon` is used the default spinner is not displayed at all (it would make the field's width jump); supply a `loading` slot if you still want an inline indicator in that case.
 :::
 
+::: tip
+When the model already holds a value and `map-options` is used, there is nothing to map it against until the options are loaded, so the field would display the raw value. Starting with v2.27, QSelect asks for them on its own in this case: it calls your `@filter` handler once with an empty search string and without opening the menu, so the correct label shows up without any user interaction. The same happens when the model value arrives later (a record loaded from the server, for instance). A value that the loaded options do not contain is not requested again.
+:::
+
 You can dynamically load new options when scroll reaches the end:
 
 <DocExample title="Dynamic loading options" file="OptionsDynamic" />
--- a/ui/src/components/select/QSelect.js
+++ b/ui/src/components/select/QSelect.js
@@ -5,6 +5,7 @@ import {
   nextTick,
   onBeforeUnmount,
   onBeforeUpdate,
+  onMounted,
   onUpdated,
   ref,
   watch
@@ -191,6 +192,7 @@ export default /*#__PURE__*/ createComponent({
     let filterTimer = null,
       inputValueTimer = null,
       innerValueCache,
+      prefetchDone = false,
       hasDialog,
       userInputValue,
       filterId = null,
@@ -511,6 +513,10 @@ export default /*#__PURE__*/ createComponent({
     )
 
     watch(() => props.fillInput, resetInputValue)
+
+    // not immediate: the initial run happens in onMounted, since filtering
+    // must not be triggered while the component is still setting up
+    watch(innerValue, prefetchUnmappedOptions)
 
     watch(menu, updateMenu)
 
@@ -1265,7 +1271,10 @@ export default /*#__PURE__*/ createComponent({
           }
         },
         () => {
-          if (state.focused.value && filterId === localFilterId) {
+          if (
+            (keepClosed || state.focused.value) &&
+            filterId === localFilterId
+          ) {
             clearTimeout(filterId)
             state.innerLoading.value = false
             innerLoadingIndicator.value = false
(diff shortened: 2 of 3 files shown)


rstoenescu:
Thank you for the excellent analysis and the PR! This has landed in dev via 28e42951b, based on your work, with one addition: an opt-out prop (`no-option-prefetch`) for apps that must not fire requests without user interaction, for example when the parent component loads the options itself or the backend is paginated. The internals were also slimmed down so that selects without an armed prefetch register no extra watcher or hook. Your root-cause writeup and the keep-closed abort() fix made this an easy one to work with. It will be part of the next `quasar` version (v2.28). Closing the PR since the work landed directly on dev, with credit to you.

PR #18531, opened 2026-08-29
fix(QIntersection): don't remount the content when disable is toggled #12668

### LINKED ISSUE #12668: Elements wrapped in q-intersection are remounted on toggeling disable-prop
### What happened?

Docs for QIntersection > disable:

> Disable visibility observable (content will remain as it was, visible or hidden)

in v1 toggeling the disable property did not cause a umount/mount of the currently visible elements, but now it does, even though the DOM representation is identical for those elements.

Currently visible elements to not remount on disable-property change

### Reproduction URL

https://codesandbox.io/s/prod-glitter-cizkj0?file=/src/pages/Index.vue

### How to reproduce?

1. Go to provided reproduction link - there is a progress bar indicator for wrapped elements after mounting for 1 second
2. Click button to change disable-prop
3. Elements show progress bar again - indicating a remount

### Flavour

Quasar CLI (@quasar/cli | @quasar/app)

### Areas

Components (quasar), Directives (quasar)

### Platforms/Browsers

Firefox, Chrome

### Quasar info output

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

giaBaoJS:

- [x] Bugfix

- [x] No

- [x] Any necessary documentation has been added or updated

**Other information:**

Fixes #12668.

## The bug

`hDir()` writes `data.key = key + condition` (`ui/src/utils/private.render/render.js:43`), so whatever is passed as its `condition` ends up in the vnode key. `QIntersection` passed `!props.disable` there (`ui/src/components/intersection/QIntersection.js:118` before this PR), so toggling `disable` flipped the root key between `maintrue` and `mainfalse`, Vue saw a different vnode type and destroyed/rebuilt the whole subtree.

That contradicts the component's own API description for `disable`, "Disable visibility observable (content will remain as it was, visible or hidden)", and it is a v1 -> v2 regression per the issue.

It is the same class as c434fc1c ("hDir bakes its condition into the vnode key"), which removed the transient `isRuntimeSsrPreHydration` part of that condition last week. `!props.disable` was the remaining key flip.

## The fix

The `hDir` condition is now the constant `true`, so the key never changes. Disabling is expressed through the directive value instead: `QIntersection` passes `false` when `disable` is set, and the `Intersection` directive treats Boolean `false` as "no observer", disconnecting it and clearing the cached `cfg`/`handler` so a re-enable rebuilds the observer from scratch.

Mounting with `disable` still creates no `IntersectionObserver` at all, which the pre-existing `[(prop)disable]` test asserts.

### Public API

This widens the accepted `v-intersection` value from `Object | Function` to `Boolean | Object | Function`; only `false` is special, everything else keeps its current meaning. `v-ripple` (`binding.value !== false`, `Ripple.json` declares `["Boolean", "Object"]`) and `v-close-popup` already accept a Boolean the same way. `Intersection.json`, the generated `IntersectionValue` type and the directive docs page are updated together, and the docs page gets a `Disable` example alongside Ripple's.

### Behavior change beyond the fix

With `once`, an observer that has already fired is no longer re-armed by toggling `disable` off and on again. The re-arming only happened because the element was recreated, and it contradicted `once`.

## Blast radius

`hDir` has 11 other call sites; none of them are touched and `render.js` is unchanged. Worth noting separately: `QChip` (`ui/src/components/chip/QChip.js:254`) passes `props.ripple !== false && !props.disable` as its `hDir` condition and has the same latent key flip. I left it alone here to keep this PR to one logical change.

## Validation

Everything run from `/ui` unless stated.

- `pnpm test` (all five suites): unit 264 files / 5950 tests, hydration 89 / 142, hydration-pwa 2 / 6, umd 4 / 78, e2e-ssr 1 / 239, build 1 / 9. All pass.
- Baseline on `dev` before the change: unit 264 / 5948, hydration 89 / 142. The two new tests are the delta.
- `pnpm test:specs:check`: 264 files OK, exit 0.
- Root `pnpm lint:check`: "All matched files use the correct format", no oxlint findings in the changed files. (Two pre-existing `tsconfig-error` lines for the `app-vite` playgrounds, whose gitignored `.quasar/` is not generated in this checkout, are unrelated.)
- `/docs` `pnpm test`: unit 63 / 303, e2e-ssr 1 / 317 routes, SSG build clean, so the new example page renders and hydrates without console output.
- `git diff --check` clean.

Counterfactual for the new tests, both against the browser-mode Chromium suite:

Reverting only `QIntersection.js`:

```
FAIL src/components/intersection/QIntersection.test.js > [QIntersection API] > [Props] > [(prop)disable] > toggling it keeps the content mounted
AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
 ❯ src/components/intersection/QIntersection.test.js:208:33
    208|         expect(unmountedFn).not.toHaveBeenCalled()
```

Reverting only `Intersection.js` (3 failures, including the pre-existing `disable` test):

```
FAIL src/components/intersection/QIntersection.test.js > ... > [(prop)disable] > type Boolean has effect
AssertionError: expected [ <Anonymous Class>{ …(5) } ] to have a length of +0 but got 1
FAIL src/directives/intersection/Intersection.test.js > [Intersection API] > [Value] > as Boolean
AssertionError: expected [ <Anonymous Class>{ …(5) } ] to have a length of +0 but got 1
```

## Summary by CodeRabbit

- **New Features**
  - Added support for disabling the `v-intersection` directive by setting its value to `false`.
  - Re-enabling the directive automatically resumes observation.
  - Disabling preserves the observed element and its content.

- **Bug Fixes**
  - Disabling now cleanly stops observation and prevents outdated callbacks from running.
  - One-time observation remains inactive after being disabled.

- **Documentation**
  - Added guidance and an interactive example demonstrating conditional directive enablement.

### DIFF
--- a/docs/src/examples/Intersection/Disable.vue
+++ b/docs/src/examples/Intersection/Disable.vue
@@ -1,0 +1,62 @@
+<template>
+  <div class="relative-position">
+    <div class="example-area q-pa-lg scroll">
+      <div class="example-filler" />
+
+      <div
+        v-intersection="enabled ? onIntersection : false"
+        class="example-observed text-center rounded-borders"
+      >
+        Observed Element
+      </div>
+
+      <div class="example-filler" />
+    </div>
+
+    <div
+      class="example-state rounded-borders absolute-top q-mt-md q-ml-md q-mr-lg text-white"
+      :class="visibleClass"
+    >
+      <div class="text-center">{{ visible ? 'Visible' : 'Hidden' }}</div>
+
+      <q-toggle v-model="enabled" label="Observe" dense />
+    </div>
+  </div>
+</template>
+
+<script setup>
+import { computed, ref } from 'vue'
+
+const enabled = ref(true)
+const visible = ref(false)
+
+const visibleClass = computed(
+  () => `bg-${visible.value ? 'positive' : 'negative'}`
+)
+
+function onIntersection(entry) {
+  visible.value = entry.isIntersecting
+}
+</script>
+
+<style lang="sass" scoped>
+.example-state
+  background: #ccc
+  font-size: 20px
+  color: #282a37
+  padding: 10px
+  opacity: 0.8
+
+.example-observed
+  width: 100%
+  font-size: 20px
+  color: #ccc
+  background: #424242
+  padding: 10px
+
+.example-area
+  height: 300px
+
+.example-filler
+  height: 500px
+</style>
--- a/docs/src/pages/vue-directives/intersection.md
+++ b/docs/src/pages/vue-directives/intersection.md
@@ -63,6 +63,12 @@ By passing in an Object as the directive's value (instead of a Function), you ca
 
 <DocExample title="Supplying configuration Object" file="ObjectForm" no-edit />
 
+### Disable
+
+Passing in Boolean `false` instead of a Function or an Object disables the directive: the Intersection Observer is dropped until a handler is supplied again. The DOM element is untouched in the process, so whatever it wraps keeps its state.
+
+<DocExample title="Disable" file="Disable" no-edit />
+
 ### Advanced
 
 Below is a more advanced example of what you can do. The code takes advantage of the HTML `data` attribute. Basically, by setting `data-id` with the index of the element in a loop, this can be retrieved via the passed in `entry` to the handler as `entry.target.dataset.id`. If you are unfamiliar with the `data` attribute you can read more [here](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) about using it.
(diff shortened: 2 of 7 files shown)


giaBaoJS:
Pushed the tests that should have come with the behaviour change here. This PR stops a toggled `disable` from re-arming an already fired `once` observer, so QIntersection now asserts that toggling `disable` off and on leaves that single observer spent. The directive side covers the matching guarantee: the disable branch clears the cached handler, so an entry queued before the disable has nothing left to run and cannot dereference the observer it just dropped.

rstoenescu:
Thanks for contributing!

PR #18539, opened 2026-09-04
fix(ui): compare binary data by content

lprnmns:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

`is.deepEqual()` treated every pair of ArrayBuffers with the same constructor as equal because ArrayBuffer contents are not enumerable. DataView comparisons could also loop indefinitely because the existing ArrayBuffer-view path iterated a missing `.length` property.

This compares ArrayBuffer bytes through Uint8Array views and handles each DataView's exposed byte range. Existing typed-array comparison behavior is unchanged.

- [x] Any necessary documentation has been added or updated, or explained in the PR's description

No documentation change is needed: this restores the documented ArrayBuffer support of `is.deepEqual()` without changing its API.

Validation:

- `pnpm --dir ui exec vitest run -c test/vitest.config.js src/utils/is/is.test.js` — 70 passed
- `pnpm lint:check` — passed
- `git diff --check` — passed

The aggregate `pnpm test` baseline was also attempted. Its build/spec checks and many browser tests passed, but parallel browser suites exhausted the host file-watcher limit (`ENOSPC`); the isolated affected suite is green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deep equality comparisons for `ArrayBuffer` and `DataView` values by checking their byte contents.
  * Prevents matching binary data from being treated as different, or differing data as equivalent.

* **Tests**
  * Added coverage for matching and non-matching `ArrayBuffer` and `DataView` values.

### DIFF
--- a/ui/src/utils/is/is.js
+++ b/ui/src/utils/is/is.js
@@ -68,6 +68,17 @@ export function isDeepEqual(a, b) {
       return true
     }
 
+    if (a.constructor === ArrayBuffer) {
+      return isDeepEqual(new Uint8Array(a), new Uint8Array(b))
+    }
+
+    if (a.constructor === DataView) {
+      return isDeepEqual(
+        new Uint8Array(a.buffer, a.byteOffset, a.byteLength),
+        new Uint8Array(b.buffer, b.byteOffset, b.byteLength)
+      )
+    }
+
     // oxlint-disable-next-line eqeqeq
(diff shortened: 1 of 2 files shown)


rstoenescu:
Thanks for contributing!
One follow-up is needed by me next.

PR #18540, opened 2026-09-04
fix(ui): Handle null-prototype objects in deepEqual

lprnmns:

- [x] Bugfix

- [x] No

**Problem**

`is.deepEqual()` throws for objects created with `Object.create(null)` because those objects do not inherit callable `valueOf()` or `toString()` methods.

**Solution**

Only use custom conversion methods when both operands provide callable, non-native methods. Objects without `Object.prototype`, and asymmetric overrides, then use the existing structural key comparison.

**Validation**

- [x] Submitted against `dev`.
- [x] Focused browser suite: 79 passed; regression failed before the fix with the expected `TypeError`.
- [x] Specs validation for `utils/is/is` passed.
- [x] Root formatting and lint checks passed.
- [x] Full UI test run passed: build 11, unit 6,206, hydration 143, hydration-PWA 6, UMD 78, SSR E2E 239.
- [x] After the review follow-up, focused tests, Specs validation, formatting and lint were rerun on the final head.
- [x] Documentation remains accurate; no API, types, or options changed.
- [ ] Cordova/Electron manual testing — not applicable to this platform-independent utility fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved deep equality comparisons for objects when one or both values lack custom `valueOf` or `toString` methods.
  - Correctly handles objects that use default conversion methods or have no standard object prototype.

- **Tests**
  - Added regression coverage for comparing objects with asymmetric `valueOf` or `toString` implementations in either operand order.
  - Retained coverage for prototype-less objects.

### DIFF
--- a/ui/src/utils/is/is.js
+++ b/ui/src/utils/is/is.js
@@ -99,7 +99,8 @@ export function isDeepEqual(a, b) {
     if (
       typeof a.valueOf === 'function' &&
       typeof b.valueOf === 'function' &&
-      a.valueOf !== Object.prototype.valueOf
+      a.valueOf !== Object.prototype.valueOf &&
+      b.valueOf !== Object.prototype.valueOf
     ) {
       return a.valueOf() === b.valueOf()
     }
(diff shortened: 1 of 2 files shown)


rstoenescu:
Thanks for contributing!

PR #18541, opened 2026-09-06
fix(Ripple): the argument and the modifiers follow runtime changes

giaBaoJS:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### The bug

`v-ripple`'s `updated()` hook only looks at the directive value:

```js
updated(el, binding) {
  if (binding.oldValue !== binding.value) {
    ...
    if (ctx.enabled && Object(binding.value) === binding.value) {
      setOptions(ctx, binding)
    }
    ...
  }
}
```

`setOptions()` is the only place the argument and the modifiers are read, so both are frozen at mount:

1. `v-ripple:[color]` with no value: the value is `undefined` on every render, so the guard never opens and the ripple keeps its mount-time color forever.
2. A modifiers object that changes while the value reference does not: `early` / `stop` / `center` / `keyCodes` stay stale.
3. Setting the value back to `true` after an object: the `Object(value) === value` check skips `setOptions()`, so the object's options survive. Mount with `true` and update to `true` end up with different settings for the same input.

Case 2 happens with Quasar's own QBtn. `rippleProps` is `computed(() => ({ center: props.round }))` and the `ripple` computed does not read `props.round`, so toggling `round` at runtime hands the directive a new modifiers object and the same value reference. The ripple keeps starting from the press point instead of the center (or the other way round).

### How to reproduce

```html
<q-btn :round="isRound" icon="star" label="x" @click="isRound = !isRound" />
```

The first click makes the button round, but the ripple keeps originating from the press point. Same for a dynamic argument:

```html
<div v-ripple:[color]>tap me</div>
```

Changing `color` has no effect after the first render.

### Root cause

`updated()` gated everything on the value reference, and `setOptions()` (the only reader of `binding.arg` and `binding.modifiers`) was reachable only through that gate and only for object values.

### The fix

`updated()` now re-derives the options whenever the value, the argument or the modifiers object differs from the one the context was built with, and `setOptions()` records the argument and the modifiers on the context for that comparison. This is the shape the four touch directives were given in 4e7f7b6 and ScrollFire in ee88f70; `v-ripple` was the remaining directive with a documented argument that never re-read it.

As noted in 4e7f7b6, the identity check only skips work for `withDirectives()` callers holding a stable object (QBtn, QChip, QTab and StepHeader all pass one from a computed); a compiled template passes a new modifiers object on every host re-render, exactly as before this PR.

### Tests

Two tests added to `Ripple.test.js`:

- `follows the argument and the modifiers when they change at runtime`: mounts with `arg = 'orange-5'` and empty modifiers, asserts the click ripple is `text-orange-5`, then switches to `arg = 'red'` and `{ early: true }` and asserts that a click no longer produces a ripple (the early listener replaced the click one) while a `pointerdown` produces a `text-red` one.
- `drops the object options when the value goes back to true`: mounts with `{ color: 'red' }`, then sets the value to `true` and asserts the next ripple is not `text-red`.

Both fail on `dev` (`expected 2 to be 1` and `expected [ 'q-ripple', 'text-red' ] to not include 'text-red'`) and pass with the fix. `pnpm test:specs:check`, the full `src/directives` unit suite, the ripple hydration test and the QBtn / QChip / QTabs / QStepper suites all pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Ripple effects now react correctly when their color or modifiers change at runtime.
  - Switching the “early” interaction modifier now updates the ripple trigger appropriately.
  - Object-based ripple options are cleared when changing back to the default enabled state.
- **Tests**
  - Added coverage for dynamic ripple configuration changes and option resets.

### DIFF
--- a/ui/src/directives/ripple/Ripple.js
+++ b/ui/src/directives/ripple/Ripple.js
@@ -158,6 +158,8 @@ function setOptions(ctx, { modifiers, value, arg }) {
   const cfg = { ...ctx.cfg, ...modifiers, ...value }
   const keyCodes = cfg.keyCodes || 13
 
+  ctx.arg = arg
+  ctx.modifiers = modifiers
   ctx.early = cfg.early === true
   ctx.stop = cfg.stop === true
   ctx.center = cfg.center === true
@@ -195,6 +197,8 @@ export default /*#__PURE__*/ createDirective(
           const ctx = {
             cfg: cfg.ripple,
             enabled: binding.value !== false,
+            arg: void 0,
+            modifiers: void 0,
             early: false,
             stop: false,
             center: false,
@@ -211,18 +215,23 @@ export default /*#__PURE__*/ createDirective(
         },
 
         updated(el, binding) {
-          if (binding.oldValue !== binding.value) {
-            const ctx = el.__qripple
-            if (ctx !== void 0) {
-              ctx.enabled = binding.value !== false
+          const ctx = el.__qripple
+          if (ctx === void 0) return
 
-              if (ctx.enabled && Object(binding.value) === binding.value) {
-                setOptions(ctx, binding)
-              }
+          const { value, oldValue, arg, modifiers } = binding
 
-              bind(el, ctx)
-            }
+          ctx.enabled = value !== false
+
+          if (
+            ctx.enabled &&
+            (value !== oldValue ||
+              arg !== ctx.arg ||
+              modifiers !== ctx.modifiers)
+          ) {
+            setOptions(ctx, binding)
           }
+
+          bind(el, ctx)
         },
 
         beforeUnmount(el) {
--- a/ui/src/directives/ripple/Ripple.test.js
+++ b/ui/src/directives/ripple/Ripple.test.js
@@ -387,6 +387,67 @@ describe('[Ripple API]', () => {
       expect(wrapper.findAll('.q-ripple').length).toBe(2)
     })
 
+    test('follows the argument and the modifiers when they change at runtime', async () => {
+      const arg = ref('orange-5')
+      const modifiers = ref({})
+      const TestComponent = defineComponent({
+        setup() {
+          return () =>
+            withDirectives(h('div'), [
+              [Ripple, void 0, arg.value, modifiers.value]
+            ])
+        }
+      })
+
+      const wrapper = mount(TestComponent)
+
+      await wrapper.trigger('click')
+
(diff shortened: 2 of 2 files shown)


rstoenescu:
Thanks @giaBaoJS. This landed on `dev` as cde626534, with you as co-author.

The diagnosis was right on all three counts, and I confirmed the QBtn one against real code: toggling `round` leaves `ctx.center` stale at `false`, exactly as you described, because `rippleProps` hands over a new modifiers object while the `ripple` computed keeps the same value reference.

Three things changed on the way in.

**The modifiers are compared by content, not by identity.** Your note about a compiled template passing a fresh object on every render is the important one, and it cuts the other way too: an identity check would re-derive the options, and allocate, for every rippled element on every render of its host. `Ripple.json` declares three modifiers, and a template modifier is always `true`, so those flags are read directly instead of walking the keys:

```js
function modifiersChanged (a, b) {
  return (
    a !== b &&
    (a.early !== b.early || a.stop !== b.stop || a.center !== b.center)
  )
}
```

Benchmarked same-process, interleaved, medians over 41 blocks of 300k: 18ns per update against 32ns for a two-pass key walk with one modifier set, 22ns against 51ns with two, and flat in the number set. Callers passing a stable object (QBtn, QChip, QTab, StepHeader), and templates written without modifiers (Vue substitutes a shared frozen object), settle on the identity comparison. The cost is that an undeclared key dropped in the modifiers slot no longer brings a re-read of its own; it belongs in the value.

**The tests were split.** CodeRabbit's comment held up. The test moved the argument and the modifiers in the same tick, so either one alone satisfied the condition, and I confirmed it by mutation: deleting `arg !== ctx.arg` left all 26 tests green, and so did deleting the modifiers term. They are now separate cases under `[Argument]` and `[(modifier)early]`, each holding the other two inputs still, alongside your value-back-to-`true` case under `[Value]`. Each fails when its own branch is removed.

**QBtn covers the real-world case**, under `[(prop)round]`: two off-center clicks give different transforms while the button is rectangular and identical ones once it is round, so it does not lean on the transform format. There is also a `[Generic]` test pinning the content comparison, which fails the moment it degrades to identity.

The branch had also picked up a conflict with 6d8632db5 in the meantime, so closing this in favour of the commit. Thanks again, the writeup made it unusually quick to verify.

PR #18544, opened 2026-09-08
fix(debounce): guard reentrant immediate calls

lprnmns:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

An immediate debounced callback can call itself again before the timer is assigned, causing a second leading invocation within the same wait window. Independently reproduced on current `dev`:

```js
let calls = 0
const fn = debounce(() => {
  if (++calls === 1) fn()
}, 100, true)
fn()
// Previously calls === 2; expected 1.
```

Schedule the timer before invoking the callback, retaining the initial leading-edge decision. The regression also checks that another leading invocation is allowed after the wait. The public signature is unchanged; existing trailing, cancellation and receiver/argument tests remain intact.

Validation:
- Targeted debounce test: regression failed before the fix (expected 1 call, received 2); all 15 tests passed after.
- `pnpm --dir ui test:specs --target utils/debounce/debounce`: passed.
- `pnpm lint`: passed.
- Full root `pnpm test`: passed (6,246 unit, 143 hydration, 78 UMD, 11 build, 6 hydration-PWA and 239 SSR tests). Four SSR routes required the suite's automatic retry: intersection-6-iframe, regression-sweep, screen-height and tree-perf.
- `git diff --check`: passed.

Cordova and Electron application tests were not run.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed immediate debounce behavior so calls made within the callback are correctly delayed and debounced.

- **Documentation**
  - Clarified that the debounce wait period starts before an immediate callback runs.

- **Tests**
  - Added coverage verifying nested calls are delayed and execute only after the debounce period.

### DIFF
--- a/docs/src/pages/quasar-utils/other-utils.md
+++ b/docs/src/pages/quasar-utils/other-utils.md
@@ -357,7 +357,7 @@ If your App uses JavaScript to accomplish taxing tasks, a debounce function is e
 
 Debouncing enforces that a function not be called again until a certain amount of time has passed without it being called. As in "execute this function only if 100 milliseconds have passed without it being called."
 
-With `immediate: true`, the wait period starts before the callback runs, so calls made from within that callback are debounced too.
+When `immediate` is `true`, the wait period starts before the callback runs, so calls made from within that callback are debounced too.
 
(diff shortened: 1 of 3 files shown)


rstoenescu:
Thanks again for contributing! This will go into 2.30.2

PR #18548, opened 2026-09-13
fix(ui): extractDate matches the longest locale name, not the first

giaBaoJS:

- [x] Bugfix

- [x] No

## The bug

`getLocaleData()` (`ui/src/utils/date/date.js`) builds the `MMM`/`MMMM`/`ddd`/`dddd`/`dd` parser alternations in language pack declaration order:

```js
const days = '(' + dateLocale.days.map(escapeRegex).join('|') + ')',
  ...
```

Regex alternation is first-match, and the parser is only anchored at the start (`new RegExp('^' + regexText)`, line 260, deliberately so that trailing text is allowed). So wherever the name token is not followed by a literal that would force the engine to backtrack into the longer branch, a name that is the prefix of a later one wins, and `formatDate()` output does not round-trip back through `extractDate()`.

A concrete round-trip, entirely within the shipped `vi` pack:

```js
import { date } from 'quasar'
import vi from 'quasar/lang/vi'

const value = new Date(2023, 11, 1) // December 1st 2023

date.formatDate(value, 'YYYY MMM', vi.date)
// "2023 Th12"

date.extractDate('2023 Th12', 'YYYY MMM', vi.date)
// January 1st 2023  <- "Th1" matched first, "2" left as trailing text
```

Same shape on the day side:

```js
import tr from 'quasar/lang/tr'

date.extractDate('2023-W01-Pazartesi', 'GGGG-[W]ww-dddd', tr.date)
// Sunday January 8th 2023, not Monday January 2nd
// "Pazar" (Sunday) is the prefix of "Pazartesi" (Monday)
```

This contradicts the round-trip contract documented at `docs/src/pages/quasar-utils/date-utils.md:472`.

## Which locales are affected

I round-tripped `formatDate()` through `extractDate()` for every month name and every day name of all 74 shipped packs, across the `MMMM`, `MMM`, `dddd`, `ddd` and `dd` tokens. On `dev` that produces **16 failing round-trips across 7 locales**:

| locale | token | formatted | parses back as |
| --- | --- | --- | --- |
| az-Latn | dddd | `2023-W01-Bazar ertəsi` | Sun Jan 8 (want Mon Jan 2) |
| az-Latn | ddd | `2023-W01-B.E` | Sun Jan 8 (want Mon Jan 2) |
| az-Latn | ddd | `2023-W01-Ç.E` | Tue Jan 3 (want Wed Jan 4) |
| ckb | MMMM, MMM | `2023 مانگی دووانزە` | month 2 (want 12) |
| cs | MMMM | `2023 Červenec` | month 6 (want 7) |
| kur-CKB | MMMM, MMM | `2023 مانگی دووانزە` | month 2 (want 12) |
| lt | ddd | `2023-W01-Pn` | Mon Jan 2 (want Fri Jan 6) |
| tr | dddd | `2023-W01-Pazartesi` | Sun Jan 8 (want Mon Jan 2) |
| tr | dddd | `2023-W01-Cumartesi` | Fri Jan 6 (want Sat Jan 7) |
| vi | MMMM | `2023 Tháng Mười Một` | month 10 (want 11) |
| vi | MMMM | `2023 Tháng Mười Hai` | month 10 (want 12) |
| vi | MMM | `2023 Th10`, `Th11`, `Th12` | month 1 |

All 16 pass after the fix.

One further failure, `lv` `daysShort`, is **not** fixed and is out of scope: `lv.js` lists `"Pi"` for both Monday and Friday, so that name is genuinely ambiguous no matter how the alternation is ordered. That is language pack data, not parser behaviour. It is the reason the fix is not simply "reorder the packs": custom locales passed by users have the same problem, and reordering the packs would not help them.

## The fix

Sort each alternation longest-first before joining, so a prefix can never shadow the longer name it prefixes:

```js
function namesRegex(list) {
  return (
    '(' +
    [...list].sort((a, b) => b.length - a.length).map(escapeRegex).join('|') +
    ')'
  )
}
```

Notes:

- The sort works on a copy. `extractDate()` resolves the matched text back to a number with `months.indexOf(...)` (`:544`), `langOpts.days.indexOf(...)` (`:429`) and `daysMinList.indexOf(...)` (`:437`), all against the **unsorted** arrays, so month and day numbering is untouched.
- `Array.prototype.sort` is stable, so equal-length names keep their declaration order and nothing else about the generated regex moves.
- `escapeRegex` still runs on every name, so the existing metacharacter behaviour is unchanged.
- No public API change, no locale file touched.

## What the tests prove

9 new tests in `ui/src/utils/date/date.test.js`, under `[(function)extractDate]`:

- 4 parameterised cases building a custom locale from `en-US` with an `Alpha` / `AlphaBeta` prefix pair, one per name list (`months`, `monthsShort`, `days`, `daysShort`).
- 1 case for `dd`, where a one-character day name is the only way to produce a prefix pair after the two-character shortening.
- 2 cases round-tripping every `vi` month name through `YYYY MMMM` and `YYYY MMM`.
- 1 case round-tripping every `tr` day name through `GGGG-[W]ww-dddd`, and 1 doing the same for `az-Latn` through `GGGG-[W]ww-ddd`.

Two deliberate choices worth flagging, because a weaker test here passes without the fix:

1. **The masks put the name last.** With a trailing literal in the mask (`MMMM D YYYY`), the regex engine backtracks out of the short branch and the bug is hidden. My first draft used exactly that mask and 3 of the cases stayed green against unpatched source. Masks ending in the name token (`YYYY MMMM`, `GGGG-[W]ww-dddd`) are what actually exercise the ordering.
2. **The locales were chosen for prefix pairs.** `vi` has `Th1`/`Th12` and `Tháng Mười`/`Tháng Mười Hai`; `tr` has `Pazar`/`Pazartesi`; `az-Latn` has `Ç`/`Ç.E`. `tr` `daysShort` (`Paz_Pzt_Sal_...`) has no prefix pair at all, so a `tr` `ddd` test would have been dead weight, which is why that one uses `az-Latn`.

## Validation

Run from `ui/`, with `ui/dist` built first (`node build/ensure-fresh-build.js`).

| command | before | after |
| --- | --- | --- |
| `npx vitest run -c test/vitest.config.js src/utils/date/date.test.js` | 145 passed | **154 passed** |
| `pnpm test` (`test:specs:check` + `test/parallel.js`) | | **all 6 suites pass, exit 0** |

`pnpm test` breakdown after the change, zero failures and zero skips:

- `test:unit` 272 files, 6378 tests
- `test:e2e:ssr` 1 file, 240 tests
- `test:hydration` 90 files, 150 tests
- `test:umd` 4 files, 78 tests
- `test:build` 2 files, 11 tests
- `test:hydration:pwa` 3 files, 8 tests

Lint: `pnpm lint:ui:check` passes (`oxfmt --check ./ui` + `oxlint ./ui`), and `git diff --check` is clean. Full-repo `pnpm lint` also reports 3 `tsconfig-error` entries for `docs/jsconfig.json`, `app-vite/playground-js/jsconfig.json` and `app-vite/playground-ts/tsconfig.json`; those reproduce on an unmodified checkout (missing generated `.quasar/tsconfig.json`) and are unrelated to this change.

**Counterfactual.** Reverting only `date.js` and keeping the new tests turns all 9 red, with behavioural assertions rather than errors, for example:

```
× round-trips every vi month name through the YYYY MMMM mask
  AssertionError: expected 2023-09-30T17:00:00.000Z
  to strictly equal 2023-10-31T17:00:00.000Z
```

that is, `Tháng Mười Một` parsed as October instead of November.

**Mutation testing.** Four mutants applied to the patched source, each breaking a different claim above, each caught:

| mutant | result |
| --- | --- |
| sort ascending (`a.length - b.length`) | 9 failed |
| leave `days` and `daysShort` unsorted, months only | 4 failed |
| sort in place (drop the `[...list]` copy) | 10 failed, including the pre-existing `resolves the weekday of an ISO week date from the dd token` and `keeps localized parsers isolated by all embedded locale names` |
| leave `daysMin` unsorted | 1 failed (the `dd` test) |

The third mutant is the one that shows why the copy matters: sorting in place reorders the very arrays `indexOf` uses. The fourth confirms the `dd` test is load-bearing rather than incidental coverage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved date parsing for locale names where shorter names overlap with longer names.
  * Ensured date parsing and formatting work consistently across Vietnamese, Turkish, and Azerbaijani locale names.

* **Tests**
  * Added coverage for overlapping custom month and day names.
  * Added round-trip tests for supported localized date names.


rstoenescu:
Thansk for contributing! This will go into v2.33.0

PR #18550, opened 2026-09-15
fix(QTable): rows-per-page options follow a controlled pagination

giaBaoJS:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### What is wrong

The rows-per-page selector builds its option list from `innerPagination`, the
table's own copy of the pagination state, instead of the page size actually in
effect:

```js
// ui/src/components/table/table-pagination.js:164
const opts = props.rowsPerPageOptions.includes(innerPagination.value.rowsPerPage)
  ? props.rowsPerPageOptions
  : [ innerPagination.value.rowsPerPage, ...props.rowsPerPageOptions ]
```

With `v-model:pagination`, `setPagination()` emits `update:pagination` instead
of writing `innerPagination` (`table-pagination.js:98-105`), so `innerPagination`
never moves past what the `pagination` prop held at mount. Every other consumer
reads `computedPagination`, which is where the prop lands.

Repro, with the table rendering three rows as asked:

```html
<q-table
  :rows="rows"
  :columns="columns"
  :rows-per-page-options="[5, 10]"
  v-model:pagination="pagination"
/>
```

```js
const pagination = ref({ page: 1, rowsPerPage: 5 })
// later, from outside the table:
pagination.value.rowsPerPage = 3
```

The selector's dropdown still offers only 5 and 10. Its `model-value` is the
effective page size (`QTable.js:1002`), so with 3 missing from `options` nothing
in the list is marked as selected. The mirror case is a stale entry: mount with
`rowsPerPage: 7`, move the parent to 10, and the list keeps offering 7 - a size
the app never put in `rows-per-page-options`, which a click then applies.

### What the fix does

Reads `computedPagination.value.rowsPerPage` - the page size in effect - when
deciding whether the option list needs the current size prepended.
`useTablePagination()` has no other use for `innerPagination`, so the argument is
dropped.

An uncontrolled table is unaffected: without an `update:pagination` listener
`computedPagination` *is* `innerPagination`.

### What the tests cover

`table-pagination.test.js`: a controlled parent moving to a page size outside the
options (it gets prepended), and moving between sizes (the previous one is
dropped, and a size already in the options is not duplicated).

`QTable.test.js`: the end of that path - the `options` QTable hands its
rows-per-page `QSelect` after the parent writes a size the options do not list.

All three fail on `dev` with the values above and pass with the fix.

### Boundary

`QTable.js:462` has the same stale read: `scrollTo()` emits `virtualScroll` with
`to: innerPagination.value.rowsPerPage - 1`. In the repro above,
`scrollTo(1)` reports `{ index: 1, from: 0, to: 4, direction: 'increase' }` while
only 3 rows are rendered. I left it alone because swapping in
`computedPagination` there would still be wrong in two other cases that are not
about control: `rowsPerPage: 0` reports `to: -1`, and a partially filled last
page reports the full page size. The value that is right in all of them is
`computedRows.value.length - 1`, which is a change of its own and not this bug.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the rows-per-page selector to accurately reflect controlled pagination changes.
  * Updated available page-size options when a parent changes the selected rows-per-page value.
  * Ensured the table displays the correct number of rows after controlled pagination updates.


rstoenescu:
Thanks again for contributing!
I'll also push a perf follow-up commit.
This goes into v2.33.0

Issue #18097, opened 2025-08-13
Usability bug in pagination component when using input


MaximilianKresse:
### What happened?

I've found a small usability bug related to the Pagination component together with the input field.

After pressing enter in the input field of the pagination component the focus is still in the field. Especially on mobile this feels a bit off. 

### What did you expect to happen?

The field shouldn't be focused after the press of enter.

### Reproduction URL

https://jsfiddle.net/1nymuabf/1/

### How to reproduce?

1. Click onto the page number to focus the input field.
2. Insert any number.
3. Press enter.

### Flavour

Quasar CLI with Vite (@quasar/cli | @quasar/app-vite)

### Areas

Components (quasar)

---

What do you think about it?

brianyimjh:
have submitted a PR (#18109) for this bug fix

brianyimjh:
Have submitted a new PR for this. #18112

rstoenescu:
Fix will be available in Quasar v2.18.5

PR #18552, opened 2026-09-18
fix(QPagination): the input field clears once the page is submitted

giaBaoJS:

- [x] Bugfix

- [x] No

**Other information:**

In `input` mode, the field keeps the text you typed after you hit Enter, even when the page you get is not the page you typed. Set `max` to 10, type `99`, press Enter: the model clamps to 10 and the placeholder becomes `10 / 10`, but the field still reads `99`.

`QPagination.js:514` hands the QInput a `value` prop. QInput declares `modelValue`, not `value`, so `value` falls through to `attrs`, and `getControl()` in `QInput.js` spreads its own `value: getCurValue()` after the attrs it forwards. The prop is overwritten before it reaches the element, so `newPage` never drives the field. `updateModel()` sets `newPage` back to `null` precisely to bring the placeholder back, and that reset has been going nowhere.

Binding `modelValue` instead makes the field follow `newPage` again: typing still renders through QInput's pending-value handling for `type="number"`, and the reset on submit clears the field so the placeholder shows the page actually in effect.

The test types a page above `max`, submits it, and checks the field is empty while the placeholder reports the clamped page. It fails on `dev` with `expected '99' to be ''`.

Boundary: this only changes what the field displays. Focus still stays in the field after Enter on desktop (#18097, whose PRs were closed) - untouched here. Blur already cleared the field through QInput's own `onFinishEditing`, so that path is unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed pagination input binding so manually entered page numbers are handled correctly.
  - Corrected the displayed pagination state after submitting a value beyond the maximum page.


rstoenescu:
Thanks again!

PR #18555, opened 2026-09-21
fix(colors): textToRgb keeps the alpha of the rgb() form

giaBaoJS:

- [x] Bugfix

- [ ] Yes
- [x] No

**Other information:**

`colors.textToRgb('rgb(255, 0, 0, 0.5)')` returns `{ r: 255, g: 0, b: 0 }`. The alpha is dropped.

`reRGBA` captures the alpha into `m[5]` for both spellings, because `,?([01]?\.?\d*?)?` sits outside the `(a)?` group. The branch that reads `m[5]` was gated on `m[1]`, which is only the literal `a` of the function name, so the four-argument `rgb()` form parsed fine and then lost its alpha.

No docs change: they already describe the fixed behaviour. `color-utils.md` documents the accepted input as `rgb(R, G, B<, A>)` and says "If Alpha channel is present in the original object it will be present also in the output". `rgb(r, g, b, a)` is the CSS Color 4 spelling too.

`rgba(r, g, b)` without an alpha still yields `a: 100`, now pinned by a test. QColor is untouched by this: it filters every value through `testPattern.anyColor()`, whose `rgbRE` rejects a fourth argument.

Boundary: I left `testPattern.rgbColor` / `anyColor` alone, so `rgb(255, 0, 0, 0.5)` is still not a valid QColor model. Widening them changes a documented validation rule and QColor's accepted input, which belongs in its own PR.

`pnpm test` in `ui/` is green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved color parsing for `rgb(...)` values that include an alpha channel.
  * Ensured `rgba(...)` values without an alpha component are treated as fully opaque.

* **Tests**
  * Added coverage for integer, decimal, and leading-dot alpha values in RGB color strings.


rstoenescu:
Thanks for contributing!

